PR #25, opened 2016-12-31
Use SDL 0.27.2 (from 0.25)

DenialAdams:
Hi,

I got the urge to update this library to the latest version of rust-sdl2, which merges the image and mixer crates into features of the sdl2 crate.

As far as I can see all examples still work!

Also, it seems there are two breaking changes on our side (if I'm understanding this right :) )

1) play_num_times should take an i32 instead of an isize to match the rust-sdl2 api change
2) MouseButton instead of Mouse in the mouse button events

Very new to rust, so hope all looks ok.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,11 +18,13 @@ readme = "README.md"
 name = "ggez"
 path = "src/lib.rs"
 
+[dependencies.sdl2]
+version = "0.27.2"
+default-features = false
+features = ["image", "mixer"]
+
 [dependencies]
-sdl2 = "0.25"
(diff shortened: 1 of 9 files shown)


DenialAdams:
Sure, that works. Bumping the SDL version actually fixes a serious crash problem on my partner's Ubuntu system (https://github.com/AngryLawyer/rust-sdl2/commit/5a329763261c89bd61ca8ed037ebbb6eed95f704), but I can have the project's Cargo.toml point to my fork for now.

Thanks for looking at this! I'll try to keep the PR up to date so it can be merged when ready.

PR #27, opened 2017-01-17
Show tried paths in ResourceNotFound error message

AaronM04:
This makes the ResourceNotFound error more informative by including a list of paths that were searched for the resource.

Not implemented yet: showing path to the zip file (if there is a zip file), and the path searched within the zip file.

Suggestions for improvement are welcome; I want to get this merged to help other newbies who may encounter  #26 ...


icefoxen:
Looks okay, thanks!

PR #45, opened 2017-02-21
Fix initialization sequence.

vickenty:
Setting required OpenGL version should be done before opening a window, otherwise it has no effect.


icefoxen:
:D

PR #50, opened 2017-02-23
Draw shapes using lyon tessellation.

vickenty:
This is a quick and dirty prototype of drawing functions based on lyon crate. Probably not in a mergeable shape yet.

This PR implements line, filled circle and ellipse drawing. Line circles and ellipses do not seem to be supported in lyon currently.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,4 @@ rodio = {git="https://github.com/tomaka/rodio.git", branch="master"}
 
 rustc-serialize = "0.3"
 toml = "0.2"
+lyon = "0.3.2"
--- a/examples/triangle.rs
+++ b/examples/triangle.rs
@@ -61,6 +61,7 @@ impl event::EventHandler for MainState {
         let rect = graphics::Rect::new(450.0, 450.0, 50.0, 50.0);
         graphics::rectangle(ctx, graphics::DrawMode::Fill, rect)?;
 
+        graphics::set_line_width(ctx, 4.0);
         graphics::line(ctx, &[
             Point { x: 200.0, y: 200.0 },
             Point { x: 400.0, y: 200.0 },
--- a/examples/triangle.rs
+++ b/examples/triangle.rs
@@ -4,6 +4,7 @@ use ggez::event;
 use ggez::{GameResult, Context};
 use ggez::graphics;
 use ggez::timer;
+use ggez::graphics::Point;
 use std::time::Duration;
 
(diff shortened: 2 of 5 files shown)


nical:
Hi (lyon author here), I'm happy to see interest in using the lyon's tessellators. Don't hesitate to ask for help on irc and file issues if something doesn't work or is missing.

I see that the line drawing API here is "immediate mode" in the sense that the line seems to be specified/tessellated/uploaded each frame. This works but with a certain performance cost. If you run into performance issues, I suggest looking into treating tessellated geometry as resources that stay on the GPU (like textures). That said, you the performance may be good enough as it is.

Another thing that may improve performance (while remaining immediate mode), is to batch several drawing primitives in the same vertex/index buffer pair and submit it all in one (or at least fewer) draw call.

For the missing features like ellipse strokes, don't hesitate to send pull requests my way :) Tessellation routines for the basic shapes are rather easy to implement (I tend to work on them when I to tired to work on the interesting stuff).

icefoxen:
@nical Thank you!  That's definitely going to happen before too long I expect...  This is just a first pass.  I think that we definitely want to do less allocating of new geometry and such, but this should be fine for now.  We have to fight the same battle for sprites anyway, so.

Thank you so much for making lyon, though if it's as usable as it seems you really should remove "experiment" from its description.  That word tends to make me shy away from relying on things labeled with it.  ;-)

icefoxen:
@vickenty Looking at your code more closely, I like it in general so far.  If you have no objections/updates I'll just go ahead and merge it.  Any thoughts on how we could cache geometry for tesselated shapes?  For some things like circles and rectangles we should just be able to keep a "prototype" around and use transforms to stretch/shear/whatever it as necessary, but when we start having complicated shapes like polylines with end-caps we really won't be able to always use that solution.

Might be best to just implement something like Love2D's `Mesh` type from the get-go and use that?

vickenty:
I have pushed final tweaks today, so you can merge this if you like.

I don't think there is any need to cache geometry behind the scenes by default. I made a small benchmark, and the current naive implementation on my five year old laptop can draw about 400 circles at 60 fps. This seems to be a quite large number to me, and is probably sufficient for many applications (I guess). Love2D also does not do any caching as far as I can see.

That said, I like the idea of building around the `Mesh` type. Applications that want more performance would be able to get what they need, and if ggez provides functions to build meshes for basic geometry, the upgrade path would be straightforward.

PR #52, opened 2017-02-27
Call cleanup hook every frame.

vickenty:
Device::cleanup() function has to be called every frame to allow backend to release resources.


icefoxen:
Herp derp.  Thank you!

Issue #38, opened 2017-01-23
All loader functions should operate from `&[u8]` first, and then loading from files should be implemented in terms of those.


icefoxen:
What it says on the tin.

icefoxen:
Resolved in commit b737e0a57c67b0b40f16cfe74f03636f8fccb62f .  Huzzah!

PR #62, opened 2017-04-21
Make font_from_bytes public

DenialAdams:
This may defeat the spirit of ggez, but I'd love to have font_from_bytes public.

I'm writing a semi-asychronous asset manager and doing this allows me to create a font on a seperate thread entirely, as I don't need to worry about carrying Context and passing that in.

I feel like this should be fine, since it seems analagous to Image::from_rbga, which is public.

Thanks!

### DIFF
--- a/src/graphics/text.rs
+++ b/src/graphics/text.rs
@@ -47,7 +47,7 @@ impl Font {
         Font::font_from_bytes(&name, buf, size)
     }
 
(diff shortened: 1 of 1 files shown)


DenialAdams:
Nightly failure is not due to an issue with my code, but rather

https://github.com/rust-lang/rust/issues/40470

icefoxen:
Doesn't defeat the spirit at all, see issue #38.  I'll accept this now but be aware that the precise API might shift around a little bit before 0.3 is released.

PR #76, opened 2017-05-14
Upgrade to lyon 0.4

vickenty:


src/graphics/tessellation.rs
```diff
@@ -52,64 +56,53 @@ fn build_path(points: &[Point], closed: bool) -> path::Path {
...
             return Err(GameError::RenderError(String::from("geometry tessellation failed")));
         }
     }
     Ok(buffers)
 }
 
-pub fn build_line(points: &[Point], line_width: f32) -> GameResult<Buffer> {
-    let path = build_path(points, false);
-    let opts = path_stroke::StrokeOptions::stroke_width(line_width);
+fn build_stroke(points: &[Points], close: bool, line_width: f32) -> GameResult<Buffer> {
```
vickenty: Yes, this should have been fixed yesterday I think. Sorry for pushing bad code. The current CI failure is only on beta and was caused by a regression: https://github.com/rust-lang/rust/issues/41944

PR #80, opened 2017-05-21
Gamepad support

kampffrosch94:
Pretty basic gamepad support via SDL2.
Adds a new module (input.rs), it's pretty sparse atm.
The generell context struct now also lugs a GamepadContext around.
A Gamepad example is also included.

### DIFF
--- a/examples/gamepad.rs
+++ b/examples/gamepad.rs
@@ -1,0 +1,45 @@
+extern crate ggez;
+
+use ggez::*;
+use ggez::event::*;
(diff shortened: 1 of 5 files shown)


icefoxen:
Thank you!  I will take a look when I have a moment.

icefoxen:
Looks pretty good all in all.  One request: replace the `unwrap()`'s `Result`'s.  Add a new enum member to `GameError` in error.rs if necessary.

Issue #39, opened 2017-01-25
Make filesystem code more clever


icefoxen:
Namely, make it detect when you're running from cargo (`CARGO_MANIFEST_DIR`) and look in the right location then as well.  That would solve the problem that people have with the required `resources` location sneakily moving about; it could always look in the cargo root dir and people could just look there.

Make sure it looks it up at runtime with `std::env::vars()` instead of at compile-time with `env!()`.  That way someone building a game don't bake their build paths into something running on someone else's system.

It also starts getting to the point where it might make life easier to have a full abstract VFS tree instead of the sort of simple-and-stupid "look in place A, then look in place B, then look in place C..." system that we have now.  Or at least generalizing that system.

icefoxen:
>  Sort of off topic, but I'd like to suggest adding some way to configure resource search dirs in the Config objects. Could ultimately be related to the work for this ticket.

icefoxen:
LOVE uses PhysicsFS.  Take apart PhysicsFS and see how it works.

icefoxen:
PhysicsFS seems to work pretty much the same way the `vfs` crate works, or the same way that ggez's file stuff works currently.  It doesn't build an actual filesystem tree, it just keeps a list of search paths (which may include archive files) and searches them in sequence.  Upon reflection this is not a bad way of doing it; where we currently fall short in comparison is probably things like symlinks and path sanitizing, and having a nicely pluggable API.

Another thing worth noting is that PhysFS is multithread-safe but not really multithreaded; every interaction that may conflict will lock, and that is pretty much every interaction.  That's just life, I expect.

icefoxen:
`vfs` is somewhat inconvenient to compose, 'cause you can't use its `VFS` trait as a trait object.  So I just wrote my own, loosely trying to comply with PhysicsFS's general style and 

It resulted in me changing "only relative paths" to "only absolute paths" for no particular reason besides compliance with PhysFS, which will probably break everything ever, but meh.  It's not quite done as of commit 1c336061ff4c68d580fb82479f642c69e9b99c9f but does indeed appear to work.  And makes adding another directory root based off of a cargo feature flag *trivial*.

Totally worth.

PR #84, opened 2017-05-30
Read CARGO_MANIFEST_DIR at runtime instead of compile-time

17cupsofcoffee:
I spent last night playing around with ggez for the first time - I absolutely love it so far (*"Love2D in Rust"* is basically my dream game library, so no surprises there)!

The only major issue I ran into was when I tried to set up the `cargo-resource-root` feature - no matter what I did, I couldn't seem to get it to load my sprites. Printing out the initial state of the VFS tipped me off to the cause:

Screenshot of a console showing ggez searching for resources in the Cargo registry directory - huh?!

It seems that `CARGO_MANIFEST_DIR` gets set for each dependency during a build - meaning that calling `env!("CARGO_MANIFEST_DIR")` from within ggez actually gives you *ggez's* root directory, rather than your game's. Grabbing the value of the environment variable at runtime instead gives much more reasonable results:

Screenshot of a console showing ggez searching for resources in the expected place

Looking at #39, this was something that was planned anyway to avoid exposing build paths in your binary, and it seemed like an easy change to make, so I thought I'd give it a go :) Let me know if any changes need making, or if there's a reason I'm missing why this is a bad idea!

---

Areas I'm slightly uncertain about with this change (quite new to opening PRs!):
* Is it okay for the code to just get ignored if `CARGO_MANIFEST_DIR` doesn't exist?
* Was this issue also present on MacOS/Linux? I only have access to a Windows PC :(
* Are there any issues with the code style that need fixing? [rustfmt is broken on Windows at the minute](https://github.com/rust-lang-nursery/rustfmt/issues/1525), so I can't run it.

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -155,10 +155,12 @@ impl Filesystem {
         // Cargo manifest dir!
         #[cfg(feature = "cargo-resource-root")]
         {
-            let mut path = path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
-            path.push("resources");
-            let physfs = vfs::PhysicalFS::new(&path, false);
-            overlay.push_back(Box::new(physfs));
+            if let Ok(manifest_dir) = env::var("CARGO_MANIFEST_DIR") {
+                let mut path = path::PathBuf::from(manifest_dir);
+                path.push("resources");
+                let physfs = vfs::PhysicalFS::new(&path, false);
+                overlay.push_back(Box::new(physfs));
+            }
         }
 
         let fs = Filesystem { 


icefoxen:
Oh well that's interesting; I thought I had tested that and I thought that it worked, but now that I try again my test was broken.  So, in order:

 * Yeah, if you're not building ggez with cargo then I don't know what kind of sane default to use.  Making it the programmer's problem is valid.  ;-)
 * Yes, it apparently is present on Linux.
 * Nope, you're fine.  Re-rustfmt'ing everything if necessary is trivial.

icefoxen:
Actually if you want to be *really* pro, we need better error messages so it prints out the paths it searched when it can't find a resource.  ;-)

17cupsofcoffee:
I may have a go at that this weekend, then :smile:

17cupsofcoffee:
Ah, looking back, it looks like something similar was done in #27, but it didn't survive the refactoring from #39 - so a convenient error type already exists for this (`ResourceNotFound`), just needs the logic writing to get the VFS to be able to output it. I will definitely try to make some time to have a crack at implementing that in the next few days 👍

PR #88, opened 2017-06-03
Bring back GameError::ResourceNotFound

17cupsofcoffee:
As requested in #84, I've had a go at updating the error message returned by `OverlayFS::open_options` to contain a list of searched paths. A lot of the work had already been done for me in #27 by @AaronM04 (cheers!) - just had to find a way to bring his code into the new VFS module.

Here's the changes I've made:

* There wasn't currently a way of getting the real root path of a `VFS` trait object, so I added a `to_path_buf` method. It returns `Option<PathBuf>`, just to futureproof in case a VFS type without a logical path gets added (e.g. a in-memory filesystem).
  * This is basically what the [`vfs`](https://docs.rs/vfs/0.2.0/vfs/trait.VPath.html#tymethod.to_path_buf) crate does.
  * As part of implementing this, I had to change the `source` field on `ZipFS` to a `PathBuf` - that made it more consistent with `PhysicalFS` though, so hopefully that's alright. I also updated the error messages to be more consistent, but I can back that out if you think they were fine as was.
* With that in place, I was able to collect the `PathBuf`s in `OverlayFS::open_options` and return a `GameError::ResourceNotFound`, as the old version of the filesystem used to.
* I also updated the filesystem tests to reflect the change in error type.

the new error message, with a list of searched paths included

Let me know if there's any issues/things you want changing!

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -381,7 +381,7 @@ mod tests {
         {
             let rel_file = "testfile.txt";
             match fs.open(rel_file) {
-                Err(GameError::FilesystemError(_)) => (),
+                Err(GameError::ResourceNotFound(_, _)) => (),
                 Err(e) => panic!("Invalid error for opening file with relative path: {:?}", e),
                 Ok(f) => panic!("Should have gotten an error but instead got {:?}!", f),
             }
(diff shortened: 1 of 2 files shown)


icefoxen:
You are the best, I've been meaning to fix this for weeks.  :blush:

17cupsofcoffee:
No problem, glad I could help out :smile:

Issue #95, opened 2017-06-19
Unable to load (and save) context config file on Linux


chinatsu:
Using x86_64 rustc 1.18.0 on Arch Linux, I don't seem to be able to load nor save configuration files. With some help from @icefoxen on IRC, I think we may have found out why this is.

The related bit of code I have been working on can be found here: https://github.com/chinatsu/ggetris/blob/master/src/main.rs#L54

I did some debug prints in my attempts to diagnose what was up, the comments represent what's printed from each of the prints.

```rust
println!("{:#?}", ctx.filesystem);
/* 
Filesystem {
    vfs: OverlayFS {
        roots: [
            <PhysicalFS root: /home/cn/Hello/dev/ggetris/target/debug/resources>,
            <PhysicalFS root: /home/cn/.local/share/ggetris>
        ]
    },
    resources_path: "/home/cn/Hello/dev/ggetris/target/debug/resources",
    zip_path: "/home/cn/Hello/dev/ggetris/target/debug/resources.zip",
    user_data_path: "/home/cn/.local/share/ggetris"
}
*/

println!("{:?}", ctx.filesystem.read_config());
// Err(ConfigError("Config file not found"))

println!("{:?}", ctx.filesystem.open("/conf.toml"));
// Ok(VfsFile)

println!("{:?}", ctx.filesystem.write_config(&conf::Conf::new()));
//  Err(ConfigError("Could not write config file because a directory is in the way?"))
```

As far as reading config files go, it seems that the context's filesystem sees the config file in the `user_data_path` with the prepending slash: `/conf.toml`. As such, `read_config()` doesn't match the `conf.toml` it's looking for. This is also why I can open the file by calling `ctx.filesystem.open("/conf.toml")`

I'm unsure what the `write_config` call tries to tell me. At first I thought maybe it tried to save the config in `target/debug/resources`, and that "resources" wasn't a directory. The error message remains unchanged after creating that directory. `~/.local/share/ggetris` is also a directory.

icefoxen:
Yeah, this is because the "default config file" in filesystem.rs is "conf.toml", not "/conf.toml".  The way the file system code is set up, I had to basically either say "all paths must be absolute" or "all paths must be relative", and PhysicsFS chooses the former so I go along with that.  I just forgot to update the variable with the conf file name.  Easy fix.  This should probably be explained more fully in the documentation though.

[...]

PR #96, opened 2017-06-21
Fixing bugs in saving/writing config file

chinatsu:
I believe I've addressed everything brought up in #95, although I'll admit I'm a bit uncertain if I've done it *right*. 

I did notice that in `write_config`, it only checked if the config file already exists, and threw the "directory in the way?" error if it isn't there. To remedy that I changed it to always create a file at conf_path, which should overwrite any file with that name already there (as described in the docs). Then it checks if the file exists after that, if it doesn't exist it means something failed.

What bothers me is that now, if I try changing `CONFIG_NAME` back to "conf.toml", I'm getting the following error: 'ResourceNotFound("conf.toml", ["/home/cn/Hello/git/ggez/resources"])', so maybe I've overlooked something; I don't know what raises that error.

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -336,20 +336,6 @@ mod tests {
     }
 
     #[test]
-    fn test_write_config() {
-        let mut f = get_dummy_fs_for_tests();
-        let mut conf = conf::Conf::new();
-        // The config file should end up in
-        // the resources directory with this
-        match f.write_config(&conf) {
-            Ok(f) => (),
-            Err(e) => panic!("{:?}", e),
-        }
-        // Remove the config file!
-        f.delete(CONFIG_NAME);
-    }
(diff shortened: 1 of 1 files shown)


chinatsu:
On a related note, wouldn't it be nicer to give any paths ending at a directory a trailing slash?

icefoxen:
It might be nicer but wouldn't necessarily be correct.  I may well be misunderstanding the conventions, but my understanding is that `/foo/bar` refers to the directory object `bar`, `/foo/bar/` refers to the contents of the directory object `bar`.

chinatsu:
It just seems ambiguous to me because `/foo/bar` could also refer to a file named `bar` in the directory `/foo/`. There may very well be well established conventions on this though, so don't mind me. :D

Issue #115, opened 2017-07-28
graphics::get_screen_coordinates() does not return a well-formed Rect


jrdlph:
related to #93
The return value of `graphics::get_screen_coordinates()` does not respect the meaning of the fields of a `Rect`, so invoking `.top()`, `.bottom()`, `.point()`, etc  on the `Rect` returned will not give you what you are looking for. The `x` and `y` fields should either represent the center point of the display (as in the current implementation of `Rect`), or the implementation of `Rect` should change so that `x` and `y` represent the top/bottom left corner.

I think the litmus test should be that drawing this `Rect` should draw a rectangle that fills the viewport. This would make reasoning about other `Rect`s in relation to the viewport much simpler.

icefoxen:
Duplicate of #93

PR #136, opened 2017-08-30
Use Rect for graphics::set_screen_coordinates() param

### LINKED ISSUE #93: graphics::get_screen_coordinates() API change
Instead of a `Rect` it should probably return top,bottom, left, right coordinates to be symmetrical with `graphics::set_screen_coordinates()`

Or vice versa.  Either way those two functions should take and set the same values.  Love2D doesn't quite have these functions since it has a graphics transform stack that serves a similar purpose, as far as I can tell.

sanpii:
Fixes #93

### DIFF
--- a/src/graphics/mod.rs
+++ b/src/graphics/mod.rs
@@ -699,14 +699,11 @@ pub fn set_point_size(ctx: &mut Context, size: f32) {
 /// be whatever you prefer.
 pub fn set_screen_coordinates(
     context: &mut Context,
-    left: f32,
-    right: f32,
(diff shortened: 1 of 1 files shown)


icefoxen:
Thank you!

Still needs to make sure the concerns raised by https://github.com/ggez/ggez/issues/115 are addressed, but that's trivial enough I'll do it.

icefoxen:
For your own satisfaction you may wish to look at commit d17f580d10a9d1d0bf137326dc7170d42a26aeb6

Issue #135, opened 2017-08-29
Add "sprite batching"


termhn:
As referenced in #127, I thought I'd open up its own issue as it's something I'd be interested in contributing to, so we can discuss how to do it The Right Way TM :P 

One thing I'd like to know people's thoughts on is if you think it would be good to try to implement generic automatic sprite batching for objects with some set of common characteristics that would lend themselves to be batched or if we want to make the user manually use a SpriteBatch?

icefoxen:
Automatic batching for things that happen to be similar is a nightmare waiting to happen because it hides context-dependent information from the user.  It's that sort of trickery that makes OpenGL drivers so finicky (oh I reordered these two draw calls and now things are super slow)...  plus, gfx-rs caches unchanged context already.

If we want things to be fast we just need to select a certain amount of state that can be shared (basically the `DrawParams` object) and shove it into a UBO.  Look at how Love2D does it; it doesn't have to be super tricksy to be useful.

termhn:
Fair enough, basically what I was thinking as well.

I do think it would be useful for SpriteBatching to be generic to all Drawables rather than just Images, which unless I'm understanding how this would work wrong, should be possible.

icefoxen:
That's a bit of a taller order because `Mesh` breaks the assumptions that `Image` and `Text` both have.  (ie, that it's a textured quad.)  Still, it might be possible.

termhn:
True. Perhaps, at the start anyway, have it only extend to rectangles at first so that `point`s and other uses of rectangles in procedurally-generated graphics (as had been mentioned in #127) would be able to take advantage without having to make small rectangle images to use as sprites.

dpogretskiy:
I've made a tool, for parsing json from TexturePacker and creating other representation from it, i did some animations based on it already. Not sure about performance, but it basically shares one image to draw everything. Ui part is just a bunch of copy-pasted code, but other things can be useful.

https://github.com/dpogretskiy/spritesheet-marker

termhn:
Gonna start playing with this later today I think. The way LOVE does it is nice--it's useful just on its own for batching one image on each sprite but also extensible enough through adding "Quads" which specify UVs on the underlying texture for each sprite to use that it also works in conjunction with sprite sheets/atlases... 

I'm not sure I like that every sprite instance must have its own quad in LOVE; I was thinking it might be better to be able to add "quads" to the batch and get an ID back for each, and then be able to optionally specify a quad ID when you add a new sprite instance to the batch.

[...]

Issue #127, opened 2017-08-19
Rendering increasing numbers of points becomes quite slow


Sushisource:
In my looking through the code I think this isn't so much specific to points, but rather just having lots and lots of polys being drawn (since the points are converted into rects, which are converted into polys).

In my case I'm rendering a cellular automata simulation, so there are many thousands of points being drawn. The slowdown starts to get pretty bad around the 6k mark, on my machine at least.

At least in the points case, I think this could be optimized by using GL_POINTS primitives, or whatever the backend-agnostic gfx-rs equivalent is, if such a thing exists.

Hopefully this is helpful info. I'm happy to contribute PRs, albeit probably pretty slowly.

icefoxen:
I suspect the problem is less a matter of not using GL_POINTS (since iirc it is not well supported these days anyway) and more that each draw sends a separate openly draw call.

I believe the Right Answer is sprite batching, which lets you collect a large buffer of mostly-similar drawing instructions and then send them to the GPU all at once. There's a little precursor code to that but it's not connected to anything yet.

For cellular automata you might be well served just by creating a buffer with all your pixels in it, turning it into an Image with `Image::from_rgba8`, and displaying that. Example here: https://gist.github.com/icefoxen/ffde9dc776848c80e25656038e9248f9

icefoxen:
However if you link some example code that displays the problem I can profile it and see what's actually going on.

Sushisource:
Yeah, I'm by no means an OpenGL expert (or even amateur). Your solutions sound like they make perfect sense. I'll try out the Image based approach next.

Here's the game (well, very very start of something that maybe will be a game eventually) I'm doing the CA thing for. Forgive the mess 🤕. I didn't make a pared-down example but the amount of code here is pretty small anyway:
https://github.com/Sushisource/Rustlike

I just committed converting the code from glium to ggez. The slowdown while rendering many points should be pretty evident.

Manghi:
[...]

PR #144, opened 2017-09-08
Sprite batching

termhn:
Addresses #135 / will address #127 in the future

Well... it works!

Certainly not done yet, but proof of concept is working. I'm getting ~2000 fps with 10,000 instances of the base tile being drawn each frame on my machine.

### DIFF
--- a/src/graphics/mod.rs
+++ b/src/graphics/mod.rs
@@ -86,14 +86,23 @@ gfx_defines!{
 
     /// Internal structure containing global shader state.
     constant Globals {
-        transform: [[f32; 4];4] = "u_Transform",
+        projection: [[f32; 4]; 4] = "u_Projection",
         color: [f32; 4] = "u_Color",
+    }
+
+    constant GlobalTransform {
+        translation: [f32; 2] = "u_Translation",
+        scale: [f32; 2] = "u_Scale",
+        offset: [f32; 2] = "u_Offset",
+        shear: [f32; 2] = "u_Shear",
+        rotation: f32 = "u_Rotation",
     }
 
     pipeline pipe {
         vbuf: gfx::VertexBuffer<Vertex> = (),
         tex: gfx::TextureSampler<[f32; 4]> = "t_Texture",
         globals: gfx::ConstantBuffer<Globals> = "Globals",
+        transform: gfx::ConstantBuffer<GlobalTransform> = "GlobalTransform",
         rect_instance_properties: gfx::InstanceBuffer<RectInstanceProperties> = (),
         out: gfx::BlendTarget<ColorFormat> =
           ("Target0", gfx::state::MASK_ALL, gfx::preset::blend::ALPHA),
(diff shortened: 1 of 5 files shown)


icefoxen:
This looks pretty great, thank you so much! I'll go through it in detail this weekend.

src/graphics/mod.rs
```diff
@@ -297,10 +297,18 @@ impl GraphicsContext {
             pipe::new(),
         )?;
 
-        let (quad_vertex_buffer, quad_slice) =
+        let rect_inst_props = factory.create_buffer(
+            1,
+            gfx::buffer::Role::Vertex,
+            gfx::memory::Usage::Dynamic,
+            gfx::SHADER_RESOURCE
+        ).unwrap();
```
icefoxen: Please remove this unwrap?  It should return a Result.

src/graphics/spritebatch.rs
```diff
@@ -51,29 +54,64 @@ impl SpriteBatch {
...
-        for s in &self.sprites {
-            graphics::draw_ex(ctx, &self.image, *s);
+        let gfx = &mut ctx.gfx_context;
+        if gfx.data.rect_instance_properties.len() < self.sprites.len() {
+            gfx.data.rect_instance_properties = gfx.factory.create_buffer(
+                self.sprites.len(),
+                gfx::buffer::Role::Vertex,
+                gfx::memory::Usage::Dynamic,
+                gfx::TRANSFER_DST
+            ).unwrap();
```
icefoxen: Again, remove the unwrap; feel free to change the return type of the function.

src/graphics/spritebatch.rs
```diff
@@ -51,29 +54,64 @@ impl SpriteBatch {
...
         }
+        gfx.encoder.update_buffer(
+            &gfx.data.rect_instance_properties, &self.sprites[..], 0
+        ).unwrap();
```
icefoxen: Here too.

src/graphics/spritebatch.rs
```diff
@@ -51,29 +54,64 @@ impl SpriteBatch {
...
-    /// Alters a sprite in the batch to use the given draw params.
-    pub fn set(&mut self, handle: SpriteIdx, param: graphics::DrawParam) {
-        self.sprites[handle] = param;
+    /// Alters a sprite in the batch to use the given draw params
+    pub fn set(
+        &mut self,
+        handle: SpriteIdx,
+        param: graphics::DrawParam
+    ) {
+        self.sprites[handle] = param.into();
```
icefoxen: Honestly we should check that `handle` is within bounds of the spritebatch.

termhn:
Sure thing. :) Also going to work on getting the batch-wide draw param working now...

src/graphics/spritebatch.rs
```diff
@@ -51,29 +54,64 @@ impl SpriteBatch {
...
+        let src_width = param.src.w;
+        let src_height = param.src.h;
+        // We have to mess with the scale to make everything
+        // be its-unit-size-in-pixels.
+        // We also invert the Y scale if our screen coordinates
+        // are "upside down", because by default we present the
+        // illusion that the screen is addressed in pixels.
+        // BUGGO: Which I rather regret now.
+        // let invert_y = if gfx.screen_rect.h < 0.0 { 1.0 } else { -1.0 };
+        // TODO: Figure out whether implementing this is needed/how to do it cleanly
```
termhn: I think implementing a method on Image called, say, `convert_param()` that would do this for you might be helpful... problem is I don't want to make you pass in the context to the `add()` method but I'm not sure what a good way would be otherwise to get access to `gfx_context.screen_rect`

termhn:
Transform for the full batch is now implemented & showcased in the example. Performs really well even on my macbook.

icefoxen:
AWESOME.  Looks great and performs smooth as butter on my laptop with an Intel HD 530 GPU.  Thank you so much!  :heart:

PR #150, opened 2017-09-13
Generic mesh batching with helpers for common meshes

termhn:
Further adds on to #135 and now addresses #127. Allows support for batch rendering of meshes. To be completely honest, I'm not sure I'm 100% sold on the exact way that it currently works. I might try to do it in another way where every sprite batch stores its own vertex buffer and texture so there doesn't need to be two different types. Then make a Builder struct that would provide helpers for creating one of those batches as you want it.

Thoughts?


icefoxen:
It's interesting but I suspect you may be over generalizing a bit. It's not even clear atm how ggez should handle textured meshes, which is inevitably going to have to change this.

Start from the use case and design to be convenient for that, instead of just generalizing because it seems like a good idea. Rust makes it really easy to generalize too early.

termhn:
Yeah, I think you're right, especially given that the new mesh builder basically allows you to do this. 

The main use case I wanted to support was the one(s) in #127 as far as batching points and rects for use with cellular automata simulations.

termhn:
Closing this since I don't think it's the best option.

Issue #149, opened 2017-09-12
Better camera/transform controls?


icefoxen:
See https://github.com/ggez/ggez/issues/135#issuecomment-328914981 for start of discussion.

Continuation from IRC:

```
15:36 < termhn> Icefoz: what do you think about, instead of adding an actual camera, simply adding a view matrix to the Globals uniform and exposing it in the same way so that outside cameras can modify it... this way we aren't burdened by supporting an official camera but allow third party cameras to have tighter integration into the rendering pipeline than having to implement separate draw commands and just modify the parameters
15:37 < Icefoz> That's a better idea.
15:38 < termhn> I kinda like that idea, might write up a demo pr later, wouldn't be too hard I think...
15:38 < Icefoz> Check out the Love2D docs but don't cleave to them too closely.
15:38 < termhn> will do
15:39 < Icefoz> Also I assume you are aware of https://github.com/ggez/ggez/blob/master/src/graphics/mod.rs#L513-L527
15:40 < Ralith> not sure why a camera would involve "separate draw commands"
15:40 < Ralith> separate from what?
15:40 < termhn> ralith: not actual separate draw commands to the gpu
15:41 < termhn> just a new trait that extends the current ggez "Drawable" abstraction
15:42 < termhn> see: https://github.com/ggez/ggez-goodies/blob/master/src/camera.rs#L99
15:42 < Icefoz> I still really think you're yak-shaving with the camera thing, termhn.
15:42 < Icefoz> Do you have a particular use case you want to support?
15:43 < termhn> https://github.com/ggez/ggez/issues/135#issuecomment-328914981
15:44 < Icefoz> That's in depth enough I'll have to read it when I'm awake.
15:44 < Icefoz> Which hasn't happened in like three weeks.
15:44 < kvark> Ralith: I'm too lazy now to dig it up and point you more precisely, sorry
15:47 < termhn> Icefoz: Basically, sprite batches don't integrate that well with an external camera because you have to add instances while thinking about default ggez screen/pixel coordinates and then place all other objects and move the batch as a whole thinking in camera world-space coordinates
15:48 < Icefoz> I would assume you have to add your instances while thinking about the spritebatch's coordinate system, and then when you draw that gets transformed to the world's coordinate system.
15:49 < Icefoz> I don't know if there's a better way of doing it but I agree it would be nice.
[...]

Issue #145, opened 2017-09-10
Send transform data to the gpu as matrices


termhn:
Will become easier when #102 hits, but this would mean we could compute one matrix on the cpu for each object and then send that to the gpu... or just convert all the current translations into matrices and send those... Not sure of the performance impact would be of either of those options (would have to profile) but in either case, it would make the operations happening in and data being sent to the shaders much more clear.

icefoxen:
Do we really want to compute one matrix per object on the CPU, rather than on the GPU which is 1000x better at it?

I don't know, maybe. Having each object get a matrix rather than parameters might make it simpler to eventually  support different shaders...

termhn:
Well, doing it on the CPU means it happens once per object instead of once per vertice. Also, correct me if I'm wrong, but I don't think the gpu is actually better at doing math, only at doing many math problems at the same time... so I think it might actually be less efficient to do on the gpu.

Could be nice for shaders, too.

icefoxen:
If most objects are 4-6 vertices...  

I honestly don't know.  It deserves some research, I admit.

termhn:
Haha yeah, that's exactly what I was thinking; for a 3d engine where there's often hundreds or thousands of vertices it seems like it could make more sense, but... I don't know really... that's why I included the "dunno performance impact, would have to profile" disclaimer originally ;P

icefoxen:
@termhn Can this be closed or is there more you want to do?

termhn:
Yep can be closed.

PR #151, opened 2017-09-14
new transform pipeline

termhn:
Possibility for what a new transform pipeline could look like as discussed in #149 and #145. The instance data cannot be sent as a matrix directly, though it would be possible to encode it which I might look into in the future. Currently, I anecdotally see a ~5-10% FPS increase on my desktop when running the spritebatch example vs the current master.

* Only send one combined MVP matrix to GPU
* use nalgebra to generate ortho projection matrix
* store view matrix in a vec with helpers for changing it
* store model matrix in a vec with helpers for changing it
* add helper conversion from DrawParam to matrix

### DIFF
--- a/examples/astroblasto.rs
+++ b/examples/astroblasto.rs
@@ -12,7 +12,6 @@ use ggez::{Context, GameResult};
 use ggez::graphics;
 use ggez::timer;
 use std::time::Duration;
-use std::ops::{Add, AddAssign, Sub};
 
 /// *********************************************************************
 /// Basic stuff.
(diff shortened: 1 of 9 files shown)


icefoxen:
I like it. I like it a lot. Need to go through it in more detail, but it looks very nice.

icefoxen:
Aieee, I should have merged this before rewriting and refactoring mesh stuff XD

termhn:
Heh. I'm working on a bit of change to the api to be more like how Love2D does it, but I'm not sure that it's actually that much better than how it is now. Possibly more user friendly? I dunno I'll push a commit here in a bit and you can take a look.

icefoxen:
Okay!  Two questions:

1) Is it ready to merge?  Keeping in mind details can be amended later.
2) Can you rebase it off of current master, given all the refactoring that happened with Mesh?  :/

termhn:
1. Yes, probably, though I have not actually tested some of the functions thoroughly or at all (specifically the new ones I added yesterday.
2. I'll take a look but I don't see why not... this shouldn't have affected that all too much I don't think

termhn:
Alright I'm going to start rebasing this on current master now

Issue #152, opened 2017-09-14
Antialiasing


termhn:
I think we should definitely allow some kind (or possibly multiple kinds) of built-in antialiasing. For games other than pixel-art based (and even then), antialiasing is basically a requirement IMO. 

Possible options include using MSAA or doing post-processed antialiasing like FXAA in a shader. I honestly haven't done much research on these things so I don't have anything in mind as far as the best way to do this, but wanted to put it down to discuss as it's definitely something I want to work on in the future.

icefoxen:
Doing FXAA is the easy way but requires some sort of shader framework.  Doing MSAA doesn't need shaders but does need backend support and I don't know exactly what the implications of that are right now.

termhn:
MSAA is now implemented, not sure if we really need to add FXAA as well but it could be a good excuse to start working on user shaders...

nlordell:
@termhn  I started looking at how to add support for shaders à la LÖVE (http://blogs.love2d.org/content/beginners-guide-shaders) and posted what I found in a new issue #157 - not sure if you are interrested...

termhn:
Testing on my macbook today and multisampling appears to not work. Could be why I was getting screwed up in the first place. Not sure if it's an OS X thing or what; haven't had time to dig deeper.

icefoxen:
I made an example program to play with this; it works great on my laptop.  Though it doesn't support 16x MSAA, so I made the default 1x.

termhn:
Ha! That was the problem. My laptop doesn't support 16x either, works great with 8x though. Perhaps I should add a note about that in the documentation...

Also IMO it should default to either 4x or 8x; one of the things that was a bit of a turn off for me from ggez at the beginning was the really aliased look which makes things look worse than many other engines, when in reality it's just as capable.

I wonder if there's a way to automatically detect the capabilities of the host's graphics card as far as mutisampling goes...

icefoxen:
Alas, graphics drivers.  When I tried to set it to 16x it panicked with a "cannot create video context with the given settings" type error, while it sounds like yours ran and just didn't use MSAA.  Do you have an nvidia graphics card by any chance?  :D

I lean towards defaulting to 1x because it means that, well, nobody will ever have the examples fail to run with a "cannot create video context with the given settings" type error.

I'm sure there's ways to query the graphics card capabilities but I'm not sure that SDL or gfx-rs exposes any of them...  hmm.  That would be a VERY good thing to have though.

termhn:
[...]

PR #156, opened 2017-09-18
Start for MSAA

termhn:
Addressing #152 Began trying to implement MSAA. This should in theory work as it is now, but gfx_window_sdl needs to be patched to allow the surface to be multi-sampling aware. Working on that now but it's not working yet for some reason. Need to run it through renderdoc or apitrace or something to inspect closer.

### DIFF
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -55,6 +55,7 @@ impl From<FullscreenType> for SdlFullscreenType {
 ///     vsync: true,
 ///     min_dimensions: (0, 0),
 ///     max_dimensions: (0, 0),
+///     samples: NumSamples::One,
 /// }
 /// ```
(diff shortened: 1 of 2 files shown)


icefoxen:
Oh cripes more?

icefoxen:
As long as you're rewriting the ggez graphics system anyway, do you have any clever ideas for making it possible for users to provide custom shaders? Think about it.

termhn:
Heh, I've been thinking about it. I need to look more into how gfx actually builds shader programs and stuff. I like the way LOVE does it API wise a fair amount, but I haven't really looked into how that would look like with gfx at all.

termhn:
Alright, it actually did and does work, I just didn't notice because I had the number of samples relatively low. I'm dumb.

It's now defined in the WindowMode with a default of 16 samples, but it's not yet editable after the window has been created.

Everything looks so niiiiiiiiiiiiice with 16 samples *_*

PR #158, opened 2017-09-19
Shader support

nlordell:
There are still a few things to fix before this should be merged in
- Finalize the API
- There is some unsafe code in the PixelShaderHandle that I want to get rid of
- The MSAA changes weren't properly merged in. In particular the PSO that we create per PixelShader does not observe the MSAA settings.

### DIFF
--- a/src/graphics/pixelshader.rs
+++ b/src/graphics/pixelshader.rs
@@ -2,14 +2,14 @@
 //! with ggez for cool and spooky effects. See the `shader` example for a
 //! taste...
 
-use gfx;
 use gfx::*;
 use gfx::handle::*;
 use gfx::pso::*;
 use gfx::pso::buffer::*;
 use gfx::shade::*;
+use gfx::state::*;
 use gfx::traits::{FactoryExt, Pod};
-use gfx_device_gl::{CommandBuffer as GlCommandBuffer, Resources as GlResources};
+use gfx_device_gl::Resources as GlResources;
 use std::cell::RefCell;
 use std::fmt;
 use std::io::prelude::*;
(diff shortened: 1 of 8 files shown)


src/graphics/pixelshader.rs
```diff
@@ -0,0 +1,299 @@
...
+use std::io::prelude::*;
+use std::marker::PhantomData;
+use std::path::Path;
+use std::rc::Rc;
+
+use Context;
+use error::*;
+use graphics;
+
+/// A `PixelShader` reprensents a user-defined shader that can be used with a
```
termhn: I think this should be `PixelShaderGeneric`?

src/graphics/pixelshader.rs
```diff
@@ -0,0 +1,299 @@
...
+/// ggez graphics context independent of gfx::Resources type.
+///
+/// As an end-user you shouldn't ever have to touch this, use `PixelFormat`
```
termhn: And here, `PixelShader`

termhn:
Right, so... I think this is awesome & have an idea or two about it:

1. Perhaps it would be better to store references to shader programs (or their generic `PixelShaderDraw`) inside a Vec similar to how samplers are currently stored instead of only storing an Rc to the current one and having the user store them? It would, in theory, eliminate the unsafe code I think... Add a `shaders: Vec<Box<PixelShaderDraw<R, CB>>>` to the `GraphicsContextGeneric` and then handles to shaders can simply be an index into the cache, perhaps with its own type (`type ShaderIdx = u16;` or something). Then keep a `current_shader: ShaderIdx` that keeps the index of whatever the currently selected shader is. Since I don't think people will want to/have a reason to delete shaders mid-program, it shouldn't be a problem. We can also then just wrap the default shader into a PixelShader and put it in the first index of `shaders`. 
2. I'm not sure I'm sold on the giving back a lock mechanism. It's cool, and I could probably be sold on it, but I feel like it moves away from the simplicity of LOVE's api. Plus if we use a cache as suggested above, we could have our `graphics::set_shader` signature be something like `graphics::set_shader(&mut self, shader: Option<ShaderIdx>)` and if `None` is given then just set the `current_shader` to 0 and more closely emulate the LOVE api.

I didn't put a huge amount of thought into the specifics of what I said above so I may be missing something obvious that would prevent that from working.

termhn:
Hmmm. Alright, after thinking about it more, I don't think my idea of putting things in a cache is really practical because of `send`, which I hadn't thought about.

termhn:
I implemented multisampling support as it is for the default PSO for user shader PSOs here https://github.com/nlordell/ggez/pull/1

nlordell:
I was thinking the same thing. The PixelShaderDraw trait is used to create trait objects to hide the type of the user shader constant data `<C>`.

However, it can still be done if we split a shader into a program and a handle and have both share the `buffer: Buffer<Resources, C>` bit allowing the handle to be able to `send` and keeping all the shaders in a cache. The added benefit of something like this is that is simplifies the ownership model (since the GraphicsContext owns everything) and we can implement the default shader as a PixelShader and avoid duplicate code paths for creating PSO (which led to MSAA not being respected for PixelShaders).

We can also easily have an API like this if we keep a shader cache:

```rust
graphics::set_shader(ctx, my_shader);
{
    let lock = graphics::use_shader(ctx, my_shader);
}
```

The idea behind the lock was also to guarantee that the shader was still alive when the `GraphicsContext` tried to call it. If we want to remove it and be "safe" I think a shader cache is the way to go. I'm going to take a stab at implementing it and then we can decide if we like it better or not

termhn:
Awesome. I took a stab at implementing it using a cache and got it all working well until `send` and then I was like "well shoot, I forgot about that," lol. I was thinking basically the same thing as far as splitting into program and handle and was going to take a stab at implementing it but it seemed like it was going to be almost a rewrite and I didn't have the time to do that haha.

Looking forward to it.

nlordell:
I think I got it. Also made it so the default shader is just a PixelShader that gets created when the GraphicsContext gets created. Let me know what you think.

nlordell:
Just some thoughts...
- Do we want to transform the shader source like LÖVE does?
- Do we want to add more "built-in" uniforms and fragment shader inputs to match LÖVE like screen size and screen position?

termhn:
Looks sweet.

1. I guess we can discuss further in the issue based on @icefoxen 's comments, but IMO we should just make people write raw shaders and have a way to easily auto select. Trying to support our own shader language and compile it to several different shader languages sounds like a huge pain for not that much gain to be honest.
2. Hmm, that could be helpful. I'm not sure, I haven't looked into the full list of LOVE's built-in uniforms, but I don't think it would be a bad idea.

icefoxen:
1. Not yet, maybe never.  Too much work for not enough gain.
2. Mmmmmmmaybe but I don't think it's terribly necessary.  Might be useful sometimes but unless we have a use case where it's either super common or super necessary I don't think it's worth the effort yet.

We have an advantage over LÖVE here: gfx-rs lets users define their own pipelines, uniforms, vertex types, etc.  So we already have an extensible way for users to define new data types, shader pipelines, and easily shuffle data from ggez into shaders.  I feel like that makes it unnecessary to add ways to transform/enhance the built-in shaders.  (I might be wrong, but it's also very easy to add later if we decide we want to.)

PR #159, opened 2017-09-19
Send instance transform as matrices encoded in column vectors

termhn:
Last piece of #145. It works, but I haven't done any testing to see if there's any impact on performance. The code definitely is cleaner though, especially in the vertex shader.

### DIFF
--- a/src/graphics/mod.rs
+++ b/src/graphics/mod.rs
@@ -79,11 +79,10 @@ gfx_defines!{
     /// Internal structure containing values that are different for each rect.
     vertex InstanceProperties {
         src: [f32; 4] = "a_Src",
-        dest: [f32; 2] = "a_Dest",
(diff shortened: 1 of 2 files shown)


icefoxen:
Any particular reason you're splitting them into separate vec4's instead of just a mat4?

icefoxen:
Also we should do some performance testing just to see what it looks like.  We need a cool particle system demo anyway...  Hmmm.

termhn:
Instance Buffers don't support mat4's in gfx, only vectors.

icefoxen:
per kvark this is a limitation of the GPU hardware, which I'd just never bumped into before.

icefoxen:
Is this ready to merge?  I never know if one of your PR's is good to go or still WIP...

termhn:
Sorry I should be more clear about that haha, but yeah I think this is ready to merge

PR #165, opened 2017-09-26
Apply mouse input in the input_test example

coder543:
I just thought it was strange the `input_test` looked like it was halfway to letting you move the circle on the X axis, by reading the code, but it didn't actually let you move the circle with the mouse. So, here's a basic commit to add that.


icefoxen:
:D

PR #167, opened 2017-09-27
Issue 162 - graphics_settings example should be better

Manghi:
Hi ggez team,

Please let me know what, if any, changes or improvements need to be made before this is approved. I tried to align it with how ggez is written based on the existing code but if I missed something, or did something completely wrong, please let me know. 

For setting fullscreen, I juggled between what I have presented here and using `set_mode(...)` and ended up deciding on the former since it was a bit simpler. (I didn't have to copy WinMode out of conf... well I don't know of any better way.)

Thanks,
Manghi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
  * Did the same for `get/set_point_size()`
  * Integrated nalgebra to provide point and vector types.
  * Removed inaccurate `timer::sleep_until_next_frame()`, added `timer::yield_now()`.
+ * Added MSAA (thanks termhn!)
+ * Added graphics_settings example to show hot to play with graphics modes
+ * Made the render pipeline just use matrices instead of separate transform elements
 
 # 0.3.3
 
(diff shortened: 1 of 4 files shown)


icefoxen:
Overall looks decent, I want to play with your window mode get/set functions a little before I merge and see if there's a way to make it cleaner.

icefoxen:
Corrections:

The getter functions on `conf::WindowMode` seem redundant, since all the fields in it are public anyway.  If we want to change that we should do it all at once.

I would expect `is_fullscreen` to be bool

apart from that it's all good though! :heart:

Manghi:
Do you think we should add in the getters or keep the fields public? I generally believe that very few fields should be left public but I'd like to know your thoughts.

Doesn't `is_fullscreen` [already return a bool ](https://github.com/ggez/ggez/pull/167/files#diff-3d38872253ac1a673a47a51a3a5c5517R939)or am I missing something?

icefoxen:
> Do you think we should add in the getters or keep the fields public?

That's... really not a terrible question in general that I don't have an off-the-cuff answer to.

For `Conf` and its related types, everything exists explicitly to be serialized to and from text anyway, so exposing them as public fields in plain-ol-data seems to be the minimum friction path.  I'd really prefer it if Rust had properties, but, we gotta work with what we have currently.

I think for now we should leave the variables public instead of using getters/setters.  If we decide to use getters/setters it should happen everywhere at once.  Feel free to open a separate issue about this.  :-)

Issue #161, opened 2017-09-20
Stroked circles are not closed?


icefoxen:
See `examples/graphics_settings.rs` in commit 0c5ee550f59f6cea60260810eee1796a7b1742fd

Probably a lyon bug; verify and punt upstream.

icefoxen:
`graphics::circle(ctx, DrawMode::Line(3.0), Point2::new(400.0, 300.0), 100.0, 4.0)?;` results in this:

![image](https://user-images.githubusercontent.com/1335133/30786671-b30d0d38-a147-11e7-82cc-83474c89d2e7.png)

`graphics::circle()` is:

```rust
pub fn circle(ctx: &mut Context,
              mode: DrawMode,
              point: Point2,
              radius: f32,
              tolerance: f32)
              -> GameResult<()> {
    let m = Mesh::new_circle(ctx, mode, point, radius, tolerance)?;
    m.draw(ctx, Point2::origin(), 0.0)
}
```

`Mesh::new_circle()` is:

```rust
    pub fn new_circle(ctx: &mut Context,
                      mode: DrawMode,
                      point: Point2,
                      radius: f32,
                      tolerance: f32)
                      -> GameResult<Mesh> {
        let mut mb = MeshBuilder::new();
        mb.circle(mode, point, radius, tolerance);
        mb.build(ctx)
    }
```

`MeshBuilder::circle()` is:

```rust
    pub fn circle(&mut self,
                  mode: DrawMode,
                  point: Point2,
                  radius: f32,
                  tolerance: f32)
                  -> &mut Self {
        {
            let buffers = &mut self.buffer;
            match mode {
                DrawMode::Fill => {
                    // These builders have to be in separate match arms 'cause they're actually
                    // different types; one is GeometryBuilder<StrokeVertex> and the other is
                    // GeometryBuilder<FillVertex>
                    let builder = &mut t::BuffersBuilder::new(buffers, VertexBuilder);
                    t::basic_shapes::fill_circle(t::math::point(point.x, point.y),
                                                 radius,
                                                 tolerance,
                                                 builder);
                }
                DrawMode::Line(line_width) => {
                    let builder = &mut t::BuffersBuilder::new(buffers, VertexBuilder);
                    let options = t::StrokeOptions::default()
                        .with_line_width(line_width)
                        .with_tolerance(tolerance);
                    t::basic_shapes::stroke_circle(t::math::point(point.x, point.y),
                                                   radius,
                                                   &options,
                                                   builder);
                }
[...]

PR #169, opened 2017-09-27
Update to lyon 0.8.

nical:
Lyon 0.8 is fresh out with bug fixes, performance improvements and newly supported bevel line joins (although ggez doesn't appear to expose stroke options).

There were few enough API changes that it looks like none of them affected ggez (I did `cargo build --all` and `cargo build --examples`), so the PR is disappointingly simple (unless I built the wrong branch or something).

This fixes issue #161.


icefoxen:
I just did this anyway.  :D

Issue #172, opened 2017-10-03
Support gradients?


TatriX:
Is it possible to make a gradient?

termhn:
I don't believe there's an easy helper function to do it. If you need it to be dynamically generated, you can do that with the current master version by writing and applying your own `PixelShader` (an OpenGL fragment shader that gets injected into the drawing pipeline). If you don't need it to be dynamic, then probably the easiest way is to just make a texture with a gradient and draw that.

TatriX:
Can I make something like [this](https://love2d.org/wiki/Gradients)?

termhn:
Yes. Untested, but something like this would give you the same result. You'll then have to scale it by however much you want to actually get the gradient to be the size you want or create a function similar to the `drawinrect()` in that love documentation to do it for you.

```rust
enum GradientDirection {
    Horizontal,
    Vertical
}

fn gradient(ctx: &mut Context, colors: &[graphics::Color], direction: GradientDirection) -> graphics::Image {
    let buf_size = colors.len();
    let mut buffer = Vec::with_capacity(buf_size);
    for color in colors {
        buffer.extend(color.into().iter());
    }
    let mut result;
    match direction {
        GradientDirection::Horizontal => {
            result = graphics::Image::from_rgba8(ctx, buf_size as u16, 1, &buffer);
            result.setFilter(graphics::FilterMode::Linear);
        },
        GradientDirection::Vertical => {
            result = graphics::Image::from_rgba8(ctx, 1, buf_size as u16, &buffer);
            result.setFilter(graphics::FilterMode::Linear);
        }
    }
    result
}
```

TatriX:
Cool, thanks!

PR #178, opened 2017-10-04
Gradients

termhn:
Based on #172.

Implements a gradient creation helper function on Image and corresponding `draw_in_rect()` helper function with new `RectDraw` trait. Also gives an example in the `drawing` example.

### DIFF
--- a/examples/drawing.rs
+++ b/examples/drawing.rs
@@ -12,6 +12,8 @@ struct MainState {
     image1: graphics::Image,
     image2_linear: graphics::Image,
     image2_nearest: graphics::Image,
+    gradient: graphics::Image,
+    gradient_rect: graphics::Rect,
     zoomlevel: f32,
 }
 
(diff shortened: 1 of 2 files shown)


TatriX:
Is it possible to extend your solution to support, say, radial gradient without breaking the API?

termhn:
Uhhhh not using this same techinque, no, I don't think so.

TatriX:
Convenience over generality... Probably this should be added as an extension and not as a core feature.

TatriX:
Probably solution like this could be a good choice, but I don't know is it possible with current state of the ggez.
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createLinearGradient

termhn:
Perhaps. This is a direct adaptation of the way that LOVE does it, which is to create either a vertical or horizontal strip of pixels with each color, set linear filtering, and then scale it and let the graphics backend's rasterizer take care of actually creating the gradient because of the way it does linear filtering of textures. If we wanted to implement an api like Canvas has (or add a radial gradient) we would need to write those algorithms ourselves.

I dunno, thoughts @icefoxen? LOVE is a heavy inspiration of ggez, but do we want to adopt all their decisions? I don't really have a strong opinion one way or another other than to say properly supporting gradients would be a fair amount of work for a fairly small gain IMO. The amount of use you have for wanting to dynamically generate a gradient not inside a shader seems small to me; if you need more precision you should probably be creating a texture in Photoshop or another graphics package of your choice.

TatriX:
True. I ended up simply using an image.

icefoxen:
It seems like the Right Solution would be to properly support color attributes for vertices (something I've been putting off for a little while).  Then gradients would just be a matter of setting the vertex colors correctly, and you could make builder functions that create rect's with the right vertex colors.

Gradients alone don't seem a common enough use to justify the special case, especially when the general case isn't much harder.  Sorry.

termhn:
Closing this for now based on above.

PR #180, opened 2017-10-05
Added Canvas support for off-screen rendering

nlordell:
Added support for off-screen renderering. Combined with PixelShaders feature you can start making some pretty cool effects. I also added an example that combines Canvas with PixelShader to do some cool dynamic shadow effects:

shadows


icefoxen:
Oh my goodness.

I will confess I don't actually know how to do this myself, so the amount of actual review I can do might be limited, but I guess I should take this as a learning experience.  :heart:

nlordell:
I basically found this and used it as a reference, adapting it to ggez -
https://github.com/mattdesl/lwjgl-basics/wiki/2D-Pixel-Perfect-Shadows

I'm not sure if this is the best way to do it or not but I thought it was a neat way to show off what off screen rendering with Canvas and the shaders feature can do.

I went back and tried to add some comments so that things make more sense.

termhn:
This is really cool. I submitted a PR to your branch that makes the shadow demo a bit more aesthetically pleasing by hacking some things. To really do this accurately I'd need to use different blending modes for adding together the different render targets, but for now, this basically fakes it.

Also, it seems that the new render targets created by canvases are not affected by multisampling settings. I think we should either have it be affected by the context's and/or have it be taken as an input into the `create_canvas()` function how many samples you want the target to have.

nlordell:
I think for the multi-sampling we should probably read from the Conf.window_mode.samples from the context when creating the texture.

nlordell:
New version looks super smooth, thanks @termhn!

shadows

nlordell:
So I looket at the `Factory::create_render_target()` implementation and it does indeed create a texture without multisampling:

```
    fn create_render_target<T: format::RenderFormat + format::TextureFormat>
                           (&mut self, width: texture::Size, height: texture::Size)
                            -> Result<(handle::Texture<R, T::Surface>,
                                       handle::ShaderResourceView<R, T::View>,
                                       handle::RenderTargetView<R, T>
                                ), CombinedError>
    {
        let kind = texture::Kind::D2(width, height, texture::AaMode::Single);
        let levels = 1;
        let cty = <T::Channel as format::ChannelTyped>::get_channel_type();
        let tex = try!(self.create_texture(kind, levels, SHADER_RESOURCE | RENDER_TARGET, Usage::Data, Some(cty)));
        let resource = try!(self.view_texture_as_shader_resource::<T>(&tex, (0, levels-1), format::Swizzle::new()));
        let target = try!(self.view_texture_as_render_target(&tex, 0, None));
        Ok((tex, resource, target))
    }
```

So fixing this is fairly easy and just involes implementing our own version of `create_render_target` with multisamping.

sanpii:
I have a strange behaviour with the `render_to_image` example: unitialized zones and blinking (in fact, the image blink quickly).
canvas

The shadows example works fine 👍

nlordell:
Hmm, looking at the code there is an extra `graphics::present()` call that shouldn't be there. Maybe that is the problem. Let me know if that fixes it

sanpii:
@nlordell perfect, thank you.

sanpii:
I think it may be useful to add `width()` and `height()` methods to `Canvas`.

nlordell:
@sanpii You can currently get the width and height by getting the `Canvas`'s `Image`. Something like this:
```
let canvas = Canvas::new(...);
let width = canvas.get_image().width();
```

We could add the methods directly to `Canvas` instead, I'm not sure what is best and will differ the decision to someone else :)

PR #185, opened 2017-10-08
Blend modes!

termhn:
An easy to use API (basically the same as LOVE's) to switch the blend mode of the current shader program. The main limitation is that it requires you to provide a list of blend modes that you want your shader program to support when you create it (or `None` to apply only the default Alpha blend mode). The default shader program comes with all of them by default. This allows us to do things like this :D (the updated shadows example).

image

TODO:
* ~~Still need to figure out how to implement the Lighten and Darken blend modes using gfx's API--for now they're just placeholders. All the rest of the blend modes work, though.~~
* ~~Perhaps restructure where a few things live; I'm not sure it makes sense to have the `graphics::set_blend_mode()` function inside `pixelshader.rs`~~

### DIFF
--- a/examples/shadows.rs
+++ b/examples/shadows.rs
@@ -362,9 +362,7 @@ impl event::EventHandler for MainState {
         self.render_light(ctx, light, center, canvascenter)?;
 
         // Now lets finally render to screen starting with out background, then
-        // the shadows and lights overtop and finally our foreground. Note that we set the
-        // light color as the color for our render giving everything the "tint"
-        // we desire.
+        // the shadows and lights overtop and finally our foreground.
         graphics::set_canvas(ctx, None);
         graphics::set_color(ctx, graphics::WHITE)?;
         graphics::set_blend_mode(ctx, BlendMode::Alpha)?;
--- a/examples/shadows.rs
+++ b/examples/shadows.rs
@@ -217,8 +217,12 @@ impl MainState {
         };
         let foreground = Canvas::with_window_size(ctx)?;
         let occlusions = Canvas::new(ctx, LIGHT_RAY_COUNT, 1, conf::NumSamples::One)?;
-        let shadows = Canvas::with_window_size(ctx)?;
-        let lights = Canvas::with_window_size(ctx)?;
+        let mut shadows = Canvas::with_window_size(ctx)?;
+        // The shadow map will be drawn on top using the mutiply blend mode
+        shadows.set_blend_mode(Some(BlendMode::Multiply));
+        let mut lights = Canvas::with_window_size(ctx)?;
+        // The light map will be drawn on top using the add blend mode
+        lights.set_blend_mode(Some(BlendMode::Add));
         let occlusions_shader =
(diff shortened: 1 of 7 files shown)


icefoxen:
Nice.  Is it possible to use this to set blend modes for arbitrary images without custom shaders being involved as well?  I vaguely recall doing something for that but don't remember what it looks like.

termhn:
Hmm... you could just switch the blend mode, draw the image, then switch it back to the default blend mode, but I don't think it's possible (at least with this) to have an image have a different blend mode. From what I've gathered from kvark, the blend mode is an immutable piece of the gfx PSO, and so to change blend modes we create multiple PSOs (one for each blend mode) and then have it use the correct one based on the requested blend mode. We could add a piece of state on the ggez `Image` type that would automatically attempt to switch the blend mode when it is `draw`n, though.

I've made the default shader program have PSOs for all of the blend modes, so you can set any blend mode when you have the default shader active. For user shaders, the user has to specify all the blend modes that they want to be able to use with that shader.

nlordell:
This is super cool!

termhn:
Alright, all the TODOs are now implemented (hopefully) and I experimentally added in the ability for you to individually set the blend mode of any Drawable, which will override the general blend mode that is set, based on @icefoxen 's comment.

termhn:
@nlordell Can you test this on your macbook? I hadn't until now and it seems as though something is being messed up with the new high dpi stuff on mine. I probably messed something up when rebasing it and didn't notice until now heh. 

@icefoxen I think this will be ready to merge once I get the above-mentioned bug fixed

icefoxen:
This looks generally fantastic!  I just need to grok what the Pso structs are doing more fully.

termhn:
Basically, in gfx, because Vulkan and DX are more strict about how blend modes work than GL is, blend modes are a baked in piece of state for a PSO and you can't change it dynamically. After chatting with @kvark on IRC and looking how he does it in three-rs, the best way to change blend modes is to just make multiple PSOs with a blend mode baked in... this is fine for when you only have one shader program, but since shaders are also baked into a PSO, if you want to have multiple shaders which each support all the blend modes, you also have to make a PSO per blend mode per shader.

The `PsoSet` struct is basically just a hash map for easily storing each shader set's PSOs and then retrieving them based on a BlendMode.

Since creating all the blend mode PSOs for multiple shaders would be a bit ridiculous when you probably aren't going to use all of them, for custom shaders you have to specify when you create the shader a list of BlendModes that you want to be able to use with that shader. If you supply `None` then it will only create one PSO with the default `Alpha` blend mode.

icefoxen:
Makes sense.  Sounds, as I suspected from the placement and organization of the code, rather similar to the `SamplerCache`.  But thanks for the excellent explanation.  ...can you add it as a comment to the source?  :-P

termhn:
Yes, it's basically a modified `SamplerCache`, ahah. And yeah, I can!

termhn:
Done! Ready to merge if you are I think.

PR #193, opened 2017-10-17
make sure when rendering the light and shadow canvases we take DPI into account

nlordell:
@termhn Finally got around to testing the shadows example on my mac and there was indeed an issue with it when high DPI is enabled, this pull request should fix that.

Also there it seems to be only running at 30FPS on my mac and I'm not sure why... I'm going to investigate a bit and see if I can it out.


termhn:
Sweet, works on mine as well now. I'm getting about 55 fps on mine, which *is* less than I'd expect I think... I'd have to profile to figure out where the slowdown is happening though.

PR #196, opened 2017-10-28
Typo fix

osa1:


src/graphics/mod.rs
```diff
@@ -861,7 +861,7 @@ pub fn set_default_filter(ctx: &mut Context, mode: FilterMode) {
 /// Recall that a `Rect` currently the x and y coordinates at the center,
 /// so if you wanted a coordinate system from (0,0) at the bottom-left
 /// to (640, 480) at the top-right, you would call this function with
-/// a `Rect{x: 320, y: 240, w: 640, h: 480}`
+/// a `Rect{x: 320f32, y: 240f32, w: 640f32, h: 480f32}`
```
michaelbachner: I don't know the style guide of this project, but I think that `Rect{x: 320.0, y: 240.0, w: 640.0, h: 480.0}` tends to be better readable. 

Or do they need to be `f32`?

icefoxen: 320.0 etc would be better, yes.

Issue #204, opened 2017-11-13
Can I add some methods on Rect like .contains(point), .overlaps(rect), etc...?


Zac-Garby:
I think it'd be really useful to have this functionality, since it's such a common use of rectangles. Some ideas of methods I had were:

 - `.contains(&self, &Point) -> bool` If a rectangle contains a point.
 - `.overlaps(&self, &Rect) -> bool` If a rectangle overlaps another rectangle.
 - `.fits(&self, &Rect) -> bool` If another rectangle can fit inside this one.

Also, some other methods for translation/scaling which _could_ be done using operator overloads, or just normal methods:

 - `.translate(&self, x: f32, y: f32)` Translates the rectangle by (x, y). Could be `rect + (x, y)`
 - `.move_to(&self, x: f32, y: f32)` Moves the rectangle to (x, y)
 - `.scale(&self, sx: f32, sy: f32)` Scales the rectangles dimensions by (sx, sy). Could be `rect * (sx, sy)`
 - `.scale_centered(&self, sx: f32, sy: f32)` Like the last one, but scales about the middle of the rectangle instead of the origin
 - `.set_center(&self, x: f32, y: f32)` Sets the middle position of the rectangle to (x, y)

Zac-Garby:
EDIT: Just realised the coordinates of a rectangle are in fact the center, so the last 2 methods would be redundant.

Manghi:
Might I add 'intersects' as well? This was [present in ggez 0.2 ](https://docs.rs/ggez/0.2.2/ggez/graphics/struct.Rect.html#method.intersection)but is not in ggez 0.3+. It would be nice to have this back.

```Rust
fn has_intersection(&self, other: Rect) -> bool
Determine whether two rectangles intersect.
```
```Rust
fn intersection(&self, other: Rect) -> Option<Rect>
Calculate the intersection of two rectangles. The bitwise AND operator & can also be used.
```
```Rust
fn intersect_line(&self, start: Point, end: Point) -> Option<(Point, Point)>
Calculates the intersection of a rectangle and a line segment and returns the points of their intersection.
```

Zac-Garby:
Yeah, good point - although `has_intersection` is the same thing as `overlaps`.

icefoxen:
[...]

PR #205, opened 2017-11-14
Add some useful Rect methods

Zac-Garby:
They're not tested, but they _should_ work. Should I add some tests?

#204


svenstaro:
Definitely add some tests.

Zac-Garby:
Done

PR #213, opened 2017-11-29
Fix bounds check in Spritebatch::set

katis:


svenstaro:
That we had this wrong is indicative of bad test coverage. Can you make a test that shows this failure?

icefoxen:
Yeah, our test coverage is basically garbage right now.  :-(  I don't know a good way to make most of it more comprehensive, since we can't create graphics contexts on the CI servers.

svenstaro:
Or maybe we can. I will research that. I made something similar work once. What I'm thinking right now is to self-host drone ci and set it up on my server.

onelson:
Yeah, this is tough. I've been trying to make clear divisions in my code between the stuff that is doing calculations used for drawing vs stuff that actually needs a render surface, then I can thread the needle and test each separately. It's not always convenient or achievable.

icefoxen:
Part of the problem is ggez's everything-bundled-together-in-one-`Context` model makes it harder to pick things apart and test them individually.  I kind of want to break that up a little bit  -- LÖVE splits subsystems into modules, so you can make a LÖVE program that just doesn't initialize the graphics and sound modules for example.  Doing that in Rust seems like it would involve lots of messy "is this subsystem initialized?  Can I do stuff with it?" code, but it may be worth it anyway.

Then there's just the problem that, say, figuring out whether a drawing function is working correctly or not from a unit test is hard and I can't think of a good way of doing it.  :-(

IIRC I played with having Travis install X packages and try to test graphics stuff, but it failed because it still didn't have a running X server or usable GPU drivers.

onelson:
Yeah, I was thinking this sort of thing could be handled with "features" in an idiomatic way.

aochagavia:
Can this get merged in the meantime? :)

icefoxen:
Yeah, sorry for the delay.  I wanted to write up a proper unit test for it before merging and haven't had time.  x_X

PR #214, opened 2017-12-04
Fix docs version in README

JosefKuchar:


icefoxen:
Good catch, thank you!

Issue #216, opened 2017-12-06
Using multiple colors in a SpriteBatch


17cupsofcoffee:
Is is possible (or might it be possible in future) to set the drawing color on a per-sprite basis with the new sprite-batching functionality? I'm guessing not, since `DrawParam` doesn't take a `color` parameter, but there might be something I'm missing.

Context: I'm working on a console-style renderer for a roguelike, which involves me drawing individual characters from a greyscale spritesheet - that seems like low-hanging fruit to combine into a single draw call, but I'm using `graphics::set_color` to tint the characters at the minute, so I think I might be out of luck.

icefoxen:
ggez sets colors basically in a shader global per draw call right now, and not per-vertex, which is honestly a little atypical.  I have mixed feelings about this style of drawing; it's mostly there 'cause it's inherited from LÖVE.

Adding a `color` field to `graphics::InstanceProperties` would be straightforward.  Then you'd want to remove the existing color field from `graphics::Globals`, and probably remove `graphics::set_color()` as well; having two "color" state variables seems like asking for trouble.  I don't *object* to such an API change (honestly I would prefer to have less implicit state rather than more) but am hesitant.  Looking at <https://love2d.org/wiki/SpriteBatch:setColor> suggests that they basically use two color state variables and it has not always been a good experience for them.  :-P

17cupsofcoffee:
Thanks for the explanation! It's not a massive issue for me if things stay as they are (my game is turn-based/not particularly flashy, so being slightly inefficient redrawing the map probably isn't the end of the world...), but I may have a go at writing a pull request if I get time, to see how well it'd work.

icefoxen:
Hmmm, upon consideration, using two state variables may not be the *worst* thing ever... this is basically what we do for filter modes already.  There's a global default that everything that gets created uses, and then you can override it on a per-object basis.

That said, you generally don't want to change the filter mode for an image after creating said image, so.

The more I think about it the less I like the global color state, but I also *usually* want to go through life not forcing everyone to supply a color parameter to every draw call ever.  Perhaps we should have it set per object, since that's generally how graphics cards end up using it anyway in most API's?  That still doesn't work cleanly with sprite batches though, since they all draw from the same object.  So we'd need to be able to specify draw color per object *anyway*.

[...]

PR #218, opened 2017-12-06
Move color settings into InstanceProperties/DrawParam

17cupsofcoffee:
As discussed in #216!

This PR takes a first stab at removing some of the implicit state around drawing things with a color, removing `set_color` and adding a `color` field to `InstanceProperties` and `DrawParam`.

Fully anticipating that this might not end up going anywhere (not entirely sold on the API as is), but it might give people smarter than me an idea of where to go next. Plus if I'm honest, I just wanted an excuse to ferret around in the codebase :p

Couple of observations:
* I am wholly unfamiliar with GFX and OpenGL, so apologies if I've made any rookie errors with the shaders!
* I'm not sure how I feel about having to use `draw_ex` every time I want to draw something with a color 🤔 
* Some thought would need to be given as to how this would interact with the primitive drawing API (`circle`, `line`, etc.)
  * This is why I haven't updated the examples yet - everything's in white!

### DIFF
--- a/src/graphics/mod.rs
+++ b/src/graphics/mod.rs
@@ -133,12 +133,12 @@ gfx_defines!{
         col2: [f32; 4] = "a_TCol2",
         col3: [f32; 4] = "a_TCol3",
         col4: [f32; 4] = "a_TCol4",
+        color: [f32; 4] = "a_Color",
     }
 
     /// Internal structure containing global shader state.
     constant Globals {
         mvp_matrix: [[f32; 4]; 4] = "u_MVP",
-        color: [f32; 4] = "u_Color",
     }
 
     pipeline pipe {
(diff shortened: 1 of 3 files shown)


icefoxen:
Looks reasonable, though I'm still not sure if this is an API change we *should* make (yet?).  🤔

The mesh drawing code shouldn't need any changes to have this work, I'll test it out.

17cupsofcoffee:
Yeah, I'm not entirely sold on it either - I'm under no illusions that I know better than the LOVE devs do :p

And the mesh code itself shouldn't need changes, but as far as I could tell there's no way to set the DrawParam when using `graphics::rectangle` and the like, so they're only able to use the default - in this case, that makes everything draw in white.

icefoxen:
Yeah you'd have to make a `Mesh` object explicitly with a `MeshBuilder` or such and call `draw_ex()` on that.

icefoxen:
Closing this since we've found a more general solution that seems to work, see #216 and associated commits.

PR #246, opened 2018-01-02
Update gfx to 0.17 and sdl to 0.31

DenialAdams:
I did this in hopes that it would fix the screenshot procedure I transplanted from the gfx example. It did not :) but hopefully this will save you a couple of minutes of effort later.

Side note: what happened to the rust-gamedev IRC channel? :(


src/graphics/mod.rs
```diff
@@ -1301,7 +1301,7 @@ impl Image {
             return Err(GameError::ResourceLoadError(msg));
         }
         let kind = gfx::texture::Kind::D2(width, height, gfx::texture::AaMode::Single);
-        let (_, view) = factory.create_texture_immutable_u8::<gfx::format::Srgba8>(kind, &[rgba])?;
+        let (_, view) = factory.create_texture_immutable_u8::<gfx::format::Srgba8>(kind, gfx::texture::Mipmap::Provided, &[rgba])?;
```
DenialAdams: I know jack about Mipmaps. I went with the Provided variant instead of Allocated becausue that is what the older instances in gfx switched to when this landed: https://github.com/gfx-rs/gfx/pull/1470

Let me know if that checks out, examples seem OK

icefoxen:
Thanks, I've been meaning to do this. I should find time to review this properly sometime this week.

src/graphics/mod.rs
icefoxen: That's fine.  Mipmaps are basically multiple versions of a texture resampled at different resolutions.  `Provided` says "I have given you all the mipmaps you need", the alternative is `Allocated` which says "I'm going to generate mipmaps later so allocate space for them".

DenialAdams: Thanks for the explanation! I often feel that the gfx are really lacking :/

icefoxen: I'm actually going to be stepping back from ggez a little partially so I have time to work on gfx docs, so.

DenialAdams: Awesome! Your wiki pages are great, so I'm sure the docs will substantially improve :)

PR #248, opened 2018-01-04
Remove commented-out code from astroblasto

neganp:
This commit removes a few commented-out (deprecated?) lines of code
from `examples/astroblasto.rs`.


icefoxen:
Derp, thank you!

PR #249, opened 2018-01-05
Screenshot

### LINKED ISSUE #244: Present canvas / image to file
It'd be neat if I could render a canvas, and then output that (or perhaps the underlying Image?) to a file. This would be nice for screenshots, but also for other stuff like rendering my world in full size so I can see it all at once rather than drawing to the size of the viewport.

(sorry if this is doable already, I couldn't figure out how!)

DenialAdams:
Basically taken verbatim from the gfx example.

Works fine with canvases too (once I adjusted the viewport accordingly.)

It's a bit slow - you can see my note about instantiating the download buffer on the fly. I imagine acting on that may speed things up, but ensuring that it is always the correct size everywhere in the codebase looked like a pain. I can try to take a shot at it if you're interested.

Reversing the images is a bit weird, maybe there's a better way to handle that.

Closes #244

### DIFF
--- a/examples/astroblasto.rs
+++ b/examples/astroblasto.rs
@@ -576,7 +576,10 @@ impl EventHandler for MainState {
             }
             Keycode::Space => {
                 self.input.fire = true;
-            }
+            },
+            Keycode::P => {
+                graphics::screenshot(ctx, "screenshot.png");
+            },
             Keycode::Escape => ctx.quit().unwrap(),
             _ => (), // Do nothing
         }
--- a/examples/astroblasto.rs
+++ b/examples/astroblasto.rs
@@ -578,7 +578,8 @@ impl EventHandler for MainState {
                 self.input.fire = true;
             },
             Keycode::P => {
-                graphics::screenshot(ctx, "screenshot.png");
+                // Because this is an example, we silently discard any error while screenshotting
+                let _ = graphics::screenshot(ctx, "screenshot.png");
             },
             Keycode::Escape => ctx.quit().unwrap(),
             _ => (), // Do nothing
--- a/src/error.rs
+++ b/src/error.rs
@@ -201,6 +201,20 @@ impl From<gfx::PipelineStateError<std::string::String>> for GameError {
     }
 }
 
+impl From<gfx::mapping::Error> for GameError {
+    fn from(e: gfx::mapping::Error) -> GameError {
+        let errstr = format!("Buffer mapping error: {:?}", e);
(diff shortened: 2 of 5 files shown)


DenialAdams:
Oh, pretty pictures. One from astroblastos and two from my game, since I'm happy I got it working and want to show off :P
astro
ss
level

icefoxen:
Great work!  But I'm not going to merge it quite yet because I want to play with it and see if we can make it a bit nicer.  I feel like we should be able to break it down into pieces a little: Maybe a function to fetch the current `Canvas` into an `Image` (even if it's the screen render target), a function to turn an `Image` into an RGBA `Vec<u8>` or such, and a convenience function around `image::save_buffer()` to actually save the data... ideally to the save into the user's save game directory or such.  (This might actually matter in the future for platforms like android, 'cause the game can't assume it has write permission everywhere.)

Also since it calls `gfx.encoder.flush()` in the middle of this function that might lead to surprising effects... if you draw some stuff, call `screenshot()`, then draw more stuff and call `graphics::present()`, I don't think what you draw before calling `screenshot()` will show up.  It should be easy to work around this by creating an `Encoder` just for this function; it's just a vec and some metadata, so not a huge deal.  I'm not super worried about speed anyway, at least right now, though I'm a little suspicious of adding `Bind::TRANSFER_SRC` to the `Canvas` texture because I don't know exactly what difference that makes.

Again though... great work, and thank you!  You've done the hard part.

DenialAdams:
Thanks for the feedback!

I'll work on this sometime soon, though it might be a few days.

> Bind::TRANSFER_SRC

Yeah, I meant to flag that - I wasn't sure what the impact would be of that. But it was necessary to not panic when screenshotting a canvas. Unfortunate that every canvas has to pay that price though

icefoxen:
Okay, checked with the gfx devs and `TRANSFER_SRC` has no effect on the OpenGL backend, it's only used for some checking at the gfx level.

DenialAdams:
Rebased and started to dig into some of the things you mentioned.

Sorry for the long delay on this, school starting up took a lot more out of me than I thought!

DenialAdams:
Took a stab at saving to the game's config directory, as well as using a fresh encoder so we don't trample the existing encoder state.

icefoxen:
Very nice!  Sorry for my own delay; I keep meaning to fiddle with this and merge it (and the other PR's) and my weekends keep getting devoured by other things.

It looks good, I'll have to eyeball it one more time when I'm actually awake.

Issue #245, opened 2018-01-02
Playing the same sound several times appends instead of playing simultaneously


jupart:
Currently, ggez::audio::Source.play() uses rodio::Sink.append(), which tacks the sound onto the end of the currently playing sound instance (when playing the same Source). To me, play() should probably play the new sound immediately, and not care about any other instances of the sound playing.

Similar rodio issue with possible solution: https://github.com/tomaka/rodio/issues/141

icefoxen:
As discussed on IRC: Storing a SoundData and creating Source's from it as necessary should work, but you have to keep a hold of the Source until it finishes or else it will stop when it's dropped...  ggez should be able to keep track of this for you and GC Source's when they're done playing.

Cloning a `Source` also doesn't seem to work 'cause it contains a `rodio::Sink`, which isn't `Clone`.  Honestly we might be better off ripping up the sound API some and implementing something a bit more intuitive based on some of `rodio`'s newer stuff, like `DynamicMixer`.

icefoxen:
Also see #208, #75

jupart:
What do you think about letting `ggez::audio::Source` have a queue of `rodio::Sink`'s it can use, say `Vec<rodio::Sink; 4>` (everywhere I look folks seem to allow 4 simultaneous identical sounds: Kivent, Amethyst). When you call `Source.play()`, it'll grab an `.empty()` sink if there is one. Not sure what to do if there isn't one, maybe just ignore the `play()` call. 

We'll lose a little control over `Source.pause()`, `.stop()`, `.resume()`, and whatever else, but we could provide `.stop_all()`, `.stop_last_played()`, etc. which I think would be good enough for most use cases.

As a user, you could either use a stored container of `Source`'s or create a new `Source` for every thing (entity) that needs it.

Additional stuff (should probably be in a separate issue). I'd like to add:
* `SoundData::from_file()` for loading up `SoundData`'s to build `Source`'s as needed since `Source` doesn't impl Clone/Copy
* `SpatialSource` which uses `rodio::SpatialSink` and related set/getters or make `Source` use `SpatialSink` all the time, with defaults that make it act like a regular `Sink`
* `Source.repeat(bool)`

Thoughts/plans for later:
*  some way to create two loop points (loop point and point to return to after loop) on a `Source`, maybe using `rodio::Source.periodic_access()` or `.take_duration()`with a 
* `sink.empty()` callback

jupart:
[...]

PR #250, opened 2018-01-08
Sound development - SoundData.from_file(), Source.repeat

jupart:
This PR implements part of what is being discussed in https://github.com/ggez/ggez/issues/245. The second commit changes the way Source.play() works, ignoring the call if the sound is already playing, never simply appending more sound to the end of the rodio::Sink, which I thought was unintuitive as a user.

### DIFF
--- a/examples/astroblasto.rs
+++ b/examples/astroblasto.rs
@@ -245,8 +245,8 @@ impl Assets {
         let rock_image = graphics::Image::new(ctx, "/rock.png")?;
         let font = graphics::Font::new(ctx, "/DejaVuSerif.ttf", 18)?;
 
-        let shot_sound = audio::Source::new(ctx, "/pew.ogg")?;
-        let hit_sound = audio::Source::new(ctx, "/boom.ogg")?;
+        let shot_sound = audio::Source::new(ctx, "/pew.ogg", false)?;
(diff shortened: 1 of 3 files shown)


icefoxen:
Thank you!  I'll try to find time to look over this in detail soon.

jupart:
Sure! TODO:
- [ ] Update examples
- [ ] Update failing tests
- [ ] Add tests for new `fn`'s

icefoxen:
Ugh, sorry for the delay looking at this.  Especially when the conclusion is, we can't take this for a 0.4.1 release because it changes a public API in a non-backwards-compatible way: adding a `repeat` argument to `Source::new()` will break existing code.

Possible solutions: Make a `SourceBuilder` that has the `repeat()` property, or otherwise find some other place to put the `repeat` argument.

This is still gonna get pulled but not into the 0.4 branch.

jupart:
You're right, I can change `Source::new()` to match existing API and the user can just use `Source::to_repeat()`after it's created. Is that OK with you?

icefoxen:
That would be great, thank you!

icefoxen:
@jupart Finally got around to merging this, sorry it's taken so long.  I renamed things a little in an attempt to keep the API consistent, see https://github.com/ggez/ggez/commit/b497e3e1dd79d65531e23a5d98847ae0e56b3289

Thank you again!  :heart:

PR #255, opened 2018-01-11
remove timer::sleep_until_next_frame from docs

rardiol:
is this actually correct?


icefoxen:
Thank you!  Guess I missed that one.  For future reference though, updates should be made to the `devel` branch; see CONTRIBUTING.md for details.

Issue #234, opened 2017-12-22
Known rough spots


icefoxen:
These are just the areas that don't get banged on a lot in my own usage, and so might be sub-par.

 * audio -- https://github.com/ggez/ggez/issues/75
 * mouse and gamepad (not many tests or examples for those)
 * ~~breaking up `graphics/mod.rs` might be nice; it's 1300 lines.~~ Done in #256 
 * unit tests! #380
 * ~~Bitmap fonts~~ Refactor into separate crate.
 * ~~Filesystem -- could be refactored to top-level functions like the rest of the API instead of methods on the `Filesystem`~~ Done in b85a8d0eecf6a0f191028b7a6c2aea70d02447f0
 * ~~Clean up the `error` module... use `failure`?~~

CaptainKraft:
I tried to switch to a bitmap font just before submitting a jam game, and didn't figure it out fast enough, so I resorted back to ttf fonts.

It seems like there is some minimum requirement for the characters required in the font string?

icefoxen:
Shouldn't be any requirement besides them being fixed width in a horizontal line.  Were you using master or devel branch?

CaptainKraft:
It looks like I have 0.4 in my Cargo.toml.

I would get an error when I passed in a string that was just capital letters. It would complain about missing characters.

icefoxen:
Does your bitmap font have capital letters in it?  Are you passing it a description string that includes capital letters?

I mean, I assume so, but I gotta make sure.  Having problems with *only* capital letters is strange.  What error did you get?

CaptainKraft:
I passed in " ABCDEFGHIJKLMNOPQRSTUVWXYZ" and got this error:

`Error encountered: GameError FontError("Character \'i\' not in bitmap font!")`

This is the file I'm trying to use:

![image](https://user-images.githubusercontent.com/1237356/37347073-a283f2b8-2696-11e8-90f0-05239fc86a6b.png)

There is a space as the first character and I inverted the colors so you can see it on this white background.

icefoxen:
Sounds like the error message could be better then!  :D

icefoxen:
Thoughts on unit tests: We should be able to have two sets of unit tests.  One that requires a `Context` to be created and so can't be run terribly well on headless servers, and one that tests things that don't need graphics/etc.  We'd be able to run the former set of tests at least locally.

icefoxen:
[...]

PR #256, opened 2018-01-12
Break up graphics mod

codingcampbell:
Big fan of this library and the end-goal of reaching LOVE-levels of API happiness. So taking a quick stab of one of the chores at #234

New-ish to Rust here, feel free to throw this PR in the trash if it's doing something terrible. I'm particularly unsure about the `pub(crate)` stuff .. but introducing files made Rust complain a _lot_ about private fields, and `pub(crate)` _seems_ to keep it the same as if everything shared the same file.

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -1,0 +1,366 @@
+use std::cell::RefCell;
+use std::rc::Rc;
+
+use gfx_device_gl;
+use gfx_window_sdl;
+use gfx::traits::FactoryExt;
+use gfx::Factory;
+
+use graphics::*;
+use conf::WindowSetup;
+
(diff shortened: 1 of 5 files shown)


icefoxen:
Wow, thanks a lot!  `pub(crate)` should be just fine.  There might be a slightly simpler way of doing it but if so I don't know what it is.

DenialAdams:
>  There might be a slightly simpler way of doing it but if so I don't know what it is.

Why not an `Image` module?

icefoxen:
Seems like that's what we have now? Not sure what you mean.

DenialAdams:
I meant a new (sub)module but nevermind, I was totally confused and that
wouldn't help at all.

PR #264, opened 2018-01-24
Clarify and correct comments in hello_world.rs example.

CaptainKraft:
Some of the comments were unclear to me when implementing this example. I think that I've made them easier to understand. I also moved a comment block down to make it more obvious what block of code it refers to.

The comment about drawables drawing from their center does not appear to be correct here. When I run this example, the text is drawn from top-left.


icefoxen:
:D

PR #267, opened 2018-01-26
Fix variable name in documentation

atallahade:


icefoxen:
Whew, there's a lot of these.  Thank you!

PR #269, opened 2018-01-29
Add the Cellular Automata cave generation example and spritesheet.

CaptainKraft:
This example uses simple Cellular Automata generation to create a new cave periodically. It also uses a simple bitmasking technique for autotiling.

Feel free to be ruthless with your code review. I need to learn more constructs, traits, idioms, etc. in Rust. Thanks :-)


icefoxen:
It's slick but it doesn't seem to show much that doesn't already exist in the examples.  Perhaps have it a separate repo with a link in https://github.com/ggez/ggez/blob/master/docs/Projects.md ?

CaptainKraft:
Fair point. I'll continue working on it and link it as a project when it's a bit more substantial.

Thanks for looking it over!

icefoxen:
It doesn't need to be substantial to be included in the projects, just neat.  ;-)

PR #275, opened 2018-02-04
Changed misleading mousewheel event documentation.

Ratysz:
Now reflects SDL2 documentation, and reality (as tested on a Windows 10 PC with a generic mouse).


icefoxen:
Awesome, thank you!

Issue #285, opened 2018-02-20
macOS Compile Error: the type of this value must be known in this context


PSeitz:
I can't compile on macOS, tried with the latest rust nightly.

```
error[E0619]: the type of this value must be known in this context
  --> /Users/d056831/.cargo/registry/src/github.com-1ecc6299db9ec823/app_dirs-1.1.1/src/imp/platform/macos.rs:10:40
   |
10 |         Ok(Component::RootDir.as_ref().into())
   |                                        ^^^^

```

PSeitz:
Seems to be this issue:
https://github.com/AndyBarron/app-dirs-rs/issues/28

icefoxen:
Yeah not super thrilled.  The maintainer did say in that issue that they were super busy, which I can sympathize with, but it'd be nice to at least have simple fixes merged...

Not sure what we can do besides fork and update it, or vendor it and add some patches...

PSeitz:
Maybe we should temporarily redirect to the fix ```app_dirs = { git = "https://github.com/koute/app-dirs-rs" }``` and then wait what happens.

icefoxen:
That's not a fix that will work in a released package, since you can't publish things to cargo if they depend on a git repo.

If you're using the devel branch though, that's good enough.  Will work for the moment.

PSeitz:
Ah okay, well creating a new crate would be kinda overkill, but maybe the only solution...

icefoxen:
Well this is also likely a bug in nightly, given that it works on all other platforms and releases.  Hopefully that'll get fixed eventually.

icefoxen:
Fixed in PR #287, leaving this open so we keep track of it.

icefoxen:
Fixed in release 0.4.2

PR #286, opened 2018-02-20
fix dependency #285

PSeitz:
#285


icefoxen:
Can you re-do this PR against the devel branch, please?

PR #297, opened 2018-03-08
 Implemented TextEditing and TextInput events.

Ratysz:
Can be played with in the input_test example (work just fine on my Win10 instance, tested with Russian and Chinese Simplified layouts).

These are important for multi-language support, how they work and what they do is best described by SDL Wiki:
- [TextEditing](https://wiki.libsdl.org/SDL_TextEditingEvent)
- [TextInput](https://wiki.libsdl.org/SDL_TextInputEvent) (detailed explanation under Remarks is particularly useful, could be adapted for docs)
- [Text input tutorial](https://wiki.libsdl.org/Tutorials/TextInput)

When we eventually migrate to [glutin](https://github.com/tomaka/glutin), TextInput can probably be replaced as-is by [WindowEvent::ReceivedCharacter](https://docs.rs/glutin/0.13.1/glutin/enum.WindowEvent.html#variant.ReceivedCharacter).

### DIFF
--- a/examples/input_test.rs
+++ b/examples/input_test.rs
@@ -90,6 +90,17 @@ impl event::EventHandler for MainState {
                  repeat);
     }
 
+    fn text_editing_event(&mut self, _ctx: &mut Context, text: String, start: i32, length: i32) {
+        println!("Text editing: {}, start {}, length: {}",
+                 text,
+                 start,
+                 length);
+    }
+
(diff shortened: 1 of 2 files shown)


icefoxen:
I didn't implement these because they seemed redundant, but the SDL docs explain the rationale quite well.  Thank you!

Oh you even referred to them in the docstrings :heart:

Issue #306, opened 2018-03-20
C/C++ compiler required when building due to Zip dependency


brianwp3000:
Right now ggez requires a C/C++ compiler to build because of its dependency on the Zip crate. The Zip crate has an optional (but on-by-default) feature to enable handling of zip files compressed with bzip2, and it's that feature that requires a C/C++ compiler to build.

I propose that the zip/bzip2 feature gets surfaced as its own feature in ggez that is enabled by default. This way, users like me who don't want to install a C/C++ compiler can still develop with ggez by specifying "default-features=false". The only thing a person sacrifices by disabling bzip2 support is the ability to work with bzip2 zip files in their games, but that's easily worked around by the developer and bzip2 isn't the usual zip file compression scheme anyway (most zip files use Deflate).

This could also help with Issue #29 ("It would be nice if everything was pure Rust").

icefoxen:
Merged your PR, thank you!

PR #307, opened 2018-03-20
Surface bzip2 feature and put it in default features.

brianwp3000:
This solves Issue #306 by allowing developers to specify "default-features=false" so they don't need to have a C/C++ installed when using ggez. It preserves backwards compatibility, because the bzip2 feature is still on-by-default.


icefoxen:
Well that's interesting, I thought the zip crate was Pure Rust(tm). I'll look into this when I'm sober, thanks.

PR #308, opened 2018-03-21
Line caps and joins

kaedroho:
This commit adds the following methods to the ``graphics`` module. These
both exist in LOVE:

 - ``set_line_join`` - Sets the style to use to join two connected line
 segments together. Possible values are miter, miter-clip, round, bevel
 and arcs. See https://nical.github.io/lyon-doc/lyon/tessellation/path_stroke/enum.LineJoin.html
 - ``set_line_cap`` - Sets the style to terminate the end of line
 segments that aren't connected to another. Possible values are butt,
 square and round. See https://nical.github.io/lyon-doc/lyon/tessellation/path_stroke/enum.LineCap.html

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -18,6 +18,9 @@ pub(crate) struct GraphicsContextGeneric<B>
 where
     B: BackendSpec,
 {
+    pub(crate) line_join: LineJoin,
+    pub(crate) line_cap: LineCap,
+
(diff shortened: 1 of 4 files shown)


icefoxen:
Possible way to make it compatible: Make a separate `Line` type that includes cap/join info?  That's still an API break though, if we extend the `DrawMode` enum.  Sigh.

Issue #294, opened 2018-02-27
Creating context fails with "glxSwapIntervalMESA failed"


rocurley:
Whenever ggez tries to make a context, it fails with `UnknownError("glxSwapIntervalMESA failed")`. I've included at the end a backtrace running the `hello_world` example (as of commit `40d70a79bb3f46c75b4d11bd4424c036c540a9fd`).

This is running in a VM, which seems extremely likely to be the issue at some level. It's a Ubuntu 16.04 VM running on a Windows 7 host in VMWare Workstation 12. Here's the output of [glxinfo](https://github.com/ggez/ggez/files/1761581/glxinfo.txt).

I've tried modify the OpenGL version used, to no avail: I've tried `3.0`, `2.1`, and `1.1`. Here's the diff showing what I did in case I did it wrong:
```
diff --git a/examples/hello_world.rs b/examples/hello_world.rs
index 58e8e9b..c3138fd 100644
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -62,7 +62,8 @@ impl event::EventHandler for MainState {
 // do the work of creating our MainState and running our game,
 // * then just call `game.run()` which runs the `Game` mainloop.
 pub fn main() {
-    let c = conf::Conf::new();
+    let mut c = conf::Conf::new();
+    c.backend = conf::Backend::OpenGL{major: 3, minor:0};
     let ctx = &mut Context::load_from_conf("helloworld", "ggez", c).unwrap();
 
     // We add the CARGO_MANIFEST_DIR/resources do the filesystems paths so
```
And here's the backtrace:
```
cargo run --example hello_world
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/examples/hello_world`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnknownError("glxSwapIntervalMESA failed")', libcore/result.rs:945:5
stack backtrace:
   0:     0x55ef91ea6c63 - std::sys::unix::backtrace::tracing::imp::unwind_backtrace::h038efa1ead8ee277
                               at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1:     0x55ef91ea14c4 - std::sys_common::backtrace::_print::haa2469a6e6008f11
                               at libstd/sys_common/backtrace.rs:71
   2:     0x55ef91ea5c0d - std::panicking::default_hook::{{closure}}::ha5918fb5e6555fa7
[...]

PR #309, opened 2018-03-22
Logging feature.

Ratysz:
Rewrote how `ggez` handles internal logging, and implemented a new feature: `use-log-crate`; example included.

Now, by default, `ggez` will format internal log messages as `[target][LEVEL] message`, close to how a simple frontend of [`log`](https://docs.rs/log/) crate would do it.

When compiled with `--features "use-log-crate"`, internal logging will instead use macros from [`log`](https://docs.rs/log/) crate.

From end user (read: executable) perspective, it's entirely opt-in, as demonstrated by all examples (save for new "logging" example, which shows how to actually opt-in).

**Details**

This is achieved via a set of new macros, like `ggez_info!` or `ggez_error!`, that closely mimic equivalent macros from `log` crate (`info!`, `error!`, etc).

Without `use-log-crate` feature these compile down to print aforementioned `[target][LEVEL] message` line; with the feature the macros compile down to their direct equivalents from `log` crate.

As far as I know, there is no runtime overhead - when the feature is disabled, at least. With the feature, it most likely depends on chosen logging frontend.

`ggez`-depending libraries can use these new macros as well, preserving the opt-in behavior for their dependant executables.

Examples and tests were not touched, so grand total of actual changed lines is 4.

The random diffs are from rustfmt.

**Open questions**

Names. Feature itself (currently `use-log-crate`), macros (currently `ggez_X!`, with *X* being all the macros from `log` crate).

Default. Should this feature be enabled by default? (currently no)

Docs. I've put in a lot of hard links there, couldn't be bothered to learn if there's a better way (yet).

Alternative approach. It might be simpler (to understand and maintain) to embrace `log` crate internally, and provide an internal fallback logger frontend that just `println!`s the records - not entirely sure how overhead-free that would be.

That should be all the info needed to start a discussion; I'll comment if I remember anything else. Questions, edits, reviews, comments, etc are all welcome, of course.


icefoxen:
Well this is interesting.  I have no idea what to think of this.

It appears that just using the `log` crate to do logging by default does nothing, and allow users to specify a logging backend that will actually output messages.  For a library, that seems like the desired behavior.

I'm going to have to play with this more.

Ratysz:
... oh. For some reason, I assumed you knew what `log` crate does and how it does it, and [this comment](https://github.com/ggez/ggez/issues/294#issuecomment-371861897) implied that wasn't going to work for `ggez`. I blame the flu.

In light of that, my suggested alternative approach might be the way to go - I'll tinker with that within the week, while discussion is (hopefully) happening.

icefoxen:
Derp.  XD  I know what `log` does but not that much about how it does it.  Sorry.  I haven't thought about this problem very hard lately.  This PR, I confess, makes it a little bit more clear that using something like `log` may be easier and less problematic than I expected.

Also I feel that logging is a little orthogonal to that particular issue, since I was thinking of that as an API problem, not as a communicating-information-to-humans problem.  The desired behavior in that issue is "the program can see something went wrong, then decide to keep going in a degraded state"; informing the user that something is wrong doesn't do that, so I wasn't thinking about that angle.

Ratysz:
"Alternative approach" over at #310; leaving this open until we pick a solution.

icefoxen:
:+1: 

I'll see if I can get time to play with both of these and see how they stack up against each other.

Ratysz:
Closing in favor of #331

PR #310, opened 2018-03-23
Embraced log crate, implemented default logging frontend.

Ratysz:
A take on the "alternative approach" I suggested in #309: use [`log`](https://docs.rs/log/) crate throughout `ggez`, provide an optional default frontend.

Picked [`log4rs`](https://docs.rs/log4rs/) as said frontend, [`fern`](https://docs.rs/fern/) could be an alternative choice - both of these allow powerful filtering options, which are needed because `gfx_device_gl::factory` spams Info level with just about everything. Right now, only filtering out the buffer creation messages (because they clog the console at runtime, in some examples at least).

Default logger can be cleanly disabled by disabling new `logger` feature (it's on by default); this is useful for executables wanting to roll their own `log` frontend, or release builds that don't need console logging at all.

I went ahead and done the busywork of migrating examples to use `log` macros instead of `println!`; tests still use `println!`.

Will leave #309 open for now, until we pick a solution.


icefoxen:
I like it but you might have embraced a bit too hard.  ;-)  I feel that using logging macros instead of `println!` in examples just adds unnecessary noise.

I'm a little unconvinced that we should have a logging frontend built-in, even behind a feature flag, but we should definitely have an example demonstrating how to set one up.

Ratysz:
Yeah, that looks like the least invasive option (I'm likely not experienced enough for "less is more" approach to be the default modus operandi).

However, this wild meandering of mine in both log-related PRs is not without a reason - I've been trying to both eat the cake and keep it; basically, there are two uses of `println!` macro in `ggez` itself:

- `EventHandler::quit_event()` - default callback, honestly don't think any sort of output is needed there;

- `Filesystem::print_all()` - called by `Context::print_resource_stats()`, used in 3 examples. Changing it to use `log` crate macros will leave said examples with no expected console output; keeping it intact with `println!` makes it useless in no-console apps.

I see several solutions:

1. Ignore and hope that nothing `ggez` does down the line will warrant debug output - this includes printing (or otherwise logging) error details;

2. Use `log` macros, remove `Context::print_resource_stats()` from examples;

3. Use `log` macros, tweak examples to include a small `log` frontend;

4. Combination of previous two: remove `Context::print_resource_stats()` from existing examples, add a new example that implements a `log` frontend and shows off `Context::print_resource_stats()` and whatever else `log` output `ggez` might have in the future;

5. Implement a new method in `Context` and/or `Filesystem` that uses `log` crate macros;

6. Use `log` macros, add a feature for default logging frontend that only prints `ggez`' own logs, without any fancy formatting (basically `println!`);

7. 6, to keep existing examples working as they were, and a new example that trades the default logger for a more involved implementation.

icefoxen:
Maintaining a project that's gone through a couple major rewrites in the last couple years will make you appreciate "less is more" a lot.  ;-)  Start small and work your way up!  We can always keep adding functionality as necessary.

I feel like your solution space succumbs to this a bit as well; you don't need to make the problem that complicated.  We definitely want to use `log` macros.  And we can definitely include a logging frontend crate in the examples to demonstrate how to do it.  So, get those basics working and working nicely.  But there's no reason `print_resource_stats()` can't use `println!` -- a `log_resource_stats()` function might be added, and we can maybe deprecate `print_resource_stats()` sometime, and see how we feel about it.  But generally once we have that sort of thing in place it will be obvious what needs to be done.

So my suggestion for this PR: Please pull the log4rs stuff out of ggez itself and put it in as a logging frontend in one or two of the examples, and then we can play with the results and see how we like it.

Ratysz:
Closing in favor of #331

I figured a cleaner branch would be better than essentially a full revert.

Issue #311, opened 2018-03-23
Glob imports


ozkriff:
I've noticed that there're a lot of glob imports in the code and examples

```text
$ git grep '::\*' examples src | wc -l
47
```

Are they really required? I don't think that they are considered to be very idiomatic.

icefoxen:
*Most* (not all) of those fall into a few categories:

 * `use ggez::*`, which imports a very few things: Context/ContextBuilder, GameError/GameResult, and submodules.  Count: 8
 * Re-exports of modules by `graphics/mod.rs` or `lib.rs`.  Count: 9
 * Glob-imports of `graphics` or `gfx` submodules in the graphics subsystem.  Count: 12
 * Actual glob-imports in submodules.  Count: 11

It would be nice to make this better, I admit.  I've occasionally considered a `prelude` module but it would basically just have the things that the `ggez` module has in it anyway.

 agree that using specific imports would on the whole be better, but I don't really feel like the harmful case is the majority.  I'd worry more about it in the examples than in the general code base, since those are what people are going to be looking at and emulating more often. Those come in two forms: `use ggez::*;`, which I consider Mostly Harmless, and `use ggez::event::*;` or `use ggez::graphics::*;` which I confess probably aren't ideal but have a lot of fairly verbose bits and pieces which are used a *lot*, so glob-including them makes the code a lot more concise.  We should, at the very least, replace those with `use ggez::graphics as g;` and `use ggez::event as e;` or something along those lines.

PR #314, opened 2018-03-24
anti-`glob`alization of examples

ozkriff:
#311

### DIFF
--- a/examples/astroblasto.rs
+++ b/examples/astroblasto.rs
@@ -4,10 +4,11 @@
 
 extern crate ggez;
 extern crate rand;
+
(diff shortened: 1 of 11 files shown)


itmuckel:
`use module::{self};` imports the module itself? Nice! Didn't know about that. 😄 I also never seen this in the rust book I think. Where did you learn that from?

ozkriff:
@itmuckel it's actually mentioned a few times in https://doc.rust-lang.org/book/first-edition/crates-and-modules.html . But I think that I've seen that first time in some ancient RFC a long time ago - http://rust-lang.github.io/rfcs/0532-self-in-use.html possibly.

PR #319, opened 2018-03-24
README: Add deps.rs badge

ozkriff:
r? :)


icefoxen:
That's useful for maintainers but I'm not sure what it adds for users...

icefoxen:
Suggested alternative?  https://github.com/ggez/ggez/commit/0e5c35cce39a23f3743fd64d3f062cb2d7abb204

ozkriff:
> but I'm not sure what it adds for users

The idea is that it allows users to quickly see that library is well maintained and it encourages maintainers to updated deps more swiftly.

Though I'm not sure if the badge is worth it. :)

ozkriff:
> Potential alternative? https://github.com/ggez/ggez/commit/0e5c35cce3

Yep, that's also an option if you think that the readme is overloaded with badges :)

PR #322, opened 2018-03-24
Add termhn/ggez_snake to examples

### LINKED ISSUE #321: Add termhn/ggez_snake to examples?
https://github.com/termhn/ggez_snake is simple, has no additional dependencies and doesn't use any assets - why not add it as one more example to the repo?

Examples that live outside of the core repo has a tendency to decay gradually.

ozkriff:
@termhn You don't mind? 

Closes #321

I love sneks

### DIFF
--- a/examples/snake.rs
+++ b/examples/snake.rs
@@ -1,0 +1,285 @@
+//! A small snake game done after watching
+//! <https://www.youtube.com/watch?v=HCwMb0KslX8>
+//! to showcase ggez and how it relates/differs from piston.
+//!
+//! Author: @termhn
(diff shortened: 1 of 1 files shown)


termhn:
Yeah I'm fine with it. I think we should add more comments like astroblasto has though.

ozkriff:
My English is meh, but I could try writing these comments.

@icefoxen Should I do it in this PR or we are merging this and I'll try to create another PR with comments in a few days?

icefoxen:
Let's merge it now and add docs in another PR.

PR #328, opened 2018-03-26
FAQ: Add a small note about resolution independence

ozkriff:


itmuckel:
Thank you for this useful contribution! 

How does this affect positioning of sprites? Do I then have to use position coordinates in the range [0.0, 1.0]?

ozkriff:
Yep, after this call, your coordinate system have `Rect::new(0.0, 0.0, 1.0, 1.0)` boundaries.

Issue #281, opened 2018-02-11
How to update texts


itmuckel:
Hey guys! 
I've finally updated my project to ggez 0.4 and I love the progress. This is my favorite rust project. :smile:

Now that I got the basic structure of my game, I want to update some texts on specific events or continuously (for example: slowly decreasing HP of a player's character). 
Currently I'm doing this by creating new instances of `ggez::graphics::Text` every time the value changes.

This works, but it feels wrong. Is there a way to update the text's contents without recreating (reallocating) a new text instance? If no, is this performance-wise not critical e.g. the way to go here? How are you guys doing it?

Greetings,
itmuckel

icefoxen:
Nope, recreating the Text is the best way currently. It is not a fast
operation, so you don't want to do it every frame if the text doesn't
change, but in my experience the cost of memory allocation is far smaller
than the cost of all the math needed to render the font.

The obvious optimization is to assemble text objects out of pre rendered
chunks, which we want to get to eventually but we just aren't there yet.
See issues #132 and #270 for discussion on that.


Feb 11, 2018 03:58, "itmuckel" <notifications@github.com> wrote:

Hey guys!
I've finally updated my project to ggez 0.4 and I love the progress. This
is my favorite rust project. 😄

Now that I got the basic structure of my game, I want to update some texts
on specific events or continuously (for example: slowly decreasing HP of a
player's character).
Currently I'm doing this by creating new instances of ggez::graphics::Text
every time the value changes.

This works, but it feels wrong. Is there a way to update the text's
contents without recreating (reallocating) a new text instance? If no, is
this performance-wise not critical e.g. the way to go here? How are you
guys doing it?

Greetings,
itmuckel

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/ggez/ggez/issues/281>, or mute the thread

icefoxen:
@itmuckel I'm closing this.  If you have any further questions please feel free to re-open it.

alanpoon:
Any plans to support text_edit??

Issue #132, opened 2017-08-28
Integrate gfx_glyph


ishitatsuyuki:
It's a high quality text rendering library with line breaking handling, which should be better than the current codebase.

icefoxen:
It would be nice, yes!  Glyph caching is on the to-do list anyway; I've been looking at gfx_glyph as a way of doing this but haven't really tried it out.

icefoxen:
Looks okay but there's a few rough edges.  The main ones:

 * `GlyphBrush` takes a lifetime for the font it refers to but doesn't seem to have an option for a `Cow` or such, so if you dynamically load a font and try to store it and a `GlyphBrush` together you end up with the Dreaded Self-Borrowing Struct
 * It assumes your render target is using format `gfx::format::Depth` but by default ggez uses ``gfx::format::DepthStencil`

ishitatsuyuki:
I have cross posted the second issue. I'm not sure about the solution of the other two, so I will leave it for further discussion.

alexheretic:
@icefoxen regarding your first rough edge; gfx_glyph takes `Into<SharedBytes<'a>>` as font-data rather than a reference to a byte array, so you can use owned data (since v0.3).

Let me know if you find more annoyances with gfx_glyph.

Ratysz:
So I've been tinkering with this (not everything is on my public fork), and ran into `graphics::Text` being incompatible with how `gfx_glyph` does things.

`graphics::Text` stores a texture and draws that in `Text::draw_ex()`; "updating" it is remaking it via `Text::new()`, with a `graphics::text::Font` reference (it's an enum that determines how the `Text`s texture is created).

`gfx_glyph` frontloads all fonts into a `GlyphBrushBuilder`, which builds into a `GlyphBrush`. It queues text fragments (`VariedSection`s) that contain all drawing information, including a `FontId` (handle to a font stored in the `GlyphBrush`), and draws them to a `RenderTargetView` all at once with `GlyphBrush::draw_queued()`.

So far, I see these approaches on how to marry the two:

- Try to preserve current strategy, build a `GlyphBrush` instance when initializing the `Font`, store it inside (either a new variant, or rewrite of `Font::TTFFont`). Calling `GlyphBrush::draw_queued()` in every `Text::draw_ex()` defeats the purpose, so a kind of batching API is needed.

[...]

PR #342, opened 2018-03-31
Add `glyph_cache` example

dettalant:
See #132 and #281

Used external [gfx_glyph](https://github.com/alexheretic/gfx-glyph) crate in ggez EventHandler. Like a [examples/cube.rs](https://github.com/ggez/ggez/blob/master/examples/cube.rs).

This example is not best way, but some kind of game develop is must need a fast text update...

Maybe it will be help to develop game until ggez has implement glyph-cache feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,4 @@ nalgebra = "0.14"
 [dev-dependencies]
 rand = "0.4"
 clap = "2"
+gfx_glyph = "0.10.1"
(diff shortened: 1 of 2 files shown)


icefoxen:
Very nice!  I'm uncertain about merging this just yet since adding examples always has a maintenance burden involved, and as this demonstrates it really shouldn't be that hard to integrate `gfx_glyph` into ggez itself by now.  So we should just do that.  :D

dettalant:
you are right. `gfx_glyph` into ggez itself is more better way for game developer.

It sounds good...Ok, I will closed this PR. Thankyou!

Issue #268, opened 2018-01-29
Can we have pixel-based sizing for fonts?


makoConstruct:
I'm trying to make a pixelated game, using pixel art and a pixel font. Regarding text, this doesn't seem to be doable? Point sizes are unusable for various reasons:

* Why would a gamedev want to deal in Pts? What are Pts? Do they even have a fixed meaning?

* My screen lies horribly about its DPI, many screens are the same. Very few screens give a completely accurate number. It can't really be relied on for anything.

* If I had access to the measured DPI I might use it to scale the text down to compensate away the DPI-sensitive code further up the toolchain, but there doesn't seem to be a way to get that number.

To hammer my point home, I will supply a picture of the hello world example as it appears on my machine https://imgur.com/a/TUXLr

icefoxen:
A point is 1/72 of an inch.

I would love to have pixel-absolute sizes for fonts but I've yet to see a TTF renderer that really offers that.  :-/  If we can figure out how to do it, we should do it.

For now the easiest way to get pixel-absolute fonts is to use bitmap fonts.

uggwar:
The DPI is calculated from the display size, and not the size of the window. So running my little game in the same sized window on my internal and external monitor gave different font sizes. I forced ggez to use 72 as DPI and the problem was "solved". Not pretty, but not sure how to fix this properly.

icefoxen:
Notes on how SDL2 does it:

```
14:07 < m12y_> Icefoz: i've relied on SDL2's drawable_size for text rendering and it's worked perfectly for me (especially on mac/ios retina)
14:19 < m12y_> it returns the resolution in physical pixels (e.g. what you want to pass to glViewport), whereas the normal `size` method returns the resolution in (DPI aware) virtual pixels. i use`size` to lay out my UIs, and then `drawable_size` to determine the actual font size to draw in that UI
```

icefoxen:
Honestly just having a function that builds a `Text` to a given pixel size would be very easy, even if it gets abused.  :|

Sushisource:
I'm doing this by basically determining the pt -> pixel sizes experimentally and just saving that info statically for now. Having that function would be killer.

icefoxen:
See commit 99fce38, it's a very small change.  Just needs testing.

icefoxen:
Tested, it works.

icefoxen:
Closing.  :cry: See #587 for basically the reason why.

Issue #211, opened 2017-11-26
Outlined Text


itmuckel:
I'm currently developing a game in which I have to draw a lot of text. For that I use `ggez::graphics::Text`.
Unfortunately it is hard to read the text because of some background images. 

Is there a builtin way to draw text with black outlines?

Of course, I can draw the text twice, first black and a bit bigger, then in its final color a bit smaller to achieve this effect. But I thought I'll ask first, before implementing something myself, that's already there. ;-)

icefoxen:
There isn't a built-in way; I don't think this is supported by rusttype.

However, I vaguely recall that some fonts DO have an "outlined" face, similar to "bold" or "italic", so that might be worth looking into.

michaelbachner:
Yes, my first thought was getting an outlined font. But using it showed only the outlines and nothing inside. 

Wait... I have an idea. When you got the filled and the outlined version of a font, draw the filled version first in black or whatever color and then draw the outlines on top in white or whatever color. 

But this is very font specific and I'd like a general solution. There are far more non-outlined fonts.

svenstaro:
This could be done using a shader or a post processing algorithm. Shouldn't be too bad to implement this naively. I'm just not sure we want they in ggez but I don't think it'll hurt.

itmuckel:
I tried out two approaches.
1. add black font underneath with an offset (shadow):
![image](https://user-images.githubusercontent.com/18561536/33433083-8f979d7e-d5da-11e7-8114-dc7d2ed01ea2.png)
2. add four(!) black fonts underneath with an offset (one for each side):
![image](https://user-images.githubusercontent.com/18561536/33433448-e43399b8-d5db-11e7-8c8a-b22ce8bf06e0.png)

@svenstaro: If it's not so hard, then could you tell me how to do this? My game is not so graphically intense, so I have no problem drawing 5 texts, where one should be, but it feels so wrong. ;-)

icefoxen:
Both techniques honestly look pretty good to me.  Drawing a text object 5 times should not be a huge problem since it's drawn via the GPU, but I agree that it feels wrong.  :-)

To disable smoothing when scaling text you want to use this function: https://docs.rs/ggez/0.3.4/ggez/graphics/struct.Text.html#method.set_filter , set the filter mode to `Nearest`.  Making the font renderer itself not try to do anti-aliasing is not something I know how to do though...

termhn:
[...]

Issue #108, opened 2017-07-19
Support for multi-line text


arteme:
It would be nice to be able to render multi-line text. Currently line breaks (``"\n"``) are treated as text characters.

icefoxen:
Rendering multi-line text is nontrivial because different use cases may have different ideas of how much space should go between lines.  The font gives you this information but you never have to actually use it.

Still, a sane default might be nice.

icefoxen:
There IS now support for creating wrapped text yourself; see `Font::get_wrap()`

icefoxen:
If we do #132 then this will come with it, I think.

icefoxen:
Fixed by #362!

Issue #92, opened 2017-06-11
More advanced font rendering


expenses:
Hi, I've been having a great time with ggez so far, but I've found font rendering to be a little lacking.

Some features that would be appreciated:

* Setting the filtering mode of the font (nearest neighbor, linear interpolation etc.)
* Setting the colour of the font, and specifying a color in `Text::new`
* Changing the line height of the font

For some idea I'd also check out [love2d's font documentation](https://love2d.org/wiki/Font). Thanks a lot for a great library!

icefoxen:
Filtering modes and colors are properties of the `Text`, not the font.  See <https://docs.rs/ggez/0.3.1/ggez/graphics/struct.Text.html#method.set_filter> for setting the filter mode, otherwise the one set by `graphics::set_default_filter()` will be used.  Text, like everything else, is drawn with the color set by `graphics::set_color()`, though one might argue that you should be able to set the color per `Text` object.  (Odd that love2d sets the filter mode on the font; I might have to look at that more.)

Cool shiny technicolor text rendering by specifying colors for particular strings should not be *super* hard to add to the engine but is not planned for the immediate future.  PR's welcome!  :D

Line height is a legit request

icefoxen:
Fixed by #362!

PR #362, opened 2018-04-27
gfx_glyph integration redux

Ratysz:
This is a work in progress; I feel that having specific diffs in front of us makes it easier to discuss exact implementation of #132 (start of discussion can be found there).

(Side note: trying to adopt verbose commit messages, instead of post-factum dissection in comments.)


Ratysz:
I think I've reached a point where I'll have to start adding (even more) code that's trivial and/or will be removed if we absorb `TextCached` into `Text` (or outright replace latter with the former).

It's functional as it is, though I'm not completely sure the transforms in `TextCached::draw_ex()` all work as they should, or if there's a better way to do that entirely.

Summary:

- Added a `pub(crate) glyph_brush: GlyphBrush` field to `GraphicsContext`; it initializes with the same data as `Font::default_font()` (which had to be slightly rewritten to hopefully avoid duplication).

- Added a `Font::GlyphFont` variant. It can be created by loading a file with `Font::new_glyph_font()`, or re-created (retrieved from the brush) from a `usize` ID via `Font::get_glyph_font_by_id()`. Don't think there's any (significant) overhead for re-creation. Said font variants's scale information is stored within it, as with current ones (there is no hard reason for it, other than keeping it out of `TextCached` API).

- Added `TextCached` drawable. API is nearly identical to `Text` - the getter methods for width and texture and whatnot are missing; most of them don't make sense in the context.

- Added an (undocumented, but straightforward) example for experimenting with said `TextCached`.

There are a few standing questions:

- General API. Might be worthwhile to review what methods `Text` actually needs before we start merging/replacing it; that depends on if we're keeping the rendered `Font` variants (and if we're not, there will be a lot of dead code in text.rs).

- `TextCached` bounds, and everything to do with them. From what I understand, `gfx_glyph` offers automatic wrapping, done via `Section::bounds` (and `Section::layout`) field - this can probably be used by/as `TextCached::width()` and `TextCached::height()`.

- Performance. Two things to investigate: whether or not having separate `Font::GlyphFont`s for separate font scales is faster than using the same `Font::GlyphFont` with different scales; and, how much overhead is calling `glyph_brush.draw_queued()` in every `TextCached::draw_ex()`, over just once, somewhere, somehow.

- The example. If we end up seamlessly integrating `TextCached` into `Text`, it will be entirely pointless.

icefoxen:
Sorry, took me a couple days to find time to take a good look at this.

I like it.

Only note, you should probably use [`DrawParam::into_matrix()`](https://github.com/ggez/ggez/blob/master/src/graphics/drawparam.rs#L48) to set up the transform for `TextCached::draw_ex()`.  Hopefully that makes it all work out correctly.

On to pondering questions...

 * General API.  I really want to be able to just remove the pre-rendered `Text` and `TTFFont` entirely and use `gfx_glyph` for everything.  I want to make an 0.5.0 release that does that, pretty much; my hope is that `gfx_glyph` will be entirely superior.  We can also deal with #268 in the process.
 * Bounds, I haven't looked at yet.  `gfx_glyph` definitely has a richer API for text drawing than ggez currently does, and we should take full advantage of that.  Love2D isn't quite much guidance because it hides all the Cool Formatting Stuff behind `print` and `printf` functions that allow coloring and wrapping of strings; we should probably use something like the [`TextParam`](https://github.com/ggez/ggez/blob/devel/src/graphics/glyphcache.rs#L89) type from dettalant's PR.  #92 and #108 would get solved by this.  Maybe theoretically #211 someday...  Should look at that more.
 * Performance I think we have to look at empirically.  I will whip up some benchmarks when I have time.
 * The example is good for a) profiling, and b) showing off how shiny our stuff is now.

So what I'm imagining now is something that looks sort of like this...

```rust
struct TextCached;
impl Drawable for TextCached { 
    fn draw_ex(self, ctx: &mut Context, param: DrawParams) -> GameResult<()> {
        let text_params = param.into(); // Makes a TextParams with reasonable defaults where needed
        self.queue(ctx, params);
        self.draw_queued(ctx);
    }
}
impl TextCached {
    pub fn queue(&self, ctx: &mut Context, params: &TextParams) {
        ctx.glyph_brush.queue(params.into());
    }

    pub fn draw_queued(&self, ctx: &mut Context) {
        ctx.glyph_brush.draw_queued();
    }
}
```

I think this would let users both do the simple thing easily, and also access the full power of the API...  a little like how `Mesh` and `MeshBuilder` do about the same thing but one is quick and easy while the other is more powerful.

icefoxen:
Hilarious benchmark results, from here: https://github.com/icefoxen/ggez/blob/gfx_glyph_integration_too/examples/glyph_bench.rs .  It draws 2500 copies of a single text object, either with the current `Text::draw()`, with the `TextCached::draw()` from this PR, and with a trivial `TextCached::queue()` queue'ing up all the copies and `TextCached::draw_queued()` drawing them all at once.

* Drawing nothing: 880 fps debug, 12,000 fps release
* Drawing with `Text::draw()`: 2 fps debug mode, 81 fps release mode.
* Drawing with `TextCached::draw()`: 2 fps debug mode, 265 fps release mode.
* Drawing with `TextCached::queue()` and `TextCached::draw_queued()`: 30 fps debug mode, 884 fps release mode.

So, there's basically no reason NOT to use `GlyphBrush` for everything.

Ratysz:
This is pretty much done, now. Summary, details, and everything else should be apparent from documentation and example.

There are still a few bits to tighten (all marked with `TODO`) and verify (how well does pixel sizing actually work, etc - mostly things that happened while I was working on this), but it's quite serviceable now.

I'll check up on my other projects, and start working on fully supplanting current `Text` with this. In a new PR, of course; probably after 0.4.3. Hoping to see any and all criticism, thoughts, and insights well before that!

PR #372, opened 2018-05-06
Game of life example

terakilobyte:
I came across ggez being used in a YouTube video and thought I'd give the game of life a try and submit a PR.

I'm still very new to Rust so please be brutal with feedback.

### DIFF
--- a/examples/game_of_life.rs
+++ b/examples/game_of_life.rs
@@ -1,0 +1,111 @@
+extern crate ggez;
+extern crate rand;
+
+use ggez::conf;
+use ggez::event;
+use ggez::graphics::{self, DrawMode};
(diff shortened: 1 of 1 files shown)


icefoxen:
Thank you!  Unfortunately, it doesn't really show anything new that the other examples don't.  If you break it out into its own project we would happily link to it in the projects file though.

Code feedback: 

 * `Box<Vec<...>>` can just be `Vec<...>`; Vec's are already on the heap
 * Using a `SpriteBatch` for drawing would be faster large grids

That's all though!  Very good work.

terakilobyte:
Cool, thanks!

I switched over to `SpriteBatch` and saw substantial performance improvements, great tip.

Issue #170, opened 2017-09-30
Resizing window via set_mode() or set_resolution() doesn't fire resize_event()


icefoxen:
Only resizing the window from the window manager does.

Weird.

This is probably an SDL thing and I *sort* of understand the logic (since the resize doesn't come from an event caused by the user) but it's still basically never what we would want, afaict.

icefoxen:
ahahaha it's worse than that... it only does it SOMETIMES!  At least on my Linux laptop.  Apparently only when you increase the screen to a resolution that's larger than what it already was?

Also if set_resolution() is redundant with set_mode() then we should remove it.

icefoxen:
This is probably a bug with SDL or such so I'm just making it not-a-blocker for 0.4

TatriX:
So how can I properly resize window *after* I created context?

TatriX:
Well that worked:
```rust
graphics::set_mode(ctx, WindowMode::default().dimensions(screen_width as u32, screen_height as u32)).unwrap();
graphics::set_screen_coordinates(ctx, Rect::new(0.0, 0.0, screen_width, screen_height)).unwrap();
```

icefoxen:
Fixed by switching to winit.

PR #374, opened 2018-05-08
Added a `GraphicsContext::resize_viewport()` call to `set_mode()`

Ratysz:
This sort of sidesteps #170, without fixing the underlying issue.

Without this, changing resolution by `graphics::set_resolution()` screws up the viewport, which can only be fixed by a resize event (ultimately triggering `GraphicsContext::resize_viewport()`).

Minimal-ish example to reproduce/test, adapted from @mpatraw's:

```rust
extern crate ggez;

use ggez::conf::{WindowMode, WindowSetup};
use ggez::event;
use ggez::{Context, ContextBuilder, GameResult};
use ggez::graphics::{self, Point2, Rect};
use std::env;
use std::path;

struct MainState {
    text: graphics::Text,
}

impl MainState {
    fn new(ctx: &mut Context) -> GameResult<MainState> {
        let font = graphics::Font::new(ctx, "/DejaVuSerif.ttf", 12)?;
        let (w, h) = (font.get_width("W") * 80, font.get_height() * 25);
        graphics::set_resolution(ctx, w as u32, h as u32)?;
        graphics::set_screen_coordinates(ctx, Rect::new(0.0, 0.0, w as f32, h as f32))?;
        let text = graphics::Text::new(ctx, "Hello world!", &font)?;
        Ok(MainState { text })
    }
}

impl event::EventHandler for MainState {
        Ok(())
    }

        graphics::clear(ctx);
        graphics::draw(ctx, &self.text, Point2::new(0.0, 0.0), 0.0)?;
        graphics::draw(ctx, &self.text, Point2::new(0.0, 100.0), 0.0)?;
        graphics::present(ctx);
        Ok(())
    }
}

pub fn main() {
    let ctx = &mut ContextBuilder::new("helloworld", "ggez")
        .window_setup(WindowSetup::default())
        .window_mode(WindowMode::default().dimensions(640, 480))
        .build()
        .unwrap();

    if let Ok(manifest_dir) = env::var("CARGO_MANIFEST_DIR") {
        let mut path = path::PathBuf::from(manifest_dir);
        path.push("resources");
        ctx.filesystem.mount(&path, true);
    }

    let state = &mut MainState::new(ctx).unwrap();
    if let Err(e) = event::run(ctx, state) {
        println!("Error encountered: {}", e);
    } else {
        println!("Game exited cleanly.");
    }
}
```

### DIFF
--- a/src/graphics/mod.rs
+++ b/src/graphics/mod.rs
@@ -708,8 +708,10 @@ pub fn set_mode(context: &mut Context, mode: WindowMode) -> GameResult<()> {
     }
     {
         let video = &mut context.sdl_context.video()?;
-        GraphicsContext::set_vsync(video, mode.vsync)
+        GraphicsContext::set_vsync(video, mode.vsync)?;
     }
+    context.gfx_context.resize_viewport();
+    Ok(())
 }
 
 /// Toggles the fullscreen state of the window subsystem


icefoxen:
I've considered something like this before and don't recall why I didn't do it; I think because it's a band-aid.  A user may have other things that need to happen when a resize event happens, such as resizing canvas's as well; making the event actually fired when it should be is probably preferable, even if we have to hack around with the SDL event queue a bit.  I'd have to double-check all the interactions to make sure what I think is happening is actually happening.

I'm hoping a little bit that this magically stops being a problem when we switch to `winit`.  :stuck_out_tongue: But that's a couple versions out still.

PR #386, opened 2018-06-03
docs/Projects.md: Add ozkriff/zemeroth

ozkriff:
Zemeroth uses ggez since https://github.com/ozkriff/zemeroth/pull/247


icefoxen:
Yay!

PR #388, opened 2018-06-03
[WIP] Port to `winit`

Ratysz:
Running PR for migration from `sdl2` to `winit`. As of writing, it's in embryonic stages - nothing is written in stone.

It will be valuable to have more eyes on it as it develops, and discussion can be used to coordinate efforts.

I don't have a concrete plan right now, just slowly rewriting and refactoring things as I happen upon them; I'll post an update once it compiles.


icefoxen:
Looks very nice in general.  Feel free to alter the shape of the `EventHandler` trait and associated event types to be more like what `winit` expects if you need to; that's the main API change I'm expecting to happen and it's entirely acceptable.

icefoxen:
Any reason I shouldn't merge this into `devel` so we can keep working from this base?

Ratysz:
Some things will break, but I kept track of them (either commit messages or
TODOs) - if that's fine then sure, go for it!

PR #395, opened 2018-06-08
Update Lyon

Lapz:
Update Lyon to fix a [bug](https://github.com/nical/lyon/issues/358)


icefoxen:
Thank you!  Could you please make this PR against `devel`, and make it `"^0.10.2"` so when 0.10.3 comes out that version is also valid?

PR #397, opened 2018-06-09
Improved `lyon` support

### LINKED ISSUE #396: Improve line caps and joins
Basically, remove the stateful bits and make you provide optional line styles in the drawing calls.

Ratysz:
(also fixes #396)

With overall builder pattern renaissance being in full swing, adding variants to `DrawMode` that utilize `lyon`s own builders seemed like the most conservative yet expressive solution. Unfortunately, this nearly completely amends #308.

As a bonus, I'm almost certain this is a non-breaking change (unless exhaustive matching over `DrawMode` is something user applications would want to do).

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -21,9 +21,6 @@ where
     B: BackendSpec<SurfaceType = C>,
     C: gfx::format::Formatted,
 {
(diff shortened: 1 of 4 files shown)


icefoxen:
Darn it Ratysz, how am I supposed to get anything done when you do it all first?

PR #402, opened 2018-06-14
[WIP] Add `GamepadContext` and event handling for gamepads

bsvercl:
This adds a `GamepadContext` struct like before, but uses `gilrs` instead of SDL2. 

There's still a little more work to be done.


bsvercl:
I'm looking for feedback on the current code. Is there anything else I should add; like utility functions similar to #406 maybe?

icefoxen:
@bsvercl If you want!  We'll add them sooner or later.  I'm going to merge this as it is now; if you want to make further improvements just open another PR.  :-)

Issue #391, opened 2018-06-06
Winit integration tracking


icefoxen:
A collection of the things worth keeping an eye on for winit integration:

 * ~~#388 (Ratysz's PR)~~
 * ~~https://github.com/tomaka/winit/issues/540~~
 * ~~https://github.com/tomaka/winit/pull/548~~

francesca64:
I was also asked to improve `get_available_monitors`/`get_primary_monitor` ergonomics for this, which is very easy but I keep forgetting about it.

Ratysz:
Oh! I am so sorry, I wanted to open an issue about that, and now it's 1 am again. I'll do that as soon as I'm back at the workstation tomorrow, if it's not there prior; I'll provide any details I can, regardless.

Ratysz:
Opened tomaka/winit#555.

Ratysz:
(sort of TL;DR is at the bottom)

So, changes that had to be made, and issues that cropped up, in no particular order:

- `TextEditing` event is gone (that's the one relaying candidate Input Method Editor string). Not sure how big of a loss it really is; `TextInput` remains, now uses Unicode `char`s instead of a `String` (which shouldn't cause any problems).

- Controller support is gone, as it was entirely `sdl2`-dependant; possible replacement is [`gilrs`](https://crates.io/crates/gilrs). I've left reexports (and half-brained usage) of `winit::AxisId` and `winit::ButtonId` in, have to verify if they're really of any use.

- Events in `event::run()` loop in general need some cross-referencing with `winit::Event`. There might be more useful ones that I haven't made use of (`WindowEvent::DroppedFile` for loading savegames or something could be neat).

- `EventHandler` will likely need a touch-up as a result of the above. So far, mouse callbacks have their coordinates changed to `f32` from `i32`, `MouseState` is gone, some more minor things like that.

- Entry point for the lib has changed: methods instantiating `Context` now return `(Context, winit::EventsLoop)`, `event::run()` now takes `Context`, `EventsLoop`, and `State`. This is due to self-borrow that would happen if `Context` owned the loop it's called in. `Context` now has a `continuing` field, to allow `Context::quit()` from anywhere in in the `run()` loop to terminate the loop.

- `MouseContext` now tracks cursor position and cursor movement delta, to use with relevant callbacks. I also ended up implementing a `KeyboardContext` that tracks last held down key - to allow filtering of repeated keycodes.

[...]

PR #403, opened 2018-06-14
`conf` module improvements

Ratysz:
This is a running PR to track changes to `conf` module, mentioned in #391 [(this comment)](https://github.com/ggez/ggez/issues/391#issuecomment-396045308).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ smart-default = "0.2"
 nalgebra = {version = "^0.15.2", features = ["mint"] }
 # Has to be the same version of mint that nalgebra uses here.
 mint = "0.5"
-winit = { version = "0.15", features = ["icon_loading"] } 
+winit = { version = "^0.15.1", features = ["icon_loading"] }
 
 [dev-dependencies]
 chrono = "0.4"
--- a/examples/logging.rs
+++ b/examples/logging.rs
@@ -172,7 +172,7 @@ pub fn main() -> GameResult {
     // This sets up `ggez` guts (including filesystem) and creates a window.
     let (ctx, events_loop) = &mut ContextBuilder::new("logging", "ggez")
         .window_setup(
-            WindowSetup::default().title("Pretty console output!"), //.resizable(true), TODO: this.
+            WindowSetup::default().title("Pretty console output!").resizable(true),
         )
         .window_mode(WindowMode::default().dimensions(640, 480))
         .build()?;
--- a/examples/render_to_image.rs
+++ b/examples/render_to_image.rs
@@ -75,7 +75,7 @@ impl event::EventHandler for MainState {
 
 pub fn main() -> GameResult {
     let mut c = conf::Conf::new();
-    //c.window_setup.resizable = true; TODO: this.
+    c.window_setup.resizable = true;
     let (ctx, events_loop) = &mut Context::load_from_conf("super_simple", "ggez", c)?;
     let state = &mut MainState::new(ctx)?;
     event::run(ctx, events_loop, state)
--- a/examples/text_cached.rs
+++ b/examples/text_cached.rs
@@ -206,7 +206,7 @@ impl event::EventHandler for App {
 pub fn main() -> GameResult {
     let (ctx, events_loop) = &mut ContextBuilder::new("text_cached", "ggez")
         .window_setup(
-            WindowSetup::default().title("Cached text example!"), //.resizable(true), TODO: this.
+            WindowSetup::default().title("Cached text example!").resizable(true),
         )
         .window_mode(WindowMode::default().dimensions(640, 480))
         .build()?;
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -4,7 +4,7 @@
 //! A `Conf` struct is used to specify hardware setup stuff used to create
 //! the window and other context information.
 //!
-//! By default a ggez game will search its resource paths for a `/conf.toml`
+//! By default a `ggez` game will search its resource paths for a `/conf.toml`
 //! file and load values from it when the `Context` is created.  This file
 //! must be complete (ie you cannot just fill in some fields and have the
 //! rest be default) and provides a nice way to specify settings that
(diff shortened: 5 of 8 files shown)


Ratysz:
Ran into something unexpected: apparently, `toml` crate [does not support](https://docs.rs/toml/0.4.6/toml/ser/enum.Error.html) serializing enums, tuples, and tuple structs. Which is weird, considering that `FullscreenType` is an enum, and it has no problem de/serializing it's `::Off` variant. It does complain about any other variant, though.

icefoxen:
Part of the problem is basically that rust has no way of distinguishing "c-like" enums which have no additional data attached from enums where some variants include tuples or structs. This makes life hard for some serializers.

Ratysz:
Yes, I sort of understand the why of it. I'm putting together a workaround that lets us keep the tuples and everything (it involves some manual `serde` sorcery).

icefoxen:
An easier option might be to just have accessor and builder methods to manage the tuples, or just drop them altogether?  :man_shrugging:

Ratysz:
This should be pretty much it. There are a few more settings I decided are not worth exposing: `gl_debug_flag`, `gl_robustness`, `stereoscopy`. First two aren't useful in context of facaded GL, the last one is too niche.

icefoxen:
I've been putting off because I expected it might be the case, but...  I like the changes in general but I hate the TOML it results in:

```toml
[window_mode]
maximized = false
hidden = false
borderless = false
always_on_top = false

[window_mode.dimensions]
option = "Some"
width = 800
height = 600

[window_mode.min_dimensions]
option = "None"

[window_mode.max_dimensions]
option = "None"

[window_mode.fullscreen_type]
type = "Off"

[window_setup]
title = "An easy, good game"
icon = ""
resizable = false
transparent = false
use_compatibility_profile = false
vsync = true
samples = "One"
srgb = false

[backend]
type = "OpenGL"
major = 3
minor = 2
```

I don't really consider the options to be hand-editable... if I want to change `max_dimensions` to be something, what am I supposed to do?  Change `option` to `Some`, obviously, and then....?  If I want to change `fullscreen_type` to be `Desktop` and have to put in a `MonitorId`, then what do I do there?

There has to be a way to make it more TOML-y...

Ratysz:
Now that you mention it, yeah, that is pretty bad. Somehow it never occurred to me to *really* look at it from user side - I was operating from assumption that most folks will work with `Conf` from Rust side, which is just not true. And, I was (still am) pretty angry at TOML's whitespace dependence.

I'll try and fix this, probably using a full intermediate struct - so that both TOML side and Rust side make sense in their own context.

Ratysz:
Done. Structs are largely unchanged since last time, but TOML now looks like this:
```toml
[window_mode]
always_on_top = true

[window_mode.dimensions]
width = 640
height = 480

[window_mode.min_dimensions]
width = 640
height = 480

[window_mode.fullscreen]
type = "Desktop"
monitor = 1

[window_setup]
title = "Testing, testing, one, two"
samples = "Four"
vsync = false

[backend]
type = "OpenGL"
version_major = 4
version_minor = 6
```
Not all fields are present because I went ahead made everything optional at TOML side. If a field is not defined, it will initialize with default value. If this is something we *don't* want, it's easy enough to change back.

Wrote a fair bit of examples/doctests - they're easy to maintain, since compiler itself will point out if something is wrong during testing. Can be explored rendered [here](http://ratysz.github.io/ggez/conf/index.html).

Terkwood:
Nice improvement!

icefoxen:
> If this is something we don't want, it's easy enough to change back.

I'd prefer having all fields have explicit values instead of "field is absent" meaning "use default value".  Makes it easier for a user to look at the config file and see everything that's going on, and it protects against defaults potentially changing.

icefoxen:
@Ratysz Some rather belated feedback...

 * Always prefer implementing `From` rather than `Into` where possible.
 * Make everything as simple as possible, but no simpler.  :-)  For example, storing `MonitorId` for `FullscreenType` instead of just using `Current` everywhere is a lot of complexity for quite little gain... I don't recall many games letting you select which monitor you want to display on.  Trying to shove `Option` types into TOML which isn't designed for them is quite a lot of complexity for little gain.  

Upon consideration I'm going to have to sit and think about this more, and ponder whether 600 lines of code for ultimately cosmetic improvements is really worth it.    There's *got* to be a better way... and maybe it's as simple as "use something more powerful than TOML".  [ron](https://github.com/ron-rs/ron), perhaps?

Ratysz:
> I'd prefer having all fields have explicit values instead of "field is absent" meaning "use default value".

My reasoning was, we're not actually generating the `conf.toml` file, unless user app does it itself, and the only way for users to know which fields they need is to look in documentation anyway (barring trial and error). Now, not saving default fields with `Conf::to_toml_file()` likely *is* a bad call.

> [...] protects against defaults potentially changing.

Yes, I haven't even considered that.

> Always prefer implementing `From` rather than `Into` where possible.

I was led to believe these are completely equivalent, as one always implicitly implements the other. Or is it for clarity?

>  [...] using `Current` everywhere [...]

Simplest solutions are the best... There's a long line of thought that can be summed up as me latching onto supporting as much of `winit` functionality as possible. This will also simplify a couple of points from the integration tracking issue.

> [...] quite a lot of complexity [...]

I'd argue it's more verbosity, rather than complexity. The `custom_ser_de` submodule (the 600 lines) is quite possibly over 1/3 whitespace, and the rest is hand-rolling what `#[serde(bound(...))]` does - because it requires the specified types to be `pub`, thus visible in API.

> [...] "use something more powerful than TOML".

That would allow using `serde` derives as-is, and completely throwing away `custom_ser_de` submodule.

I knew soon after I saw `toml` crate error on my freshly-refactored types that I was asking of it something that it wasn't meant to do. Unfortunately (?), I found ways around those limitations before resigning and reconsidering; not the first time I got carried away 😄 

I'll see if I can pick this apart and create a smaller set of changes; mostly docs.

PR #406, opened 2018-06-15
`keyboard` module improvements

### LINKED ISSUE #404: Keyboard state API
It would be really useful to have a keyboard state API, similar to the mouse and gamepad state API. This would make writing keyboard-driven games significantly easier.

Ratysz:
Resolves #404.

`keyboard` module now exposes functions:

- `get_active_mods()` - returns currently active keyboard modifiers.
- `get_pressed_keys()` - returns a slice with currently pressed down keys.
- `is_key_pressed()` - checks if a key is currently pressed down.
- `is_key_repeated()` - checks if the last keystroke sent by the system is repeated, like when a key is held down for a period of time.
- `is_mod_active()` - checks if keyboard modifier (or several) is active.

`KeyMods` is now a local bitflags struct, can be converted from `winit::ModifiersState` - this allows neat matching and keybinds (not included); and, this is how we had it with `sdl2`.

It has an extra method, `KeyMods::count()` - returns amount of flags set; this is useful for mapping modifier+modifier+key in a non-exclusive way ([example](https://github.com/Ratysz/spelspell/blob/master/src/input.rs)).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ default = ["bzip2"]
 multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
 
 [dependencies]
+bitflags = "1.0"
 zip = { version = "0.4", default-features = false }
 app_dirs2 = "2"
 gfx = "0.17"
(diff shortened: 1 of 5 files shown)


Ratysz:
Apparently! Removed it.

PR #413, opened 2018-06-27
Update to `winit` 0.16, `glutin` 0.17

Ratysz:
Tracking PR, of a sort.

`winit` 0.16 [has been released](https://github.com/tomaka/winit/blob/089816d9bab498c0ec002d7e560ad940b78d476e/CHANGELOG.md); waiting on tomaka/glutin/pull/1035.


Ratysz:
... why? It doesn't even compile yet. We'll need to update `gfx_window_glutin` as well, and it's not trivial (to do properly; as far as I can tell).

icefoxen:
Oh.  Derp.  I thought it worked with glutin updated.  I'll work on updating `gfx_window_glutin` then!

PR #433, opened 2018-07-16
Fix variable naming consistency in tutorial

sh1mmer:
Currently the code snipper to initialize state assigns it to `s` but later it's referred to as `state`. This PR changes it to be initialized as `state` as well.


icefoxen:
Thank you!

Issue #265, opened 2018-01-26
Adding Custom Cursor Support


Manghi:
Something I want to is to be able to change how the cursor appears depending on what UI element I am interacting with. I dug around in the 0.4 documentation but couldn't find anything so I decided to add this myself and give back to the community for a change! 😄 

I have something that can build and run, but it isn't working as I expect. My code changes have no effect on the cursor. I saw that rust-sdl2 added their own cursor example about a month ago and confirmed that it does operate as intended. I ended up following their example fairly closely as I kept trying things out.

My on-going changes can be found here: 
https://github.com/Manghi/ggez.git

My current suspicion is that `cursor.set()` isn't actually setting the cursor, but I don't see any error reported up until that line; as far as I can tell, the cursor is created properly for both the `from_surface()` and `from_system_cursor()` attempts. 

Any insight or help would be appreciate.

icefoxen:
Nice!

Only advice I can offer, for the easy(ish) way to turn image data into an sdl2 `Surface`, see https://github.com/ggez/ggez/blob/master/src/context.rs#L58-L83

Manghi:
Well `surface::Surface::from_file(image)` takes a path of an image and does the rest, so that part wasn't difficult. I thought perhaps there may be some ggez/SDL2 link I'm missing. 

I'll keep at it.

Manghi:
Just a thought, could this be because `gfx` is the rendering engine? I'm wondering if the modifications to the cursor are not being propagated to `gfx`, if that is the case. I have no idea how to actually check this though.

I'll try out the `Cursor::from_data` approach tomorrow just to see if not incorporating `sdl2_image` does anything.

Manghi:
Tried without `sdl2_image` with the method @icefoxen described in his post but no luck. I even back-ported these changes into to a local `ggez 0.2.2` instance just to see if it was due to the switch to gfx as a backend, also no luck. Not sure what else to dig into at the moment.

Manghi:
Looks like its a use-after-free bug in `rust-sdl2` where the Surface is dropped too soon. Moving this logic into `main` so that it remains for the duration of the program fixed it. Obviously this isn't the best solution and it will need to be fixed in that library. What is that fix exactly? I'm not sure just yet. Guess this effort will remain on hold until then.

icefoxen:
If you can make a reproduction of the use-after-free bug I bet the sdl2 crate maintainers would love to hear about it.

I'll see if I can find time to play with this more.

[...]

PR #436, opened 2018-07-18
Add cursor visibility functions

expenses:
Hi, this is kinda related to #265, but this allows adds new functions to set/get whether the cursor is hidden on the screen. It's my solution of allowing for a custom cursor one, but it might not be the best one.


icefoxen:
Thank you!  I could have sworn we had this already...

It's a little awkward, since I sort of wasn't intending to actually make another release of the 0.4 branch, and 0.5 uses winit rather than SDL so this isn't really useful for it...  But there's been one or two other minor bugfixes so maybe it's worth it.  Might get to it later tonight.

PR #437, opened 2018-07-18
[WIP] Audio rework

### LINKED ISSUE #430: Rework audio
https://github.com/ggez/ggez/issues/234

mbillingr:
Resolves #430.

Features included so far:
- [x] spatial sounds
- [x] different play modes (replace, detach, enqueue)
- [x] fading in
- [x] adjustable pitch
- [ ] <s>`PeriodicAccess` based timing timing</s>

Caveats:
- There are two types of sources now: `Source` and `SpatialSource`. This incurs plenty of code duplication but it was the easiest approach for a first prototype.
- Rodio has a very strange spatial model. They basically simulate [A-B recording](https://en.wikipedia.org/wiki/Stereophonic_sound#A-B_technique:_time-of-arrival_stereophony) *without* phase differences. Tuning spatial audio for the game scene is very awkward. [X-Y recording](https://en.wikipedia.org/wiki/Stereophonic_sound#X-Y_technique:_intensity_stereophony) would be more appropriate here.
- `.stop()` and `.play_*()` take `&mut self` because they need to reconstruct the `Sink` internally. These functions use rodio's default device because they don't have access to the context. This may be fine now but invites problems in the future.
- I experimented with `PeriodicAccess` to update an atomic counter (see my attempt [here](https://github.com/mbillingr/ggez/blob/audio_timing/src/audio.rs). Results were very inaccurate so I decided to skip this feature.
- No tests. How do you test audio, anyway?

I'm no good in API design or coding style. Feedback is very much appreciated.


mbillingr:
Arrgh! That shall teach me not to run tests locally...

icefoxen:
Awesome! At a first glance it looks quite good, apart from a couple hairy bits you have noted as such. I'm taking a bit of a break from ggez so it'll be a few days before I take a deep look.

mbillingr:
Sure, take your time.

Btw, just realized that the timing inaccuracies are probably caused by `thread::sleep_ms()` and not by `PeriodicAccess`.

mbillingr:
Ha! Nice lesson on premature optimization:

Only activate features we need? Dynamically compose sources? Afraid of expensive resampling?
Forget it all! Just throw in everything and let Rodio sort it out... which it does: [bruteforce works best](https://github.com/mbillingr/audio_benchmark).

mbillingr:
Almost ready... two open questions:
- Do you want to keep the mutating `play`/`stop` functions or should I throw in a `RefCell`?
- Do you want me to deal with the merge conflict? If so, do you prefer rebase or merge?

I'd like to have another go at timing, but this could go into another PR.

icefoxen:
Looks fine the way it is, I'll merge it and deal with the conflict.

Issue #440, opened 2018-07-24
Drawing thousands of rectangles is slow


beatgammit:
I have the following code ([full runnable gist](https://gist.github.com/beatgammit/0fb04d4ebed54a658aba607ab0916f47)):

    graphics::clear(ctx);
    for x in 0..50 {
        for y in 0..50 {
            let rect = graphics::Rect::new_i32(x*10, y*10, x+10, y+10);
            graphics::set_color(ctx, [(x as f32) / 50.0, 0.0, (y as f32) / 50.0, 1.0].into())?;
            graphics::rectangle(ctx, graphics::DrawMode::Fill, rect)?;
        }
    }
    graphics::present(ctx);

This loop takes ~35ms on my machine. I don't have anything fancy in my Cargo.toml, so I'm using default features, and this is running in release mode (debug mode is ~700ms).

Here's an example with `love` (takes <0.5-1.5ms most of the time):

    function love.draw()
      local start = os.clock()

      love.graphics.clear()
      for x=1,50 do
        for y=1,50 do
          local r = x/50
          local b = y/50

          love.graphics.setColor(r, 0, b, 1)
          love.graphics.rectangle('fill', x*10, y*10, 10, 10)
        end
      end

      print((os.clock()-start)*1000);
    end

I'm guessing drawing rectangles isn't quite optimized, but this seems quite a bit worse than "not optimized yet".

Am I doing something wrong? Or is this a bug in `gfx` or `ggez`?

Let me know what additional info you'd like to debug this. I've tried constructing the rect outside the loop and commenting out the `set_color` call, but the rendering time is still the same.

Dev system specs:

- CPU: Ryzen 1700 (3 GHz, 8 cores)
- GPU: Nvidia 960 GTX, proprietary drivers on Linux

I get similar results on Windows.

termhn:
This is a matter of love automatically doing things behind the scenes to
speed this up for you (batching draw calls). Any time you make 2500 draw
calls in a single frame it’s gonna be slow. Take a look at the spritebatch
example to see how to use SpriteBatches which will speed this up immensely.

beatgammit:
Yeah, I'll definitely use a sprite batch soon, but I was mostly concerned with the performance hit. Is `ggez` really not doing *any* batching at all? Is it basically making a separate call to the GPU at every iteration?

I guess I assumed that the `Context` was doing some kind of batching. If that's really all it is, I'll just port my code to use SpriteBatch and call it a day. :)

[...]

Issue #346, opened 2018-04-08
Automatic sprite batching


termhn:
So, when we originally implemented batching, this was out of scope, but... it recently landed as a feature in Lua and I'm interested in trying to implement it in ggez now that I have more experience with graphics programming in general. Thought I'd open an issue as far as comments on whether people think this would be a good thing to have and if so thoughts on implementation details if anyone has them.

icefoxen:
There's upsides and downsides.

Upsides: Drawing stuff becomes faster without anyone doing anything special...

Downsides: ...only sometimes, when they do drawing in a way that the sprite batcher can recognize.

I'm generally in favor but a little skeptical of sneaky under-the-hood optimizations.  OpenGL is a good example of the sort of mess that can lead to.

That said, LOVE doing this is generally a decent sign that it's doable nicely.  We should take a look at the LOVE 11.0 source code and see if there's anything particularly hairy in it, just in case.

dannyfritz:
what about doing it as a macro block that batches the calls inside of it?

termhn:
@icefoxen True, however I think this is a pretty standard optimization to do. In addition to LOVE, [three](https://github.com/three-rs/three) fairly recently implemented automatic instancing which I was thinking I might use as some amount of reference.

@dannyfritz Hmm, not entirely sure how that would work or if it could really work.

icefoxen:
@dannyfritz Doesn't seem like that would gain much over just using a `SpriteBatch` anyway, apart from maybe a little syntactic sugar...

@termhn Yeah, that's true.  It certainly won't hurt, and making the rules of what can and can't be instanced clear in the docs is easy enough.  Let's try it and see how it works.

dannyfritz:
Would that also flush the draw calls for images when doing other types of draws?

onelson:
Would this devalue callers using SpriteBatch directly?
(from someone with a budding sprite anim library that won't go anywhere)

Ratysz:
If this happens, we can give `Text` similar treatment, easily: have all draw calls on it only queue, and draw the entire queue after `EventHandler::draw()`. `gfx_glyph` handles all the caching and batching internally.

Obviously, we'll need to expose those centralized batch draw methods, for those rolling their own event loops.

icefoxen:
@dannyfritz Probably, to make sure draw ordering is correct.

[...]

PR #442, opened 2018-07-25
Add docs to indicate that primitive draws are inefficient

beatgammit:
The snake example is referenced from a youtube video and a blog post, so
I added documentation there because that's likely one of the first
places a new developer will interact with ggez and base projects on.

I noted in the graphics package that the draws are different from the
love project that it's based on since primitive draw calls seem to be
batched, whereas ggez does *not* do automatic batching. There may be
more places in the source where such notes are useful.

I also added a short note in the README that there may be performance differences between ggez and love. I feel like it would be nice to have a section that details gotchas coming from love. I personally don't have a *ton* of experience with love (or ggez for that matter), but my knee-jerk reaction in #440 was to test on love to see if I was making a fundamental mistake. Having a note in the README about high-level differences from love would be helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,9 @@ More specifically, ggez is a lightweight cross-platform game framework
 for making 2D games with minimum friction.  It aims to implement an 
 API based on (a Rustified version of) the [LÖVE](https://love2d.org/) 
 game framework.  This means it contains basic and portable 2D
-drawing, sound, resource loading and event handling.
+drawing, sound, resource loading and event handling, but finer details
+like performance characteristics may be very different (e.g. ggez does
+*not* do automatic batching).
 
 ggez is not meant to be everything to everyone, but rather a good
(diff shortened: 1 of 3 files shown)


icefoxen:
Nice, thank you!  Though #346 might be interesting to you.

termhn:
out of curiosity @beatgammit which youtube video is the snake example referenced from?

beatgammit:
@termhn I was referring to the [snake example in `piston`](https://m.youtube.com/watch?v=HCwMb0KslX8#) that the snake example for `ggez` was based on. The comments in the source for the example reference it and the github repo.

I don't know if there's a YouTube example for doing this in `ggez`, though I did consider `ggez` because of that example.

termhn:
Nice, okay gotcha. I wrote the snake example after seeing that piston one
but wasn’t sure if someone had made another video that it was included in
:) glad to hear it’s helpful haha

Issue #422, opened 2018-07-07
Rendered text width is 1 pixel larger than expected


dlight:
Tested with multiple fonts, on both 0.4 and master. Here's a test case:

```rust
extern crate ggez;
use ggez::*;

pub fn main() -> GameResult<()> {
    let ctx = &mut Context::load_from_conf("test", "me", conf::Conf::new())?;
    let font = graphics::Font::default_font()?;

    let text = "Hello There";

    let expected_width = font.get_width(text);
    let rendered_width = graphics::Text::new(ctx, text, &font)?.width();

    println!("Text: {:?}, expected: {}, rendered: {}", text, expected_width, rendered_width);
    // prints: Text: "Hello There", expected: 86, rendered: 87

    Ok(())
}
```

icefoxen:
Reproduced, thank you.

PR #443, opened 2018-07-25
Fix the text metrics test

beatgammit:
I'm not sure whether these test failures indicate an error, but they
were only off by one on every machine tested (my machine, app veyor,
etc), so I'm guessing something changed elsewhere in the system.

At the very least, this fixes the


icefoxen:
Thank you!  Apparently when #422 was fixed the unit tests didn't get corrected.  Could have sworn I did that, but...

PR #445, opened 2018-07-28
Fixed double-borrow in text.rs

reujab:
`self.calculate_dimensions` hangs on a `.write()` to `self.cached_metrics` without this fix due to a double-borrow.

### DIFF
--- a/src/graphics/text.rs
+++ b/src/graphics/text.rs
@@ -352,9 +352,8 @@ impl Text {
     /// TODO: Should these return f32 rather than u32?
     pub fn dimensions(&self, context: &Context) -> (u32, u32) {
         if let Ok(metrics) = self.cached_metrics.read() {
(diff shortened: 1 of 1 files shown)


src/graphics/text.rs
```diff
@@ -353,13 +353,11 @@ impl Text {
     pub fn dimensions(&self, context: &Context) -> (u32, u32) {
         if let Ok(metrics) = self.cached_metrics.read() {
             match (metrics.width, metrics.height) {
```
Ratysz: Nitpick; this could be an `if let`:
```rust
if let (Some(width), Some(height)) = (metrics.width, metrics.height) {
    return (width, height);
}
```

reujab: @Ratysz fixed

icefoxen:
Thank you!

PR #448, opened 2018-07-31
Mount paths before searching for conf file in ContextBuilder

harris-m:
I was trying to set up a `conf.toml` file in a `resources/` directory next to my `Cargo.toml`, and I noticed that it was not being loaded by ggez.

It seems that `ContextBuilder` tries to load a conf file before mounting any paths that have been added to it, missing any confs that may be in those paths.


harris-m:
Woops - this PR is targeting `master`, will make a new one for `devel` instead.

PR #449, opened 2018-07-31
Mount paths before searching for conf file in ContextBuilder

harris-m:
Same as #448 but on the right branch this time.


icefoxen:
Awesome, thank you!

Issue #431, opened 2018-07-11
Add moar unit tests


icefoxen:
Especially for mouse and gamepad.

https://github.com/ggez/ggez/issues/234

onelson:
I spent a little time trying to get kcov working so we could have some measurements.

It can be nice to know what areas to focus on via coverage reporting, but kcov seems to segfault on the ggez suite. Mainly I guess the report is useless. Best bet will be to design the suite by eye and feel.

In case it's interesting, here's what the output looked like.

```
kcov --verify --exclude-pattern=cargo target/cov target/debug/ggez-78e113c1ebdd0c20
kcov: 213 invalid breakpoints skipped in /home/owen/projects/ggez/target/debug/ggez-78e113c1ebdd0c20
kcov: 1 invalid breakpoints skipped in /lib/x86_64-linux-gnu/libdl.so.2
kcov: 3 invalid breakpoints skipped in /lib/x86_64-linux-gnu/librt.so.1
kcov: 6 invalid breakpoints skipped in /lib/x86_64-linux-gnu/libpthread.so.0
kcov: 2 invalid breakpoints skipped in /lib/x86_64-linux-gnu/libc.so.6

running 19 tests
test filesystem::tests::headless_test_file_not_found ... ok
test filesystem::tests::headless_test_file_exists ... ok
test filesystem::tests::headless_test_read_dir ... ok
kcov: Process exited with signal 4 (SIGILL) at 0x7ffff67aa6f3

kcov: Illegal instructions are sometimes caused by some GCC versions
kcov: miscompiling C++ headers. If the problem is persistent, try running
kcov: with --verify. For more information, see
kcov: http://github.com/SimonKagstrom/kcov/issues/18
```

![coverage report top level](https://i.imgur.com/CXE7MQU.png)

icefoxen:
Hmmm, interesting.  I tried `kcov` built on my local system from latest git master and it didn't crash, though it did report `kcov: 213 invalid breakpoints skipped in /home/icefox/my.src/ggez/ggez/target/debug/ggez-eb3008806bb6209a`.  The coverage report I got was both similar and different than yours, for example it has the same results for `mouse.rs` but different ones for `gfx_gl.../out/gl_bindings.rs` and `input/keyboard.rs`.

Not really sure what to make of it, I confess.  Not finding many results for using anything *other* than `kcov` with Rust, either.  Hmmmmm.

onelson:
[...]

Issue #425, opened 2018-07-07
Make sure getters/setters are named consistently.


icefoxen:
Just go through everything and check.  Lots of API changes have happened recently

Though there's some friction between Love2D's `get_foo()` and the more Rust-y `foo()`, which tends to be a bit context-dependent whether it's a getter or setter (builders seem different than the usual).

Ratysz:
For what it's worth, I'm partial to `get_*` (or `is_*` for boolean returns) and `set_*`, or `with_*` for builders. Can't rely on IDE or context tells alone.

termhn:
I think in Rust it’s the generally accepted style guide that thing.noun()
is a getter, and thing.get_noun() is somewhat frowned upon.

icefoxen:
For reference, winit uses `with_*` for builders, which I kind of like since it distinguishes them from non-builder setters.

beatgammit:
Is there consensus on this? I'm willing to work on this, but I need to know if we're going with the Love2D conventions or the more Rusty syntax.

I prefer to use the idioms of the language I'm working in, but I'm happy to help out with whichever direction we choose to go.

My personal motivation here is to get a release with winit support so I can work on getting my project working on web and desktop platforms before it gets too complicated. This seems like a task that's blocking release and that I can help out with :)

onelson:
I'd defer to https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis and favor `.foo()`, `.set_foo()`, and where required `.foo_mut()`

There's going to be a bigger payoff for using rust idioms than there is for "following the letter of the law" set by Love2D's api.

beatgammit:
If I don't hear back on this issue, I might start working on this by following the Rust conventions. Feel free to complain in the PR if you really like the Love2D naming convention.

I'll probably go module by module, adding some simple tests if they're super easy.

icefoxen:
Yeah, let's follow the Rust conventions then.  Thank you, @beatgammit !

icefoxen:
Implemented and finished in #467 but it's a BIG change so I kinda want people to look at it and make sure it looks sane.

PR #450, opened 2018-08-01
Rename pub(crate) keyboard getters

beatgammit:
I added a unit test for a missed function, though it's a pretty lame test :wink: 

I didn't change the `pub` functions because I wanted to get feedback on whether all of these changes should go in the CHANGELOG individually (it seems there's ~30) or if there should be a single line there that covers all of them.

If you want to merge this, that's fine, we'll just have to do all the `pub` fns later (perhaps in one PR?).

### DIFF
--- a/src/input/keyboard.rs
+++ b/src/input/keyboard.rs
@@ -188,14 +188,14 @@ impl KeyboardContext {
     }
 
     pub(crate) fn is_key_repeated(&self) -> bool {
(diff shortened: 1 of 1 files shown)


beatgammit:
#425 #431

icefoxen:
We *definitely* do not need to mention every function in the CHANGELOG individually.  A single line covers them.

PR #451, opened 2018-08-01
Fix a bunch of clippy warnings

beatgammit:
Please look through this before merging, I made a couple of style choices based on clippy that may or may not be desirable.

This fixes a good number of the lint warnings. The ones remaining are:

- type_complexity
- trivially_copy_pass_by_ref (I left these out since it's an API change)
- too_many_arguments
- new_ret_no_self
- unnecessary mut (looks like it's a work in progress)

I'm not sure if the goal should be 100% clippy pass, but I think the `type_complexity` ones should be looked at.

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -37,7 +37,7 @@ impl AudioContext {
                 "Could not initialize sound system (for some reason)",
             ))
         })?;
-        Ok(AudioContext { device: device })
+        Ok(AudioContext { device })
     }
 }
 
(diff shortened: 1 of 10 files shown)


icefoxen:
Awesome, thank you!

So far my goal has been "100% clippy pass except when it's silly or unavoidable", which I know isn't really helpful.  Notably, there's some cases where `too_many_arguments` and `type_complexity` aren't part of the external API, and are more or less unavoidable... or at least, I feel like changing things to avoid them would only actually make things more complicated.

I'll definitely take a look at the remaining lints and see what's up, thank you!

Issue #353, opened 2018-04-14
Allow getting timing data from audio playback


cessen:
The audio APIs in ggez currently don't allow any way to get timing data of audio that is playing.  This is important for e.g. rhythm games, which need to sync events with the audio that is playing.

Rodio has an API that can be used for this purpose: the [PeriodicAccess](https://docs.rs/rodio/0.6.0/rodio/source/struct.PeriodicAccess.html) source.  It allows hooking in a closure to be executed at regular intervals during the audio playback.  Combined with an atomic integer, this can allow tracking how much of the audio has been played so far.

I'm thinking that it probably makes sense to just expose a similar API, to allow people to do whatever they like at regular intervals with the audio.  It solves this issue, but may also allow for some other interesting uses (though I'm not quite sure what they would be).

cessen:
I'll take a crack at implementing this when I have the time.  But if anyone gets to it before me, I certainly won't be offended. :-)

icefoxen:
@cessen Go right ahead!  Audio really needs the help right now, alas.

nobbele:
This has been implemented, right?

PR #454, opened 2018-08-06
Audio timing API

mbillingr:
Implements audio timing as discussed in #353. This feature was missing in #437 but was easy enough to add.

I have set the default query interval to 100ms. This should add little overhead for users that don't need this feature (10 `Arc` derefs and atomic increments per second and sound). The interval can be adjusted to get higher precision or better performance.

cc @cessen

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -11,6 +11,7 @@ use std::mem;
 use std::path;
 use std::time;
 
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 
 use mint;
@@ -122,6 +123,8 @@ pub struct Source {
     repeat: bool,
     fade_in: time::Duration,
     pitch: f32,
+    query_interval: time::Duration,
+    play_time: Arc<AtomicUsize>,
 }
 
 impl Source {
@@ -142,6 +145,8 @@ impl Source {
             repeat: false,
             fade_in: time::Duration::from_millis(0),
             pitch: 1.0,
+            query_interval: time::Duration::from_millis(100),
+            play_time: Arc::new(AtomicUsize::new(0)),
         })
     }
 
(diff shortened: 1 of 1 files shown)


icefoxen:
Does an `AtomicUsize` really need to be `Arc`'ed?  Guess it depends on whether we want a `Source` to be an object, or a handle to an object.  I was leaning towards the former, especially given `Source` doesn't even impl `Clone`...

mbillingr:
I'd be happy to avoid the `Arc`, but I do not know how to give both the audio thread and the `Source` object access to the same `AtomicUsize` without running into lifetime issues...

The problem here is that we do not know who lives longer audio thread or `Source`. Shared ownership of the counter is the only solution I could think of.

icefoxen:
Yep, my bad, I was `AtomicUsize` avoids a `RefCell` or such, not an `Arc`.  Derp.

Might it be easier though to have a `Source` store the time that it started playing?

Sorry for taking so long to get to this.

mbillingr:
I don't see how storing the start time could help here. As far I understood #353  the point was not to depend on the system clock because it drifts compared to the sound cards sample rate.

Maybe it's not worth to implement this very specific functionality. Could it be an alternative to expose Rodio structures through `Source` (not sure if that's easily possible) so that downstream users can implement such functionality if they need it?

icefoxen:
> As far I understood #353 the point was not to depend on the system clock because it drifts compared to the sound cards sample rate.

Derp, my bad then.  Thanks.

> Maybe it's not worth to implement this very specific functionality. 

Eeeeh.  This system seems to do the Right Thing in that case, and doesn't add much extra complexity or overhead, and timing-exact audio is both important (for things like rhythm games), and hard to do well (afaik there's actually no possible way to make it work perfectly on Android, though that may be out of date now).  We'll roll with it by now and see how it goes.

Thank you again!

Issue #435, opened 2018-07-16
Transforms are broken -- tracking issue


icefoxen:
This has been the case for a while and I thought "oh it's just a quick fix" and I put it off forever and tried to mess with it yesterday and couldn't fix it so it's obviously not quick so I need to stop putting it off and give it its own issue.

Basically, in `Image::draw()`, the `offset` value is one way or another not being applied correctly.

There's also the various breakage with canvases and such, and in general all the drawing transform code needs to be looked over to make sure it Actually Works Right.

This is irritating since it's kinda spread out in various places and I don't understand how it composes well enough, so blah.

icefoxen:
shear should probably be a Vector2 also.

icefoxen:
Made a `transforms` example to have a nice way to play with this.  Results:

 * dest and offset are in units of pixels
 * shear and src are in units of image sizes

We want to be able to support both but not at the same time.

icefoxen:
Text with `gfx_glyph` appears to not be affected by `graphics::set_screen_coordinates()`.

Things also get screwy as heck with dpi involved, sigh...

This is what I get drawing labels and points every 100,100 pixels on an 800x600 window on a display with dpi factor of 1.5...

![image](https://user-images.githubusercontent.com/1335133/43555690-ccf0e374-95c9-11e8-9ffe-50db2fec318b.png)

icefoxen:
@termhn You're way better at math than me.  I'll buy you a cake if you help me with this, 'cause by now it's all some twisted mess and I can't figure out what's happening where, let alone what *should* be happening where.

icefoxen:
Okay, so per #439 there's issues with the offset matrix calculation:

    translate * offset * rotation * scale * offset_inverse

(shear omitted for simplicity)

and it should be

    translate * offset * rotation * offset_inverse * scale

In 0.4 that code is [here](https://github.com/ggez/ggez/blob/master/src/graphics/drawparam.rs#L78), in 0.5 it is in the [same file but slightly different place](https://github.com/ggez/ggez/blob/devel/src/graphics/drawparam.rs#L212).

So I implemented this change in 0.5 and there's TWO issues

First, the offset isn't scaled properly by the size of the image, which is adjusted [here](https://github.com/ggez/ggez/blob/master/src/graphics/image.rs#L307).  

[...]

PR #458, opened 2018-08-10
Fix draw_queued_text

robojumper:
Math is a bit complicated and involved some trial and error. The speedup from precalculating the transform matrix isn't that big, but I'd feel dirty with so many unnecessary matrix allocations.

Note that this doesn't fix text size on higher dpi levels.


icefoxen:
Oh thank god

I wanted to do this and also handle https://github.com/ggez/ggez/issues/435 better before this weekend's Ludum Dare game jam, but...

Issue #455, opened 2018-08-06
Create Image from &[u8]


mh84:
I guess there should be a possibility to create an Image directly from a given byte buffer for some reason, mine would be that i would like to store all game related data (including images) in an sqlite database, so creating a method from_bytes would be appreciate.

If this would be agreed i will take care of this soonand  create a pull request providing this "feature" (if anyone would count 10 lines of rearranged code a new feature ;) )

With best regards!

mbillingr:
This sounds like [`Image::from_rgba8()`](https://docs.rs/ggez/0.4.3/ggez/graphics/type.Image.html#method.from_rgba8) might work for you?

mh84:
Not completely, I've seen that and had to `image::load_from_memory(&buf)?.to_rgba()` (from [Image::new()](https://docs.rs/ggez/0.4.3/ggez/graphics/type.Image.html#method.new)) myself because of storing .png files there directly, so i had to add this additional dependency, which i surely like to avoid for just that one call. Another option could be to re-export `image`, that would be okay too, or maybe even better though its image manipulating strength.

termhn:
You say “just for that one call” but that one call under the hood is doing
a lot of stuff. If the image crate works excellently for all this stuff and
rustc is gonna elide all the extra code you don’t use anyway I don’t see
any reason we need to include it in ggez itself.

mh84:
Sure, i will agree with your arguments, but i further would take in discussion that from users point of view, having an interface for creating `Image`s by loading files from disk and another creating them from a given `rgba &[u8]`, the middle part where having an bytestream for any source and being in need to create an `rgba &[u8]`, this could easily be provided through the library itself imho.

onelson:
I think the only real reason to re-export the `image` crate would be if the types from `image` were to show up in `ggez` function signatures, meaning calling code would have to align their dependency on `image` with whatever `ggez` depended on. I'm not sure this is even the case.

Is the need to use `image::load_from_memory().to_rgba()` due to a color depth/channel mismatch or something? Seems like `Image::from_rgba8()` is really meant to be the thing for this.

icefoxen:
I sympathize, but it does seem like the `image` crate is the right tool for the job of decoding images. ggez uses `image` already, so adding this functionality to ggez has no advantages to performance, binary size or compile time. It's just another function to maintain and document.

It's not really a big deal, so maybe it would be nice to have anyway. But it doesn't add much benefit either.

icefoxen:
[...]

PR #460, opened 2018-08-14
Added `Image::from_bytes()`

Ratysz:
Convenience function; allows easily baking-in sprites:
```rust
let house = Image::from_bytes(
    ctx,
    include_bytes!(concat!(env!("CARGO_MANIFEST_DIR"), "/assets/house.png")),
)?;
```
Same is possible using existing `Image::from_rgba8()`, but requires using `image` crate directly, to convert raw bytes into rgba8 + dimensions.

### DIFF
--- a/src/graphics/image.rs
+++ b/src/graphics/image.rs
@@ -172,6 +172,16 @@ impl Image {
         )
     }
 
+    /// Creates a new `Image` from the given `u8` buffer,
+    /// making an educated guess about the image format.
+    ///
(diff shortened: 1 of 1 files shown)


icefoxen:
I don't like this, for the reasons described in #455 .  But if people want it, we can add it.

Ratysz:
... Somehow I completely forgot about reading that discussion two weeks ago. I concur.

PR #467, opened 2018-08-22
Rename ALL getters from `get_foo()` to `foo()`.

icefoxen:
See issue #425 for discussion.

So far I rather like this change but it's going to live in its own branch for a little because it's VERY far-reaching.

I tried having builder structs use `with_foo()` instead of `foo()` but I find I like that rather less, even if it conflates the meanings a little.

This is also very search-and-replace-y so there may be errors or typos.

API docs for this change are available for now at https://alopex.li/temp/ggez/ggez/


icefoxen:
This hasn't destroyed my life yet, so let's roll with it.

Issue #473, opened 2018-09-19
Suggestion: adding ContextBuilder::set_config method


RGates94:
Sometimes, it would be useful to be able to set all `Conf` fields of a `Context` at once.  As the `Conf` struct already exists in ggez, a method like this seems like the most logical choice:

`ContextBuilder::set_config(self, config: Conf) -> Self`

Alternatively, this method could be implemented for `Context`, which would allow for changing these properties after the `Context` has been created.

icefoxen:
I'd considered something like this but hadn't really thought up a use case, but talking to OP on Discord came up with an interesting one: easily switching between different config files to test different pre-defined settings.  Let's do it.

icefoxen:
Resolved by #474 :heart:

PR #474, opened 2018-09-21
add set config methods for Context and ContextBuilder

Giesch:
In response to #473


icefoxen:
Wow, that was quick, thank you!

The `Context::set_config` method shouldn't exist though, the conf on the Context is meant to be the one it was created with.

icefoxen:
Awesome, thank you!

PR #479, opened 2018-09-30
added new example of tic-tac-toe

sn99:
I have added a new example of a simple tic-tac-toe, nothing too fancy


icefoxen:
I appreciate the effort, but this example doesn't really use any features that the existing examples (`astroblasto` and `snake`, mainly) do not.  There's also a few bugs, such as not detecting wins using diagonals.

Issue #480, opened 2018-10-02
Modularize Context


icefoxen:
I swore there was an issue discussing this but I can't find it, so I'm making it now.

Basically, it would be nice to be able to initialize only part of a Context... for example, having graphics with no audio or such.  We're already partially set up for that by having sub-`Context` objects.  The trick is doing it without sticking `Option` and `unwrap()` everywhere.

I think it would be better to make each sub-context a trait object, and then we can, for example, have an `AudioContext` that uses `rodio` to play sounds, and a `NullAudioContext` that has all the same methods but just does nothing with them, or has any calls to them panic at runtime.  This way we don't have to check to make sure the context is valid Every.  Single.  Time.  We.  Touch.  It.

That said, trait objects have their own limitations, soooo.  This might be harder than it looks, and it doesn't look easy.

onelson:
Just wanted to tag #431 here since this sort of refactor would greatly improve the achievable isolation and make testing much more convenient.

maxdeviant:
@icefoxen I'd be interested in looking into this (with some mentoring/guidance from someone).

icefoxen:
@maxdeviant Awesome! The easy place to start, take a look at #487 and see how that works, then try to do something similar with another subsystem. Probably the next easiest to start with would be the audio context.

maxdeviant:
@icefoxen As part of this modularization, I was thinking a bit about what the pros/cons of genericizing the context traits enough so that they aren't tied to a particular crate. Cause right now even though the context is more modular, it's still tied to a particular implementation (e.g., `gilrs` for gamepad input and `rodio` for audio.

I don't know if it's feasible or aligned with the goals of ggez to do that, but I can see some definite benefits. It would make it easier for people to plug-n-play with different libraries and make it easier to stub things out for unit testing.

Just some food for thought :wink:

icefoxen:
It is squarely on the list of "it would be cool but I don't actually care enough to do it right now". Might happen for the sake of unit testing, but...

PSteinhaus:
A step into a modularization of the context has now been done in #1019.
We undid the trait object approach of the gamepad context, but we're already discussing bringing it back and extending it to the other subcontexts.

Vixeliz:
#1203 doesn't completely fix this it needs more changes but its certainly an additional step in allowing a more modular context to be made.

PR #481, opened 2018-10-02
Get examples to compile and run on macOS.

seivan:
By making `game_pad` optional.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -36,7 +36,7 @@ pub struct Context {
     /// Mouse context
     pub mouse_context: mouse::MouseContext,
     /// Gamepad context
-    pub gamepad_context: gamepad::GamepadContext,
+    pub gamepad_context: Option<gamepad::GamepadContext>,
 
     /// The Conf object the Context was created with
     pub conf: conf::Conf,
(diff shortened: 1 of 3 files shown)


Ratysz:
I think this would make more sense as a feature gate, if anything.

icefoxen:
I don't disagree but I *really* don't want to load ggez down with feature gates.  They're convenient for a programmer but really inconvenient for an end-user who just has a fixed binary.

How I *want* to do this is make trait objects for each `Context` type and then be able to have, for example, `impl GamepadContext for GilrsContext` and `impl GamepadContext for NullGamepadContext` and just have them able to be chosen/switched at runtime, without needing `Option`'s everywhere.  I think there's a discussion for this somewhere in the issues but I can't find it right now.  It also might not be feasible at all, unfortunately, but I want to try.

seivan:
I am mostly just interesting to get it to run without crashing - any approach to this works for me - but as of now the dev branch doesn't run on macOS.

Ratysz:
Hence "if anything" - I saw #480 and it does seem like the way to go; hopefully, it is indeed feasible.

Could be worthwhile, however, to use a temporary solution (be it `Option`, feature, or plain removal of the dependency) in case of this issue holding up next release.

icefoxen:
Ah, #480, that was it.  Yeah, this will definitely get merged in one way or another soon; I don't intend for this to hold up at least a pre-release, and `GamepadContext` is a good test case to try different strategies.

icefoxen:
Superceded by #487

PR #484, opened 2018-10-07
Make all doctests compilable

chaosteil:
Just to make sure that the documentation stays honest.

### DIFF
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -53,8 +53,9 @@ impl From<FullscreenType> for SdlFullscreenType {
 ///
 /// Defaults:
 ///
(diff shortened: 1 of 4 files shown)


icefoxen:
Excellent, thank you so much!

If you want to be ambitious, we could use the same treatment to the devel branch.  That might take a bit more updating though.

chaosteil:
@icefoxen Sure, I can do that! However, I did look at the branch and didn't see any additional ignored doctests? Point me to the right one and I'll get to it.

icefoxen:
I guess if there's no additional ones we can just merge the existing changes from master then.  Thanks.

PR #485, opened 2018-10-13
Adds a new feature sdl2-bundled.

gigaroby:
This enables the sdl2 bundled feature which allows cargo to
automatically download and build sdl2 from source, forgoing the need to
install native dependencies.


gigaroby:
Rebasing on devel.

PR #496, opened 2018-10-20
[WIP] Implement errors using `failure` crate

### LINKED ISSUE #421: Experiment with `failure` crate.
Might make the error module a little less annoying.

maxdeviant:
# :construction: WORK IN PROGRESS :construction: 

This PR makes use of the [`failure`](https://rust-lang-nursery.github.io/failure/intro.html) crate for handling errors.

Closes #421.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,8 @@ multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
 bitflags = "1.0"
 zip = { version = "0.4", default-features = false }
 app_dirs2 = "2"
+failure = "0.1"
+failure_derive = "0.1"
 gfx = "0.17"
 gfx_device_gl = "0.15"
 gfx_glyph = "0.12"
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,13 +2,14 @@
 
 use std;
 use std::error::Error;
-use std::fmt;
+use std::fmt::{self, Display};
 
 use gfx;
 use glutin;
 use winit;
 
 use app_dirs2::AppDirsError;
+use failure::{self, Backtrace, Fail};
 use gilrs;
 use image;
 use lyon;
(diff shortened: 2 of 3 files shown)


icefoxen:
Now that I've had some time to actually look at this...

It's very tempting; the cleanup to the (admittedly quite crappy) `GameError` type in particular is very nice.  I was kinda skeptical of whether this was worth an extra dependency and such, but frankly making error handling easier to write means it's easier to improve, so.

I MIGHT hold off on merging this until 0.5 is released, but I like it so far.  Anyone else have thoughts?

acheronfail:
FWIW I've had some great experiences using the `failure` crate and would highly recommend it. :+1: 

Although, would it be better to just convert `GameError` to an `enum`, so we can automatically derive the `Fail` trait? That way we don't even have to maintain its `impl` as well.
That would also mean there's no need to write the `impl Display for GameError` either, as that would be automatic. Then we'd have almost no boilerplate for `GameError` at all.

acheronfail:
Have a look at [`toml-query`](https://crates.io/crates/toml-query) which uses the `failure` crate and you can see how little boilerplate you actually need: https://github.com/matthiasbeyer/toml-query/blob/master/src/error.rs

maxdeviant:
@joetsoi Does using `failure` force it on everyone? As far as I know, downstream consumers can still deal with errors in the same way.

That being said, `failure` is going to (eventually) be the default error system, so people might as well get used to it 😄 

> failure is designed to make it easier to manage errors in Rust. It is intended to replace error management based on std::error::Error with a new system based on lessons learned over the past several years, including those learned from experience with quick-error and error-chain.
>
> &mdash; [source](https://github.com/rust-lang-nursery/failure#failure---a-new-error-management-story)

ozkriff:
Some critic of failure usage in libraries: https://www.reddit.com/r/rust/comments/993dic/error_management

icefoxen:
@joetsoi That's odd, what is making `GameError` be `!Sync`?

joetsoi:
see https://github.com/ggez/ggez/issues/566 i've opened a pr to glutin, so hopefully that'll get merged and fixed in next release.

icefoxen:
Actually thinking about this a little more...  With this setup, the main advantage we get is nice `display` impl's.  The main hair in ggez's error handling is the plethora of rather ugly `From` impl's that just stringify whatever error we get.  What we would honestly kind of *like* is for ggez's `GameError` type to have, like, a useful description string of some kind (for the cases where for example lyon gives a `TessellationError` and we say "did you give it a vertex with an inf or NaN in it?"), and just one actual error type instead of a big wobbly enum.  The enum breaks any time we update a dependency, and is kinda a pain to update, which is the main reason everything just gets stringified.  `failure`'s `Error` type (basically a `Box<Fail>` with a backtrace attached to it) seems like exactly what I want.  

Going to look over the criticism linked here a little more and see what it looks like.

icefoxen:
Criticism seems to be:

 * Allocations for `failure::Error` type can be heavyweight
 * Creating backtraces can be slow and require lots of added complexity (more dependencies, more compile time, etc)

It appears that one way or another we can control both of these problems.  We can derive `Fail` on our own error types, or implement it ourselves, which lets us decide how much work gets done on an error or `?` -- see https://rust-lang-nursery.github.io/failure/fail.html and the following chapter. https://rust-lang-nursery.github.io/failure/custom-fail.html#caveats-on-this-pattern mentions some problems with this technique, but they're mostly problems our current error handling already has anyway so it's not a step backwards.

We can also control how much backtrace stuff actually gets done, it appears; you can disable the `backtrace` feature in `failure`, which makes it produce null code that gets optimized out: https://docs.rs/failure/0.1.5/src/failure/backtrace/mod.rs.html#102-104 We can also just make our own error type that implements `Fail` but has no backtrace.

Then there's the "Error and ErrorKind" method, https://boats.gitlab.io/failure/error-errorkind.html which this PR implements, which is... exactly what I asked for: "What we would honestly kind of _like_ is for ggez's `GameError` type to have, like, a useful description string of some kind (for the cases where for example lyon gives a `TessellationError` and we say "did you give it a vertex with an inf or NaN in it?"), and just one actual error type instead of a big wobbly enum".  This is a little heavyweight; the `failure::Context` type basically contains a `failure::Error`, so, a `Box<dyn Fail>` plus an (optional if you remove the `backtrace` feature flag) backtrace.  But ggez basically never has an `Err` be the "happy path" as described in https://www.reddit.com/r/rust/comments/7te8si/personal_experience_can_cause_hard_to_notice/ .  An error is an error and it's either due to improper function inputs or something wrong that's outside the program itself -- file missing, graphics driver threw a hissy-fit, etc.  So, needing to allocate on creating an error is not a huge concern.  ggez will never be `no_std` anyway.  Still, I'd kinda like to avoid using `failure::Error` directly for the sake of simplicity, and for making sure our users don't need to use `failure`.

Whew!  So.  Okay.  My current thoughts: We should either take this PR as-is (`Error`+`ErrorKind` pattern) or take the slightly simpler path of deriving `Fail` on our existing `GameError` type (what the `toml-query` crate linked above does) but not using `failure::Error`.  Either way we can safely disable backtraces -- or at least, have a feature flag for them that I'd probably make disabled by default just for the sake of simplicity.  Either way it will be an incremental improvement but not rock our world.

So the last thing to consider is, do we want to force our users to use `failure` as well?  And I don't think that using `failure` does that, as long as we avoid `failure::Error`.  Basically the only things you can sanely do with a ggez error is print it out with `Debug` or `Display` anyway, so if we return `GameError` and the user doesn't want to import `failure::Fail` then they just don't have to use the traits or call the methods on it.  Implement `std::error::Error` on our `GameError` type and life is grand.

maxdeviant:
Closing this out as #421 is also now closed.

Issue #505, opened 2018-10-27
Make ggez build cleanly on Rust 2018


icefoxen:
Now that the 2018 edition's package rework is in stable, we should use it!  There's an experimental fork where I got such things working, but the implementation has changed a little, and that fork is stale anyway.  We should make those changes straight to `devel`.

dannyfritz:
Should this be closed?

icefoxen:
Yeah, probably!  Thanks.

PR #506, opened 2018-10-29
Devel updated to work on 2018

Lokathor:
Also I fixed _part of_ the translation problem. I suspect not all of it.


Lokathor:
Apparently I should link the related issue in the PR itself, so here we go: 

https://github.com/ggez/ggez/issues/505

PR #507, opened 2018-10-30
Use https for all shield images

dannyfritz:
Use https for all shield images to avoid security warning on docs.rs


icefoxen:
Awesome, thank you!

PR #508, opened 2018-10-30
📇 conf.toml example

dannyfritz:
The doc markdown link I did does not work. I'm not sure how to link to a src file in examples.


dannyfritz:
screen shot 2018-10-30 at 9 08 31 am

PR #511, opened 2018-11-11
chore: update to `glutin` v0.19

### LINKED ISSUE #478: ggez starts up with black screen on Mac Mojave
- On a Mac with Mojave installed
- `$ cd path/to/ggez/examples`
- `$ cargo run --example super_simple`

Expected: ggez window starts up with a blue screen and a white circle racing across it
Actual: ggez window remains black. If you move the window by dragging it with your mouse, it starts rendering.

---

Other examples seem to exhibit the same behavior. I know this wasn't happening earlier this week, but I *just* updated my OS and started seeing this new issue. I'm a little paranoid that it may be some local issue - so to start, can anyone else verify if they're seeing it too?

acheronfail:
Should fix #478 (for the `devel` branch)

This PR contains the initial changes required to update the `glutin` and `winit` dependencies.

Note that this won't compile because we need to wait on some dependencies:

* `gfx_window_glutin`: https://github.com/gfx-rs/gfx/pull/2468

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ gfx_glyph = "0.12"
 gfx_window_glutin = "0.27"
 glutin = "0.19"
 winit = { version = "0.18", features = ["icon_loading"] }
-image = {version = "0.19", default-features = false, features = ["gif_codec", "jpeg", "ico", "png_codec", "pnm",
+image = {version = "0.20.1", default-features = false, features = ["gif_codec", "jpeg", "ico", "png_codec", "pnm",
 "tga", "tiff", "webp", "bmp", "dxt", ] }
 rodio = "0.8"
 serde = "1"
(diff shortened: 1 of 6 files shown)


acheronfail:
@icefoxen Okay I've got those upstream dependencies merged and published.

Although, I'm not sure why these builds are failing at the moment? The errors are saying that `GameError` doesn't support the `From` trait for `image:ImageError`, but it clearly does: https://github.com/ggez/ggez/blob/7ca884c45f4ec3845e4bbf2b648d67a6a97497b8/src/error.rs#L126-L131

TBH, I don't see why it's not working... 
Other than that, everything else is green and good to go. 🙂

icefoxen:
See https://github.com/tomaka/winit/blob/master/Cargo.toml#L23

vs https://github.com/ggez/ggez/blob/devel/Cargo.toml#L40

Just need to bump the `image` version ggez uses to match `winit`.

acheronfail:
Ahh it's always something obvious, right? 🤦‍♂️ 

In any case, that now builds! 🎉 

FWIW I always forget the same thing when it comes to versioning with `cargo`. 😆

acheronfail:
I see that the `devel` branch itself it failing with that error anyway.
In that case, I think we're all good here.

PR #514, opened 2018-11-21
Updated dev-dependency rand to current version 0.6.0

TaylorDennisLee:
The latest ```rand``` version makes a change to the function signature of ```gen_range```.


icefoxen:
Awest, thank you!

PR #522, opened 2018-12-20
Update README.md with a mention of the 'hello ggez' guide

m1ksu:
Added a suggestion to read the `hello ggez` guide for new users. It's better to let the to-be users immediately know that there exists a 'getting started' tutorial such as this. It may be hard to find otherwise, as it requires manually looking at the `docs/` folder to find it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,10 @@ or `Conf` object, and then call `event::run()` with
 the `Context` and an instance of your `EventHandler` to run your game's
 main loop.
 
+## Getting started
+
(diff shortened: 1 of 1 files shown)


icefoxen:
I'm fine with this, but it also involves having the `hello ggez` guide finished...

icefoxen:
Actually, it's more finished than I thought; I thought this was a reference to... some other incomplete docs that I now can't find.  Derp.  Thank you!

Issue #98, opened 2017-07-10
Playing audio should reuse the decoder


aochagavia:
This is already documented as a comment in the implementation of `Source::play`, but I think an issue is useful so we don't forget it.

aochagavia:
I think the easiest thing would be to somehow clone the decoder. However, `rodio` doesn't implement `Clone` for `Decoder`, since it wraps types defined in other libraries that don't implement `Clone` either.

icefoxen:
Yeah, as I recall this is the main reason we don't reuse the decoder.  Definitely worth looking into.  Is it possible to implement `Clone` on the underlying types?  (Possibly not, since they may be reading from file handles or other non-easily-cloneable types, but it's worth checking.)

aochagavia:
I checked `lewton` and it seemed possible. The only problem is that you need to implement `Clone` manually for some structs because they contain large arrays (and `Clone` is implemented only up to a given length). I gave up after a while because it looks horrible and took more time than I had at the time, but it seems doable.

icefoxen:
I looked at this more and it's probably not worth the trouble... or at least, not an urgent issue.  We'd need to implement `Clone` for the rodio `Decoder`, which means going allll the way down the rabbit-hole and implementing it on the underlying `hound`, `lewton` and `claxon` types.  It's certainly possible, it just means adding `impl Clone for WhateverDecoder<Reader> where Reader: Clone` to a bunch of different crates.

Upon inspection, the actual detection of the source data type is not terribly expensive; it basically amounts to checking the magic number at the beginning of the data stream.  While being able to skip this would be convenient, all the more complicated stuff you have to do anyway to clone a decoder (resetting the stream's cursor, clearing out any other state you have, etc) probably is the majority of the work and time spent.  So I doubt clone'ing a decoder would be significantly cheaper than just creating a new one.

I might be wrong though!  I won't say no if this functionality becomes available, but it seems like the potential gains are pretty small.

icefoxen:
The gains are quite tiny, so not bothering.

PR #534, opened 2019-01-05
cache source instead of calling Decoder::new() every play

slmjkdbtl:


slmjkdbtl:
#98

slmjkdbtl:
I'm assuming creating `SamplesBuffer` is cheaper than `Decoder`

slmjkdbtl:
closed because I found that `source.convert_samples().collect::<Vec<f32>>()` is taking 10 seconds even for a small file

icefoxen:
It might be *significantly* faster in release mode than debug mode. But yes.

slmjkdbtl:
@icefoxen what about storing a
```rust
pub struct Source {
    buffer: rodio::source::Buffered,
    // ...
}
impl Source {
    pub fn from_data(context: &mut Context, data: SoundData) -> GameResult<Self> {
        // ...
        let src = Decoder::new(cursor).unwrap();
        Ok(Source {
            buffer: src.buffered(),
            // ...
        })
    }
    // ...
}
```
and play by 
```rust
self.sink.append(self.buffer.clone().convert_samples());
```

I haven't done any benchmarks though

PR #538, opened 2019-01-08
Added missing step in instructions

fedor-rusak:
These instructions require latest version of ggez plus event::run step is required to actually start loop.


icefoxen:
Thank you!  :heart:

PR #542, opened 2019-01-11
Change `use` statement so it compiles

EvilRobotOverlord:
I couldn't compile this before because rust 2018 messed this up.

### DIFF
--- a/src/graphics/image.rs
+++ b/src/graphics/image.rs
@@ -2,7 +2,7 @@ use std::io::Read;
 use std::path;
 
 use gfx;
-use image;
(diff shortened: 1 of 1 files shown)


rukai:
Are you using the latest compiler? run `rustup update`.

rukai:
Woah, this is a change in nightly https://github.com/rust-lang/rust/issues/56316

@icefoxen Looks like we actually need this as rust is breaking backwards compatibility to fix a bug.

icefoxen:
Thank you both for helping hunt this down!  :heart:

PR #544, opened 2019-01-12
add const colors

sylba2050:
Currently, only black and white are defined in ggez.
I feel that these alone are not enough.
Therefore, we added some new colors this time.
Please check if you like.


icefoxen:
I made a crate that did a more general purpose version of the same thing.  I could have sworn I published it, but apparently I never got around to it, so here it is now: https://crates.io/crates/chromatica

Not sure if that does exactly what you want though.  Might be simpler to just integrate it into ggez proper...

PR #563, opened 2019-01-22
Fix example command

dnikk:
Run this command `cargo run --example astroblasto`.
Then, the following error message is displayed.

```
error: no example target named `astroblasto`

Did you mean `05_astroblasto`?
```

So I fixed this command.


icefoxen:
Thanks, but you're gonna need to start from a fresher version of `devel` I'm afraid.  XD

Don't sweat it, I'll fix it.

PR #567, opened 2019-01-27
Drawable::dimensions()

### LINKED ISSUE #557: Add a 'dimensions' method to the 'Drawable' trait
I need my code to be generic over `Text` and `Image`. In GGEZ v0.4 I was using `.into_inner()` method to convert `Text` to `Image` and just worked with images afterward, but there's no easy way to do this with the new font rendering.

I can use `Box<Drawable>` (though, see https://github.com/ggez/ggez/issues/556) to draw the items, but I also need to know their size to use them in [zgui](https://github.com/ozkriff/zemeroth/tree/dbcafeade/ggwp-zgui)'s widgets (for layout restructuring and mouse click handling). The `Drawable` trait doesn't provide any method like `dimensions` atm.

ozkriff:
This PR:

- Adds an `approx::RelativeEq` impl for `Rect`;
- Adds a `fn rotate(&mut self, rotation: f32)` method to `Rect;
- Adds a `fn combine_with(self, other: Rect) -> Rect` method to `Rect`;
- Adds a `fn dimensions(&self, _: &mut Context) -> Option<Rect>` method to the `Drawable` trait;

Notes:
- I don't like that `dimensions` method requires `Context` - it's required only for `Text`s implementation. It can be solved by storing a `dimensions: Rect` field in `Text` during `Text`'s creation, but it increases the costs of this operation which may be not desirable.
- I really hope that I haven't messed up the math too much. The basic tests seem to be ok, but it needs some staring review.
- Also, I'm not sure that `combine_with` is a descriptive name but can't think out anything better.
- It'd be cool to remove the code duplication between `bbox_for_vertices` and `rotate`, but I don't see how it can be done since they work with different point types.

Closes #557 

_([Here's a branch of Zemeroth with a draft port to GGEZ 0.5.alpha.0 + this PR](https://github.com/ozkriff/zemeroth/tree/wip_i409_ggez_0_5_alpha_0))_

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ nalgebra = {version = "0.16", features = ["mint"] }
 # Has to be the same version of mint that nalgebra uses here.
 mint = "0.5"
 gilrs = "0.6"
+approx = "0.3"
 
 [dev-dependencies]
 chrono = "0.4"
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -126,6 +126,9 @@ impl Drawable for Canvas {
         self.image.draw(ctx, flipped_param)?;
         Ok(())
     }
+    fn dimensions(&self, _: &mut Context) -> Option<Rect> {
+        Some(self.image.dimensions())
+    }
     fn set_blend_mode(&mut self, mode: Option<BlendMode>) {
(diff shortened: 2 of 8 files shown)


ozkriff:
r?

icefoxen:
r+!

PR #578, opened 2019-02-09
Fix example 01

iolivia:
This example seems out of date with the latest official cargo release.

### DIFF
--- a/examples/01_super_simple.rs
+++ b/examples/01_super_simple.rs
@@ -19,32 +19,33 @@ impl MainState {
 }
 
 impl event::EventHandler for MainState {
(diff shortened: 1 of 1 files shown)


ozkriff:
Examples in the master branch work with https://crates.io/crates/ggez/0.5.0-rc.0 version of ggez.

Older examples for 0.4 version (latest stable release) can be found here: https://github.com/ggez/ggez/blob/0.4/examples/super_simple.rs

iolivia:
thanks! I figured I must be working off the wrong version but wasn't sure what master was supposed to point to.

Issue #580, opened 2019-02-12
drawparam fields should be public


icefoxen:
Given someone might want to impl Drawable themselves someday.

PR #581, opened 2019-02-13
Make all the properties on DrawParam public

ocboogie:
It doesn't seem to be necessary to make all the properties on `DrawParam` private outside of the crate. It makes it difficult to implement `Drawable`s when you don't have access to the properties.

Also if [ggez-goodies](https://github.com/ggez/ggez-goodies) updated to v5 it would need this, [here](https://github.com/ggez/ggez-goodies/blob/master/src/camera.rs#L107).


icefoxen:
Thank you!  :heart: 

Resolves #580 .

Issue #573, opened 2019-02-03
setVolume method for audioSource works only after play()


fedor-rusak:
Environment: Win10

So I made repeatable sound for background noise and added sound of movement that was too loud for my taste.

I decided to setVolume for step sound  during initialization (reading sound file)... and it changed nothing.

But when I added setVolume right AFTER play() it made the required change.

Is it intentional?

icefoxen:
Note to self: Should be fixed in #584.  Verify, unit test, then close.

fedor-rusak:
Updated my project to use latest version from repo. Stumbled upon requirement to use trait SoundSource.

Added it and now set_volume works as intended before actual sound playing!

Should I close the issue?

icefoxen:
Glad it works! I'm leaving this open until we have better tests for now, to make sure things don't backslide later.

sampullman:
There are a couple unit tests, eventually there might be an overhaul if CI can be improved to handle integration testing. Until then I think it makes sense to close older issues like this.

PR #584, opened 2019-02-15
Fixes problem with not being able to set the volume before playing a sound

MCluck90:
Fixes issue #573 

The problem lies in how the `stop` function works. `stop` works by dropping the current sink and replacing it with a new one. However, it doesn't carry over the volume setting from the old sink. This matters because `play` calls `stop` to restart a sound, if need be. The fix is to hold on to the volume for the old sink before dropping it then re-applying that volume setting.

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -473,6 +473,10 @@ impl SpatialSource {
         // fine because the `RodioAudioContext` uses the default device too,
         // but it may cause problems in the future if devices become
         // customizable.
+
+        // We also need to carry over information from the previous sink.
+        let volume = self.volume();
+
         let device = rodio::default_output_device().unwrap();
         self.sink = rodio::SpatialSink::new(
             &device,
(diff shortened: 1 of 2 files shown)


icefoxen:
You

Are a gorram *hero*

Can I convince you to add a unit test? :D

MCluck90:
Lol thanks! 

Yeah, no worries. I'll try to add one either this evening or tomorrow

MCluck90:
All right, sent that up. Good thing you asked for the tests because I realized that I forgot to include the fix for `Source` and not just `SpatialSource`. Should be good now!

icefoxen:
Cooooooould I convince you to refactor Source and SpatialSource to mostly use a common subtype? :D

You really don't have to. But it's on the to-do list at some point.

MCluck90:
That is a good idea. Sure, I'll take care of it! 👍🏻

icefoxen:
Definitely a starting point.  Awesome!

MCluck90:
Sweet! I'll keep plugging away on it and try to get it to a happy point

PR #589, opened 2019-02-18
Fixed: fill() is invalid

SMUsamaShah:
DrawMode::fill() to DrawMode::Fill


ozkriff:
See https://github.com/ggez/ggez/pull/578

PR #591, opened 2019-02-21
Merged SpatialSource in to Source to create a unified audio API

MCluck90:
Let me know what you think. This is a continuation of the previous work on `Source` and `SpatialSource`. Basically, `SpatialSource` is gone but there is a way of creating a spatial sound via `new_spatial` or `from_spatial_data`.

There's one thing that I think is a little weird though. If you call any spatial source specific methods (`set_position` for example) and the user built a normal source, it doesn't give any indication that this is essentially a noop. Maybe that isn't an issue, I don't know. What do you think?

### DIFF
--- a/examples/05_astroblasto.rs
+++ b/examples/05_astroblasto.rs
@@ -243,8 +243,8 @@ struct Assets {
     rock_image: graphics::Image,
     font: graphics::Font,
     // Todo: add a music track to show non-spatial audio?
-    shot_sound: audio::SpatialSource,
-    hit_sound: audio::SpatialSource,
+    shot_sound: audio::Source,
+    hit_sound: audio::Source,
 }
 
 impl Assets {
@@ -254,8 +254,8 @@ impl Assets {
         let rock_image = graphics::Image::new(ctx, "/rock.png")?;
         let font = graphics::Font::new(ctx, "/DejaVuSerif.ttf")?;
 
-        let mut shot_sound = audio::SpatialSource::new(ctx, "/pew.ogg")?;
-        let mut hit_sound = audio::SpatialSource::new(ctx, "/boom.ogg")?;
+        let mut shot_sound = audio::Source::new_spatial(ctx, "/pew.ogg")?;
+        let mut hit_sound = audio::Source::new_spatial(ctx, "/boom.ogg")?;
 
         shot_sound.set_ears([-1.0, 0.0, 0.0], [1.0, 0.0, 0.0]);
         hit_sound.set_ears([-1.0, 0.0, 0.0], [1.0, 0.0, 0.0]);
--- a/examples/05_astroblasto.rs
+++ b/examples/05_astroblasto.rs
@@ -7,6 +7,7 @@ extern crate nalgebra;
 extern crate rand;
 
 use ggez::audio;
+use ggez::audio::SoundSource;
 use ggez::conf;
 use ggez::event::{self, EventHandler, KeyCode, KeyMods};
 use ggez::graphics;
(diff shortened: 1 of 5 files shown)


icefoxen:
I agree it's a little weird.  There's some other cases where this sorta happens (`NullAudioContext` and such mainly) and I am *very undecided* about whether the correct answer is to panic or make the call a noop.

Let's think about this...

What we really have is a `Source` and a `SpatialSource` where `Source` has a strict subset of the functionality of `SpatialSource`.  So we *could* implement `Source` using `SpatialSource` just with having all the ear positions at `0,0` and immobile.  It would have an unnecessary performance cost and be a bit hacky, but it would be the simplest and most straightforward way to do it:

```rust
enum Source(SpatialSource);
impl Source {
    pub fn play(&mut self) { self.0.play() }
    pub fn stop(&mut self) { self.0.stop() }
    // ...and so on...
}
```

The repetition is annoying but not the worst thing ever.

The annoyances there are a) `Source` and `SpatialSource` are still never interchangable while having mostly the same methods, and b) repetition.  This seems like prime time for trait objects:

```rust
trait SourceTrait {
    fn play(&mut self);
    fn stop(&mut self);
    ....
}

trait SpatialSourceTrait: SourceTrait {
     fn set_ears(&mut self, ...);
}

impl SourceTrait for Source ...
impl SourceTrait for SpatialSource ...
impl SpatialSourceTrait for SpatialSource ...
```

Then people can mix together `Box<dyn Source>` and `Box<dyn SpatialSource>` however they want. The problem is that trait objects can't have any methods with generic parameters, and we use those a fair amount, though mostly for paths and `mint` point types.  Removing those would be annoying but not the worst thing ever... so about as inconvenient as the above solution.

However, I THINK we can actually do exactly that and just not bother with making the traits object-safe.  It would make things *slightly* less convenient because you couldn't have a function that returns either a `Source` or a `SpatialSource`, but frankly, if you want to do that you should probably just use `SpatialSource`'s everywhere anyway.  We don't even actually need `SpatialSourceTrait` to inherit from `SourceTrait`.  Doing things that way would *still* cause some annoying code duplication... but frankly it sounds like the only solutions are annoying ones, and if we make all the API be part of traits with a clear relationship between them then we are free to tinker around with the actual implementation as much as we want without users having to alter their code much.

icefoxen:
Tried to sketch my thoughts in #592.  Not sure if it works better than what you have here or not.  Any opinions?

MCluck90:
I like your approach better. It seems a lot more flexible and would result in fewer `match` statements that have to line up to keep everything in check. I've pulled it down and started playing with it myself. There's still the issue of getting `Source` to use `rodio::Sink` and `SpatialSource` to use `rodio::SpatialSink` but it's definitely getting closer.

Like you said, worse case we can use `rodio::SpatialSink` everywhere and lock its position for `Source`. I'll try to play with it some more between work today and see if I think of anything.

icefoxen:
Sure, if you want. Like I said in the commit message, I think the best approach would be to make what is now `SourceImpl` omit the `sink`, make `Source` and `SpatialSource` separate objects that contain a `sink`, a `SourceImpl` or whatever we call it (source state?), and refactor as necessary.

MCluck90:
Ohhhh right, yeah, I get what you're saying now. That's what I get for bouncing between this and my work too quickly. Yes, I think that's a good idea.

icefoxen:
I'm doing the same, no worries.  ;-)

MCluck90:
I took your idea and pushed it to the point of building haha. This looks a lot better in my opinion. There's a little bit of duplication but it's only forwarding calls to `SourceState` or for sections where we need to differentiate between `rodio::Sink` and `rodio::SpatialSink`.

icefoxen:
Looks very nice!  Is this ready to be merged, or is there more work you want to do on it?

MCluck90:
I think it's ready to go!

Issue #569, opened 2019-01-29
(0.5 feedback) Text::dimensions height behavior changed


ozkriff:
Was that intentional? What is the reasonong? It breaks Zemeroth's UI code, because I'm scaling labels and buttons to 1/12th of the screen's height.

GGEZ v0.4 vs v0.5:

![image](https://user-images.githubusercontent.com/662976/51888356-e16c6a00-23a7-11e9-9ccc-e6a799d484a8.png) ![image](https://user-images.githubusercontent.com/662976/51888394-fd700b80-23a7-11e9-9427-e2193e9b724b.png)

If it was intentional, is there another method to get "old-fashined" bounding box of the text?

icefoxen:
My first guess is that it is due to winit's hidpi functionality lying about how big the display is.  If the result of `graphics::os_hidpi_factor()` is >1.0 then that is probably it.

In general this isn't intentional, if we need to fix the text bounding box code to be accurate then we should do that.

ozkriff:
No, that's definitely not a HDPI problem. Here's a screenshot with both versions overlapping.

![image](https://user-images.githubusercontent.com/662976/52166703-b2e0ed00-2721-11e9-9093-3bbbeb1ea54c.png)

Strings with descender parts in them ("strength:", etc) are of the same size and strings like "moves:" clearly has less pixels in them (and thus they looks bigger when scaled to the same screen height).

> In general this isn't intentional, if we need to fix the text bounding box code to be accurate then we should do that.

Ok, I'll try to look into the code.

icefoxen:
Wow. Yeah that looks awful. Good call on noticing the descender and ascender parts are what make it different. I expect that we're just using the wrong font metrics for sizing somewhere. I will work on this as well, if we can't find it in a few days we might just want to ask for help from alexheretic.

ozkriff:
So, I've experimented a little bit with this and the solution to bring the old behaviour for oneliners is:

```diff
     fn calculate_dimensions(&self, context: &mut Context) -> (u32, u32) {
         let mut max_width = 0;
         let mut max_height = 0;
         {
             let varied_section = self.generate_varied_section(Point2::new(0.0, 0.0), None);
+            for section in &varied_section.text {
+                let h = section.scale.y.ceil() as _;
+                if h > max_height {
+                    max_height = h;
+                }
+            }
             use glyph_brush::GlyphCruncher;
             let glyphs = context.gfx_context.glyph_brush.glyphs(varied_section);
```

[...]

PR #593, opened 2019-02-21
Fix Text::dimensions height

ozkriff:
See #569


ozkriff:
@icefoxen r? :)

icefoxen:
Yeah, I'm good.  I just want to try to make unit tests for text metrics that actually work, to go with this, and want to leave this open so I don't forget.  Hopefully I'll have time to do that tomorrow.

PR #596, opened 2019-03-01
Sound returns an error for invalid audio and added tests

MCluck90:
I noticed that TODO about it being awkward that we have to play a sound to see if it's valid. I added a `can_play` method to `SoundData` that will tell us if it's possible to play that sound. Using that, the constructors for `Source` and `SpatialSource` will return an error if the user attempts to build a sound with invalid data.

I also added a number of unit tests to the audio module to test these cases as well as verify what we get back from `paused` and `playing` based on the state of the sound. All of the unit tests cover both `Source` and `SpatialSource`


icefoxen:
My hero.  :heart:

PR #598, opened 2019-03-05
Replace app_dirs2 with directories

### LINKED ISSUE #530: Investigate using `dirs` for directory locations rather than `app_dirs2`
It may or may not be better for our use case, but it's certainly gotten more use beaten on it.  The crate readme actually has a pretty good comparison between different options though, investigate.

tverghis:
This patch replaces the use of the `app_dirs2` crate with the more
heavily-used `directories` crate.

Fixes #530 (original suggestion was to use the `dirs` crate, but my research suggests that `directories` seems more analogous).

Also, this is my first PR *ever*, so please let me know if there is something I can improve!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ c_dependencies = ["bzip2", "mp3"]
 [dependencies]
 bitflags = "1.0"
 zip = { version = "0.5", default-features = false }
-app_dirs2 = "2"
+directories = "1.0.2"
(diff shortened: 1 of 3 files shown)


icefoxen:
Upon inspection, it looks like `dirs` and `directories` are made by the same person, and both maintained -- like you say, `directories` seems a better choice.

Excellent work, thank you!

PR #607, opened 2019-03-18
Add graphics::present() to starting boilerplate in GenerativeArt.md

Reificator:
Since graphics::present() is required at the end of draw() in order to see the results of previous draw operations, I've gone ahead and added it to the starting boilerplate for the GenerativeArt tutorial.

It might be useful to call out the method specifically in the tutorial text, but I'll leave that to someone more familiar with the library.


icefoxen:
Oof.  Nice catch, thank you.

PR #620, opened 2019-04-11
Fix input example

phlmn:
Seems like you missed something in 815100623af53e2c9174ddf56a274562992b0a92. 🙂


icefoxen:
Thank youuuuuuu

Issue #469, opened 2018-08-30
Add ability to integrate a .zip file into a binary


icefoxen:
@svenstaro mentioned this and it was a cool idea, being able to make a build.rs that zips up your resources directory and then include it into your .exe with `include_bytes!`.  It should take a quite minor extension to the `vfs` module to do, and it would be cool and useful (if not always the most efficient) to be able to bundle all your assets into a single .exe.

We've pondered things like this before but the idea of just using a zip file or similar archive instead of having to write one's own in-memory vfs is new.

andreivasiliu:
Trying to get my thoughts together here, ignore me (PR later).

`ZipFS::new()` only takes a filename, I'll likely refactor it to use a `ZipFS::from_file<F: Stuff>(file: F)`, and also add a `ZipFS::from_bytes()` that gives `from_file(std::io::Cursor::new(bytes))`. Maybe I should rename `::new()` to `::from_path()` but I don't know what that would break.

How does a user pass the bytes to ggez? I'm thinking:
```
ContextBuilder::add_zipfile_bytes(self, Vec<u8>) -> Self
ContextBuilder::add_file_bytes(self, contents: Vec<u8>, mount_path: P) -> Self  # won't implement here

ctx_builder
    .add_zipfile_bytes(include_bytes!("resources.zip"))
```

The new resource loading order will be:
* The resources/ subdirectory
* The resources.zip file
* The root folder of the game's "save" directory
* Paths from `.add_resource_path()`
* Zipfiles from `add_zipfile_bytes()` ***New!***

andreivasiliu:
Actually, now that I look at it, it's kinda silly to force the user into giving bytes, when a `Read+Seek` is a universal interface that could be used for all sorts of fun stuff.

Hey @icefoxen - rather than bytes, would you be okay with allowing users to give any file instead? The user code would look like this:
```
ContextBuilder::add_zip_file<R: Read+Seek>(self, reader: R) -> Self

ctx_builder
    .add_zip_file(std::io::Cursor::new(include_bytes!("resources.zip")))
```
Which is, yeah, a lot longer, but a lot more powerful.

icefoxen:
That looks reasonable to me.  Bytes and read+seek are good at different things, but for this case I think read+seek would be a fine choice.

In terms of naming, the ggez convention is that when creating a resource such as image or file, `Foo::new()` takes a path and there's also `Foo::from_bytes()` or `Foo::from_read()` or whatever that takes something lower-level.

[...]

PR #622, opened 2019-04-21
Add ContextBuilder::add_zipfile_bytes() to load zip files from memory

andreivasiliu:
This gives the ability to embed resources into the executable using
include_bytes!(), or loading zip files from arbitrary places.

See issue #469.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
    disabling it will build ggez without unnecessary C dependencies
    (currently `bzip2` and `minimp3`). [#549](https://github.com/ggez/ggez/issues/549)
  * Added (basic) spatial sound support.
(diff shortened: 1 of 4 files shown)


andreivasiliu:
Note: If anyone has any better ideas on how to do this, I'm willing to rewrite it from scratch.

andreivasiliu:
I had the exact same feeling, and looked at `Cow` for an hour straight and
almost convinced myself it wasn't the answer. Never seen your way of
writing it before though, I'ma look at it again.

icefoxen:
Merged it and made minor changes, see here: 4bc77df

Using `Cow` appears to do exactly what we want.  Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f8347b5d0cbd228ba42f498d6f308fed

Thank you so much!

PR #623, opened 2019-04-23
Resize gl context on resize

trimental:
Wayland needs the gl context to be resized for it to be displayed properly


icefoxen:
Should be resolved in commit 0cb700d , please test and make sure it is correct.  The `resize()` and `resize_viewport()` might have to be done in the opposite order, but I don't have a functioning Wayland setup to test on.

trimental:
That works

PR #624, opened 2019-05-04
Remove second mention of event::run

red-hara:
The event::run is mentioned two times, however the second mention is used without the result handle and is not required anymore.


icefoxen:
Nice, thank you.

PR #625, opened 2019-05-05
Minor typo

AndrewJakubowicz:
In the example the canvas is drawn in the top left, instead of top right.
Changed direction in comment.


icefoxen:
Thank you, good catch!

PR #633, opened 2019-05-17
Add `filter` and `set_filter` to `Canvas`

maciejhirsz:
I want to do some pixelart graphics, so my idea was to render to a fixed size canvas, and then upscale it with nearest pixel filter. Seemed like a pretty trivial and innocent change. Tests are failing on master, but this change somehow also makes different tests fail on my machine and I haven't dug out why yet.

### DIFF
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -107,6 +107,16 @@ impl Canvas {
         &self.image
     }
 
+    /// Get the filter mode for the image.
+    pub fn filter(&self) -> FilterMode {
+        self.image.filter()
+    }
+
+    /// Set the filter mode for the canvas.
+    pub fn set_filter(&mut self, mode: FilterMode) {
+        self.image.set_filter(mode)
+    }
+
     /// Destroys the `Canvas` and returns the `Image` it contains.
     pub fn into_inner(self) -> Image {
         // TODO: This texture is created with different settings


maciejhirsz:
@icefoxen FYI, the weird case of tests failing for me is:

```
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `Ok`
  --> /tmp/rust-skeptic.8AeRT3TxIn90/test.rs:43:5
   |
42 |     graphics::present(ctx)?
   |                            - expected one of `.`, `;`, `?`, `}`, or an operator here
43 |     Ok(())
   |     ^^ unexpected token

error[E0308]: try expression alternatives have incompatible types
  --> /tmp/rust-skeptic.8AeRT3TxIn90/test.rs:42:5
   |
42 |     graphics::present(ctx)?
   |     ^^^^^^^^^^^^^^^^^^^^^^^ expected enum `std::result::Result`, found ()
   |
   = note: expected type `std::result::Result<(), ggez::error::GameError>`
              found type `()`

error: aborting due to 2 previous errors
```

This PR doesn't touch anything close to it. Even weirder is the fact that the CI is passing here. There is another error I get occasionally which is just an assert fail, and I think that one is due to me running hidpi under Linux. In either case, the errors are not caused by or affected by this PR, I thought they were but they simply are not deterministic :man_shrugging:.

Anyway, with this out of the way unless there are some objection I'd be very happy to have this PR merged, as it would allow me to do nice pixel-perfect graphics, without resorting to hacks or running off my fork :).

icefoxen:
Thank you for the PR!  I just want to test it out myself before merging it since canvas behaviors are Complicated and it's easy to break.  Looks like the unit test failures are typos in the docs being caught by rust-skeptic, not sure why the CI isn't seeing them (the rust-skeptic tests can't run in headless mode perhaps?).  Hopefully I'll get this handled this weekend, but I suck at deadlines, so all I can say is it will happen eventually.  ;-)

maciejhirsz:
@icefoxen I just spent a bunch of hours debugging FBOs for good-web-game so I get how fragile this is :). From everything I can tell this is a pretty safe setting, as it only applies when the canvas texture is being rendered and has no effect on how things are rendered on the canvas.

I don't want to rush you, I know how maintaining an OSS project feels like, this is just one small thing I'm really missing for the engine to be "complete" for what I need of it :upside_down_face: .

PR #636, opened 2019-05-27
Use *_be_bytes to convert colors

maciejhirsz:
Just a micro optimization. Makes for shorter Rust code and faster execution. Difference in ASM ([playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=5206dd8f38b62430a1455da83d350674)):

```asm
; manual bitshifts
from_rgb_u32:
	movl	%edi, %ecx
	andl	$16711680, %ecx
	movl	%edi, %eax
	shll	$16, %eax
	andl	$65280, %edi
	leal	(%rdi,%rax), %eax
	shrl	$16, %ecx
	orl	%ecx, %eax
	retq

; using to_be_bytes()
from_rgb_u32_fast:
	movl	%edi, %eax  ; new register
	bswapl	%eax        ; swap order (x86 is LE)
	shrl	$8, %eax    ; shift by 8, because our tuple has 3 bytes instead of 4
	retq
```

### DIFF
--- a/src/graphics/types.rs
+++ b/src/graphics/types.rs
@@ -288,39 +288,31 @@ impl Color {
 
     /// Convert a packed `u32` containing `0xRRGGBBAA` into a `Color`
     pub fn from_rgba_u32(c: u32) -> Color {
-        let rp = ((c & 0xFF00_0000u32) >> 24) as u8;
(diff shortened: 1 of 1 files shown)


icefoxen:
Oh, nice!  Thank you.

PR #648, opened 2019-07-04
Aligned drawing

Voker57:
Not sure if I did this the best way possible. Using DrawParams seems to be not an option because aligning requires information from Drawable itself, so code would need to go into draw() of all Drawable implementations.

### DIFF
--- a/src/graphics/mod.rs
+++ b/src/graphics/mod.rs
@@ -957,10 +957,51 @@ pub fn gfx_objects(
     (f, d, e, dv, cv)
 }
 
+/// Alignment of a drawable
+#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+pub struct Alignment(pub HorizontalAlignment, pub VerticalAlignment);
+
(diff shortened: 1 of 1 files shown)


icefoxen:
I... unfortunately, have no idea what this is actually *intended* to do.  Change where the origin point of the `Drawable` is?  That's what `DrawParam::offset()` is for.  Or am I missing something?

Voker57:
> I... unfortunately, have no idea what this is actually _intended_ to do. Change where the origin point of the `Drawable` is? That's what `DrawParam::offset()` is for. Or am I missing something?

It's for automatic calculation of offset using Drawable::width/height(). I thought about it some more, and maybe there should be separate params that get handled in general impl Drawable, ones that don't need to be implemented by every Drawable separately.

icefoxen:
I sympathize, but this functionality doesn't need to be part of the core library, and IMO isn't useful enough to justify the maintenance burden.  :-(  Sorry.

That said, ggez is intentionally somewhat low-level so that people can write add-on libraries on top of it.  This might be useful in something like that?

PR #653, opened 2019-07-12
Fix resizable issue on x11

### LINKED ISSUE #613: Window is resizable regardless of WindowMode field resizable (Linux X11)
Setting the `resizable` field on `WindowMode` seems to have no effect on whether or not the window can be resized on my desktop (Linux, X11, Awesome WM). That is, it can be resized regardless of the setting. Setting max and min window dimensions to the same value is a workaround.

0x1793d1:
Fix #613.


icefoxen:
:heart:

PR #661, opened 2019-07-26
adding next_dir for snake so two subsequent fast keypresses can be registered

freiguy1:
I was playing with the examples and noticed snake didn't quite work like I was expecting when pressing two arrow keys in quick succession (within the span of one update). I added a property to the snake struct to hold this optional _next_ direction. Since it wasn't a big change or refactor, I thought it could be a good improvement to go into the repo.

Thanks!


icefoxen:
Nice, thank you!

PR #662, opened 2019-07-26
fix(examples): MeshBuilder: raw -> from_raw

sigod:
Was trying out the examples, when the compiler complained:
```
error[E0599]: no method named `raw` found for type `&mut ggez::graphics::mesh::MeshBuilder` in the current scope
   --> src\main.rs:102:5
    |
102 |     mb.raw(&triangle_verts, &triangle_indices, Some(i));
    |        ^^^
```


sigod:
Actually, scratch that. Got confused by changes in release candidates.

PR #673, opened 2019-08-21
Remove deleted project from list

pradn:
The user no longer exists on GitHub either


icefoxen:
Aww, that's a shame.  Thank you for helping fight link rot though!

PR #679, opened 2019-09-02
Implement a `FontCache` type

hecrj:
This type can be useful to measure text efficiently without being tied to the `Context` lifetime.

I use it to compute text boundaries efficiently in [`iced`]: https://github.com/hecrj/iced/blob/master/examples/ggez/renderer/text.rs#L10

[`iced`]: https://github.com/hecrj/iced

### DIFF
--- a/src/graphics/text.rs
+++ b/src/graphics/text.rs
@@ -1,6 +1,6 @@
 use glyph_brush::GlyphPositioner;
 use glyph_brush::{self, FontId, Layout, SectionText, VariedSection};
-pub use glyph_brush::{rusttype::Scale, HorizontalAlign as Align};
+pub use glyph_brush::{rusttype::Scale, GlyphBrush, HorizontalAlign as Align};
 use mint;
 use std::borrow::Cow;
 use std::cell::RefCell;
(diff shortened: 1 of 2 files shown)


src/graphics/text.rs
```diff
@@ -345,22 +369,17 @@ impl Text {
...
 
     /// Returns the width of formatted and wrapped text, in screen coordinates.
-    pub fn width(&self, context: &mut Context) -> u32 {
+    pub fn width(&self, context: &Context) -> u32 {
         self.dimensions(context).0
     }
 
     /// Returns the height of formatted and wrapped text, in screen coordinates.
-    pub fn height(&self, context: &mut Context) -> u32 {
+    pub fn height(&self, context: &Context) -> u32 {
```
hecrj: We could keep taking a mutable reference in these methods... It's not necessary anymore though, and I don't think this qualifies as a breaking change (?).

icefoxen:
Awesome, thank you!  I'll have to find some actual time to look at this before merging, but from a glance it looks good.

I THINK that changing a `&mut` to `&` isn't a breaking change, but I'll have to double-check once I've had more coffee.

icefoxen:
Looks great, thank you!  There's been a few other nice contributions, so I should try to fix a couple bugs and then make a 0.5.2 release Someday.

PR #689, opened 2019-10-06
Coroutines with async/.await support

EkardNT:
I'm opening this PR to explore whether you'd be interested in having support for coroutines added to ggez. Coroutines are a pretty useful tool for easily writing asynchronous logic that spans many game frames without needing to manually implement a state machine in `update()`. [Unity's docs](https://docs.unity3d.com/Manual/Coroutines.html) have examples of how coroutines can be useful in games.

Currently the code is pretty grungy, lots of `// TODO`s, commented-out `#[deny]`s, design questions that need answering, etc. However basic coroutine support works and you can spawn futures onto either the main game thread or a thread pool. Run `cargo run --example coroutine` for a simple example.

Before I do any further work, I wanted to get a sense for whether this is something you could see being added to ggez.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ c_dependencies = ["bzip2", "mp3"]
 bitflags = "1"
 zip = { version = "0.5", default-features = false }
 directories = "2"
+futures-preview = "0.3.0-alpha.19"
 gfx = "0.18"
 gfx_core = "0.9"
 gfx_device_gl = "0.16"
--- a/examples/coroutine.rs
+++ b/examples/coroutine.rs
@@ -1,0 +1,103 @@
+//! How to use coroutines to update game state.
+
+use std::cell::Cell;
+use std::rc::Rc;
+
+use ggez;
+use ggez::event::{self, MouseButton};
+use ggez::graphics::{self, Color, Text};
+use ggez::nalgebra as na;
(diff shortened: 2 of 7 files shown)


icefoxen:
Pretty fascinating!  I made my own experiment to do something vaguely similar, though with more emphasis on I/O like loading assets.  It worked

For now I consider this out of scope for ggez itself, since it'd be a bit of a non-trivial redesign to make everything work with it.  It also doesn't necessarily need to be part of ggez, as I think your code does a good job of showing, it can be its own library that doesn't need any help from ggez itself.  A good coroutine system like this is something that would be really useful, I think.

Mad props for your comments, too.

EkardNT:
Ok I'll look for an "extras" crate to contribute this to, or create one. The only drawback to not having first class support is that we lose the ability to have core systems opt-in to being directly accessed from a different thread, but it is probably sufficient to just say that accessing ggez Context always requires synchronizing onto the main thread.

icefoxen:
The only "extras" crate I'm aware of is `ggez-goodies`, which has some handy bits and pieces but I don't have the bandwidth to maintain it well. There may be others though.

And yeah, Context being single threaded is a fairly core assumption, at least partially due to OS's enforcing that assumption for certain things as well. Significantly changing it would basically need a new graphics backend and event loop model, and that's most of ggez right there.

PR #690, opened 2019-10-08
Reset screen coordinates as default

nowakf:
I'm not sure in what situation you want screen coordinates to reference
a previous screen-size, but I think it shouldn't be default behaviour.
What do you think?

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -145,7 +145,9 @@ pub trait EventHandler {
 
     /// Called when the user resizes the window, or when it is resized
     /// via [`graphics::set_mode()`](../graphics/fn.set_mode.html).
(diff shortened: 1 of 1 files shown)


icefoxen:
I think that whatever default one chooses will be wrong for 55% of your users. ;-)

More concretely, this is a breaking change, so I'll have to think some more. I probably won't accept it, but I'll think about it. Thank you for caring enough to make a PR!

nowakf:
Thanks for making ggez! It's a joy to work with.

icefoxen:
Yeah, this one is getting closed.  ...after over a year, I know.  Either resizing the screen drags the size of the rendering projection along with it, so it draws the same thing only bigger, or you don't do that and it draws a larger area of what your view.  Neither one is conclusively better, so I'm just leaving the default as is.

PR #703, opened 2019-11-06
Fix spelling

spenserblack:
Thanks for this awesome library!

While going over the examples, I saw this small spelling error.


icefoxen:
Awesome, thank you!

Issue #704, opened 2019-11-06
Large crate size


spenserblack:
The size of this crate is a bit large (0.95 MB @ 0.5.1). I think this is because files (mostly images) tracked by git are being included in the `.crate` file produced by `cargo package` and `cargo publish`.

Would it be possible look into [excluding](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields-optional) files that aren't necessary to compile the library? It should improve download time when a user first fetches this crate.

icefoxen:
Makes sense, sure.  Thanks!

icefoxen:
Daaaaang, #707 did this for me.  `DejaVuSerif.ttf` is built into the library as the default font, so that has to be there, but it's also 371 kb.  I welcome suggestions for a smaller font to use instead if anyone has one, the requirements are:

 * Truetype
 * Compatible license
 * Decent glyph assortment (at least latin alphabets, cyrillic and arabic would be very nice as well, Indian and CJK languages would be a bonus but are large and complicated)
 * Looks cool

icefoxen:
Some candidates:

 * Calling Code, 65 kb: https://www.dafont.com/calling-code.font?l[]=10 , latin only
 * Bitstream Vera Sans, 66 kb: https://www.dafont.com/bitstream-vera-sans.font?l[]=10 , latin only
 * Liberation Sans, 133 kb: https://www.dafont.com/liberation-sans.font?l[]=10 , latin and cyrillic

Good font viewer so you can see what glyphs they actually have: http://mathew-kurian.github.io/CharacterMap/

Contrast with the current DejaVu font which includes latin, cyrillic and also arabic glyphs... but Calling for opinions on people who use non-Latin-character-set languages and/or have slow internet... which do you prefer?  Keeping in mind that [Arabic doesn't render correctly anyway...](#378) ...

ceranco:
I found another nice one:
* Comme, 52 kb: https://www.fontsquirrel.com/fonts/comme

How do you check the if cyrillic is supported?

icefoxen:
I loaded it up in the font viewer app I linked and paged through thr character map looking for Cyrillic characters.  :D

ceranco:
Makes sense :sweat_smile:
I updated my previous comment to add the glyph support of the font.

icefoxen:
I like Comme, but let's use Liberation Sans for now, just to get a LITTLE bit of support for non-western-European languages by default.

[...]

PR #707, opened 2019-11-14
Reduce crate size

ceranco:
Hi, I saw issue #704 and thought I'd create a PR for it.

I excluded almost all of the `resources` folder from the `[package]` in `Cargo.toml` to reduce the crate size.
I left *'DejaVuSerif.ttf'* alone as removing it broke the `cargo package` build:
https://github.com/ggez/ggez/blob/e3cc91902e8dd993597e27cb81adde16301fcbe9/src/graphics/text.rs#L420-L425

This change reduces the size of **'ggez-0.5.1.crate'** from **946KB** to **380KB**.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ rust:
   - stable
   - beta
   - nightly
-  - 1.33.0
+  - 1.36.0
 os:
   - linux
   - osx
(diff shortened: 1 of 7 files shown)


icefoxen:
Oh snap, thank you so much!  I made some comments about `DejaVuSerif.ttf`in the original thread, any ideas you have are welcome.

icefoxen:
(Note to self, changed target from `master` to `devel` 'cause that's where it should go, and merged `master` into `devel` separately anyway 'cause it needed to happen.)

PR #711, opened 2019-11-22
[devel] Flatten code of 04_snake example

imanghafoori1:
Flatten code of 04_snake example ;)

### DIFF
--- a/examples/04_snake.rs
+++ b/examples/04_snake.rs
@@ -416,36 +416,38 @@ impl event::EventHandler for GameState {
     /// Update will happen on every frame before it is drawn. This is where we update
     /// our game state to react to whatever is happening in the game world.
     fn update(&mut self, _ctx: &mut Context) -> GameResult {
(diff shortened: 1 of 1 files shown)


examples/04_snake.rs
```diff
@@ -416,36 +416,38 @@ impl event::EventHandler for GameState {
...
-                        }
-                        // If it ate itself, we set our gameover state to true.
-                        Ate::Itself => {
-                            self.gameover = true;
-                        }
+        // First we check to see if enough time has elapsed since our last update based
+        // on the update rate we defined at the top.
+        // if not, we do nothing and return early.
+        if !(Instant::now() - self.last_update >= Duration::from_millis(MILLIS_PER_UPDATE)) {
+            Ok(())
```
icefoxen: Now that I look at it again, this needs to be `return Ok(());`

imanghafoori1: Oh, a silly mistake !

icefoxen:
Thank you so much!

PR #714, opened 2019-12-08
Add more color constants

Uriopass:
Added several color constants, limited to combinations of 1 and 0.
Very useful for prototyping :-)


icefoxen:
Nice.  Thank you!

PR #715, opened 2019-12-09
File paths in zip resource are also supposed to start with '/'

### LINKED ISSUE #682: Inconsistent path prefix between directories and zip files
I'd like to load my resources both from the zip file and the resources folder.
Default resources would be baked into the executable, and additional resources either in resources.zip or resources/ directory could be used to mod the game.

My problem is that resources in the zip files are obtained without the leading slash.

Example:

```rust

// resources/
let img = Image::new(ctx, "/background.png")?;

// resources.zip
let img = Image::new(ctx, "background.png")?;
```

Is there any reasoning behind this? How can I deal properly with it?

Should i just do something like:

```rust
pub fn path(ctx: &Context, path: &str) -> String {
    let slash_path = String::from("/") + path;

    if filesystem::is_file(ctx, &slash_path) || filesystem::is_dir(ctx, &slash_path) {
        slash_path
    } else {
        String::from(path)
    }
}
```

And then later:

```rust
let img = Image::new(ctx, path("background.png"))?;
```

starccy:
fix #682 problem.

### DIFF
--- a/src/vfs.rs
+++ b/src/vfs.rs
@@ -537,7 +537,9 @@ trait ZipArchiveAccess {
 
 impl<T: Read + Seek> ZipArchiveAccess for zip::ZipArchive<T> {
     fn by_name(&mut self, name: &str) -> zip::result::ZipResult<zip::read::ZipFile> {
-        self.by_name(name)
+        let filename = sanitize_path(Path::new(name))
+            .unwrap_or(PathBuf::from(&name));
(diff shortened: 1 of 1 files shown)


icefoxen:
Thank you!  I need to test this out and make sure it works as intended, maybe this weekend.  We should also add some unit tests so it doesn't happen again.  :disappointed:

icefoxen:
Looks good, thank you!

PR #716, opened 2019-12-09
Faster to_na_matrix (especially in debug)

jonaskje:
I noticed most cpu was spent in DrawParam::to_na_matrix when drawing a sprite-batch in debug.

This PR inlines the calculation of the matrix which makes to_na_matrix much faster. 

I left the original matrix multiplication code in there for reference.


icefoxen:
**You are my hero**.  I have been annoyed about this for *ages* but didn't have a good idea for how to make it better.  :heart:

PR #721, opened 2019-12-18
Add gamepads iterator wrapper

gameldar:
Adds functionality to be able to iterate through the connected gamepads - this is just wrapping the gilrs::gamepads function so that the ggez version of the GamepadId is available to use with the results.


icefoxen:
Looks good, thank you!  Gamepads in general need a bit of love, I confess, but this is a good start.

gameldar:
> Looks good, thank you! Gamepads in general need a bit of love, I confess, but this is a good start.

This was the only issue I've hit so far and that is just because I wanted to discover how many players based on the number of controllers. But I have only scratched the surface at this point!

PR #722, opened 2019-12-22
Update to nalgebra 0.19

aeubanks:


icefoxen:
Not going to happen I'm afraid, because we re-export `nalgebra` and so the exact version we use is part of the public API.  Version bumping it implies a breaking change to ggez.  :-(

aeubanks:
What about for 0.6? I don't think it makes sense to stay on an older version of nalgebra forever.
e.g. I'm running into conflicts that other crates (e.g. nphysics) that also export nalgebra.

icefoxen:
Definitely will be updated for 0.6, but might also get removed entirely.  Since ggez's public API now takes `mint` types for everything we don't need to export nalgebra at all anymore.

aeubanks:
Ah I see, thanks for the explanation!

PR #723, opened 2019-12-22
Update to rodio 0.10

aeubanks:


icefoxen:
this might be okay, I don't think we directly expose any of `rodio`'s types in our API.

PR #728, opened 2019-12-28
Add gravity-sim-v2 to project list

mkhan45:
It's not quite a game but I think it demonstrates some of the features of ggez nicely. Thanks!


icefoxen:
I wasted almost an hour just playing with this. Awesome, thanks!

PR #730, opened 2019-12-30
Nit: missed final step in comment calculation

PrototypeNM1:
The original to_na_matrix code was left as an explanatory comment, but the return value calculation was omitted - https://github.com/ggez/ggez/pull/716/files. A nitpick but it felt incomplete.


icefoxen:
Perfect,  thanks !

PR #733, opened 2020-01-12
Mesh batching and canvas improvements

bazhenovc:
Hi,

My application involves some distance field rendering on the GPU, which needs a floating point render target. I've added an extra parameter to Canvas::new() function and a bunch of other really minor things.

The other thing is the "MeshBatch" struct that can be used to render instanced geometry, works similarly to the current "SpriteBatch"-es, but tries to cache instance data and only update it when necessary.

Thanks.

### DIFF
--- a/src/graphics/mesh.rs
+++ b/src/graphics/mesh.rs
@@ -686,7 +686,7 @@ impl Drawable for Mesh {
 
 /// An index of a particular instance in a `MeshBatch`
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct MeshIdx(usize);
+pub struct MeshIdx(pub usize);
 
 /// Mesh that will be rendered with hardware instancing.
 /// Use this when you have a lot of similar geometry which does not move around often.
@@ -720,13 +720,27 @@ impl MeshBatch {
         self.instance_buffer_dirty = true;
     }
 
+    /// Returns a reference to mesh instances
+    pub fn get_instance_params(&self) -> &[DrawParam] {
+        &self.instance_params
+    }
+
+    /// Returns a mutable reference to mesh instances.
+    ///
+    /// Please note that manually altering items in this slice
+    /// will not automatically invalidate the buffer, you will
+    /// have to manually call `flush()` or `flush_range()` later.
+    pub fn get_instance_params_mut(&mut self) -> &mut [DrawParam] {
+        &mut self.instance_params
+    }
+
     /// Adds a new instance to the mesh batch
     ///
     /// Returns a handle with which to modify the instance using
     /// [`set()`](#method.set)
     ///
     /// Calling this invalidates the entire buffer and will result in
-    /// flusing the entire buffer on the next [`graphics::draw()`](../fn.draw.html) call.
+    /// flusing it on the next [`graphics::draw()`](../fn.draw.html) call.
     pub fn add<P>(&mut self, param: P) -> MeshIdx
     where
         P: Into<DrawParam>,
@@ -795,11 +809,18 @@ impl MeshBatch {
         count: usize,
     ) -> GameResult {
         let first_param = first_handle.0;
-        if first_param < self.instance_params.len()
-            && (first_param + count) <= self.instance_params.len()
-        {
-            let new_properties: Vec<InstanceProperties> = self.instance_params
-                [first_param..(first_param + count)]
+        let slice_len = first_param + count;
+        if first_param < self.instance_params.len() && slice_len <= self.instance_params.len() {
+            let needs_new_buffer = self.instance_buffer == None
+                || self.instance_buffer.as_ref().unwrap().len() < slice_len;
+
+            let slice = if needs_new_buffer {
+                &self.instance_params
+            } else {
+                &self.instance_params[first_param..slice_len]
+            };
+
+            let new_properties: Vec<InstanceProperties> = slice
                 .iter()
                 .map(|param| {
                     let draw_transform = DrawTransform::from(*param);
@@ -807,9 +828,7 @@ impl MeshBatch {
                 })
                 .collect();
 
-            if self.instance_buffer == None
-                || self.instance_buffer.as_ref().unwrap().len() < new_properties.len()
-            {
+            if needs_new_buffer {
                 let new_buffer = ctx.gfx_context.factory.create_buffer(
                     new_properties.len(),
                     gfx::buffer::Role::Vertex,
--- a/src/graphics/mesh.rs
+++ b/src/graphics/mesh.rs
@@ -221,7 +221,9 @@ impl MeshBuilder {
         self.polyline_inner(mode, points, true, color)
     }
 
-    fn polyline_inner<P>(
+    /// Create a new mesh for a given polyline.
+    /// The points given must be in clockwise order.
+    pub fn polyline_inner<P>(
         &mut self,
         mode: DrawMode,
         points: &[P],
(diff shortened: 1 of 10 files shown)


icefoxen:
Wow!  Uh, dang.  I like it but it's a big change, let me look things over a little.  Mesh batching is something I've wanted for a while but have been mainly addressing in the `ggraphics` branch.

bazhenovc:
Sure, no problem. Let me know if you'd like to change or add something to this PR, I'll be happy do do it.

By the way, there seems to be an issue with the CI - all pull request builds failed with the same message, which is unrelated to the changes made.

icefoxen:
Yeah, I am depressingly aware of the CI issues.  Thanks.

src/graphics/mesh.rs
```diff
@@ -221,7 +221,9 @@ impl MeshBuilder {
         self.polyline_inner(mode, points, true, color)
     }
 
-    fn polyline_inner<P>(
+    /// Create a new mesh for a given polyline.
+    /// The points given must be in clockwise order.
+    pub fn polyline_inner<P>(
```
icefoxen: Why is this made `pub`?  Can it be `pub(crate)` instead?

src/graphics/mesh.rs
```diff
@@ -230,6 +232,26 @@ impl MeshBuilder {
...
+        &mut self,
+        mode: DrawMode,
+        points: &[P],
+        is_closed: bool,
+        vb: V,
+    ) -> GameResult<&mut Self>
+    where
+        P: Into<mint::Point2<f32>> + Clone,
+        V: t::VertexConstructor<t::FillVertex, Vertex>
+            + t::VertexConstructor<t::StrokeVertex, Vertex>,
```
icefoxen: Exposing `lyon`'s `VertexConstructor` types publicly here are probably not a good idea, since we don't re-export `lyon` as a whole.

src/graphics/mesh.rs
```diff
@@ -624,6 +644,16 @@ impl Mesh {
...
         self.slice = slice;
     }
+
+    /// Returns a slice for this mesh that could be used for manual draw call submission
+    pub fn get_slice(&self) -> &gfx::Slice<gfx_device_gl::Resources> {
+        &self.slice
+    }
+
+    /// Returns a vertex buffer for this mesh that could be used for manual draw call submission
+    pub fn get_vertex_buffer(&self) -> gfx::handle::Buffer<gfx_device_gl::Resources, Vertex> {
```
icefoxen: Are `get_slice()` and `get_vertex_buffer()` necessary?  I'm not sure I can think of a situation where I'd want to use them.

src/graphics/mesh.rs
```diff
@@ -656,6 +686,156 @@ impl Drawable for Mesh {
...
+            MeshBatch::create_instance_buffer(ctx, instance_params)?;
+
+        self.instance_properties = instance_properties;
+        self.instance_buffer = instance_buffer;
+
+        Ok(())
+    }
+
+    /// Updates instance parameters in a batch without reallocating existing buffers
+    pub fn update_instance_params(
```
icefoxen: Might want to think about the API for `*_instance_params()` and one way or another make it the same as `SpriteBatch`, so that both batch types work exactly the same way.

src/graphics/mesh.rs
bazhenovc: Good point, there's no reason for this to be public. I'll fix that.

src/graphics/mesh.rs
bazhenovc: I've added this to be able to provide a custom vertex builder to build polylines which in turn were needed to generate proper UVs and vertex colors and currently there's no reasonable way of doing that. It is possible to iterate over vertices after the mesh was built, but this would over complicate the code and dramatically reduce performance.

I agree that this is probably not a good idea, I'd love to discuss this issue further.

src/graphics/mesh.rs
bazhenovc: They are, my distance field rendering is rendering to multiple render targets simultaneously. ggez currently does not support that natively, so I'm manually submitting draw calls to the gfx encoder, which requires access to the slice and vertex buffer.

src/graphics/mesh.rs
bazhenovc: That's a good idea, I'll do that. Do you think it's worth adding something like a "DrawableBatch" trait and implement it for SpriteBatch and MeshBatch?

As a side note, I don't really like that SpriteBatch calls flush() internally on every draw call - it really tanks performance, especially in debug builds. Since you've mentioned a graphics engine rewrite, I'd like to ask if there's any chance of swapping the global instance buffer to local instance buffers for separate batches that are not invalidated every frame?

bazhenovc:
Sounds good, I'll make the changes you've requested and be back in a few hours.

src/graphics/mesh.rs
icefoxen: I wouldn't consider it worth making a trait to combine the two *yet*, since that would probably result in extra trait-type fluff when you have to say "what kind of drawable can I insert into this batch?"

For the side note, yeah, check out https://github.com/ggez/ggez/blob/adc36f8b151a1be30c43271094f340f0d70f9fda/src/graphics/mod.rs#L189 and associated stuff in https://github.com/ggez/ggez/tree/ggraphics/ggraphics .  Each `DrawCall` has its own instance buffer, and the order in which they are drawn and switched between becomes very explicit.  Any ideas, suggestions or wishlists you have for that codebase are welcome, I'm taking a break from it a little to chew on other problems.

src/graphics/mesh.rs
bazhenovc: Lyon 0.14 has them removed in favor of regular closures, maybe upgrading it will solve this issue?

src/graphics/mesh.rs
bazhenovc: Sounds good, I'll check it out.

bazhenovc:
I've reworked the mesh batch, please let me know what do you think.

Right now it does not implement Drawable because draw() has to be mutable in order to be able to flush the intstance buffer. I'm not really happy about it, but I feel like changing the signature for Drawable requires your approval.

src/graphics/mesh.rs
icefoxen: That's an interesting idea, I'll have to try...

bazhenovc:
Cool, thanks!

I've had a quick look at the ggraphics branch that you've mentioned, currently it does not seem to support arbitrary meshes, which is something that I absolutely need.

icefoxen:
Nope, that specific functionality is still WIP.  Sorry I didn't make that clear.  XD  But actually making it function is intended to be entirely orthogonal to any other drawable type, auto-batching and all.

bazhenovc:
Oh ok, in that case I think it might be still worth merging this. Even if this mesh batch stuff will be removed later, it could be fairly useful before the graphics engine rewrite will ship.

One more side note, do you think it's worth implementing a similar instance buffer caching for SpriteBatch? I can do this for the same pull request.

icefoxen:
Yeah, that sounds pretty reasonable when you put it that way.

And, I don't know.  It's pre-coffee for me at the moment, so, what benefits would this code have for SpriteBatch?

bazhenovc:
Sorry for the late reply, was busy with some real life things.

In short, the answer is "Just for performance", existing API and behaviour will not be changed, only the internal implementation.

`SpriteBatch` has `flush()` method which is called unconditionally in `draw()`. Internally it loops over the `self.sprites` array and converts them to GPU-friendly `InstanceProperties` structure. With 500k sprites debug build needs a few _seconds_ to render the frame on my machine, which makes it impossible to debug it (it has to be interactive).

`MeshBatch` solves this by having a separate GPU buffer for instance data and a dirty flag, its `flush()` method is only called when something in the batch was changed, so it has zero runtime cost for static batches. This comes at the cost of increased memory consumption, but this should be negligible.
However it still flushes the entire buffer no matter how many items are changed, so it's impossible to have something animated and remain interactive in debug builds.

I think the best solution to this would be to have a similar implementation to `MeshBatch`, but making `flush()` public and adding some `flush_range()` method that will allow to specify which part of the buffer needs updating.
It will still flush the entire buffer by default, however that would give the user an option to explicitly update buffer regions in case performance becomes an issue, and at the same time the API will remain compatible with earlier versions.

All this sort of depends on that mutable `draw()` function in `Drawable` that we've discussed earlier.

What do you think?

bazhenovc:
So I've added that stuff in a403962, let me know if you'd like to have a similar system for SpriteBatch.

And I'll also add an example that demonstrates how to use this flushing.

bazhenovc:
Example added in ddabeba

icefoxen:
Sounds good to me, thanks!  Let me play around with it a bit and see if I can break it.  :-)

icefoxen:
Note to self: it looks like I've gotten ggez's devel branch unhecked, test this more.

icefoxen:
Looks good to me.  Merged with a glad heart, and thank you for your patience!

PR #735, opened 2020-01-14
Added v0.5 ggez/specs example with walkthrough text

Fudance:
I've added a new example/tutorial to the Projects.md. This is an example of using specs with ggez, including some basic systems for movement and collision detection.


icefoxen:
Awesome, thank you!

PR #740, opened 2020-01-22
[WIP] Multiple states

telastrus:
I've been working on a system for being able to have multiple states running without needing to create a `Vec` to store them, and I want to get input on this before I get too far in. Mainly, I reworked some of the event loop execution that broke nearly all of the examples, and I'd like to know if there's a better way to do that.

I added an example ([`multiple_states`](https://github.com/telastrus/ggez/blob/multiple_states/examples/multiple_states.rs)), to demonstrate exactly what I'm working on.

### DIFF
--- a/examples/05_astroblasto.rs
+++ b/examples/05_astroblasto.rs
@@ -596,8 +596,8 @@ pub fn main() -> GameResult {
         .window_mode(conf::WindowMode::default().dimensions(640.0, 480.0))
         .add_resource_path(resource_dir);
 
-    let (ctx, events_loop) = &mut cb.build()?;
+    let (mut ctx, events_loop) = cb.build()?;
 
-    let game = &mut MainState::new(ctx)?;
-    event::run(ctx, events_loop, game)
+    let game = &mut MainState::new(&mut ctx)?;
+    event::run(&mut ctx, events_loop, game)
 }
--- a/examples/multiple_states.rs
+++ b/examples/multiple_states.rs
@@ -4,9 +4,9 @@ use cgmath;
 use ggez;
 
 use ggez::event;
+use ggez::event::{KeyCode, KeyMods};
 use ggez::graphics;
 use ggez::{Context, GameResult};
-use ggez::event::{KeyMods, KeyCode};
 use std::env;
 use std::path;
 
@@ -105,7 +105,10 @@ impl event::EventHandler for PausedState {
 
         // Drawables are drawn from their top-left corner.
         let rect = ggez::graphics::screen_coordinates(ctx);
-        let dest_point = cgmath::Point2::new((rect.w - self.text.width(ctx) as f32) / 2.0, (rect.h - self.text.height(ctx) as f32) / 2.0);
+        let dest_point = cgmath::Point2::new(
+            (rect.w - self.text.width(ctx) as f32) / 2.0,
+            (rect.h - self.text.height(ctx) as f32) / 2.0,
+        );
         graphics::draw(ctx, &self.text, (dest_point,))?;
         graphics::present(ctx)?;
 
--- a/examples/multiple_states.rs
+++ b/examples/multiple_states.rs
@@ -6,6 +6,7 @@ use ggez;
 use ggez::event;
 use ggez::graphics;
 use ggez::{Context, GameResult};
+use ggez::event::{KeyMods, KeyCode};
 use std::env;
 use std::path;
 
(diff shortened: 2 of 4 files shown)


icefoxen:
Interesting approach, I generally do it the other way around -- just by having my `MainState` or whatever contain other states, usually as a stack.  Not sure I want this integrated in `ggez` itself.  I'm also fairly suspicious of the `unsafe` bit of code it involves.  It mutates the `ctx.state_context` from something inside the `ctx.state_context`, so it seems like it a real good way to make Strange Stuff happen if, say, you call `push_state()` multiple times for something like that.  Seems like just having a `Vec` of states would be simpler.

It's a cool idea, but I'm unfortunately going to default to "no" for the PR unless there's a convincing case for it.

telastrus:
I'd just like to clarify, do you think the concept shouldn't be implemented or the implementation itself?

In the first case, how do you keep `MainState` from becoming a monolith? I feel like games would benefit from being able to have, say, a distinct `SingleplayerState` and `MultiplayerState` (or similar). That way, you don't have to load both components in the `MainState`.

Also, how do you switch between states in your methodology? Do you just redirect the `draw` and `update` calls to whatever state is active? I feel like that switching code ends up being part of every project that wants to use multiple states.

In the second case, I can agree to some of the concerns. I thought of using a `Vec`, but that ends up being a `Vec<Box<dyn EventHandler>>`, which means the game states have to be heap-allocated. If I understand Rust correctly, my implementation works on stack-allocated states instead, which should be slightly faster.

The `unsafe` is a bit of premature optimization, and I should be able to get rid of it if desired. Mainly, since the event context can only be dropped if the context itself is dropped (from the public-facing interface), I decided to take a reference of the `event_context` separately (avoiding checks during each `poll_events` call), so I could still pass `ctx` to the state. No `Strange Stuff` should be able to happen in theory unless threading is going on (for which `Context` itself isn't thread-safe anyway), as each call to `push_stack` nests from previous calls, so no reference to `ctx` (and, as a result, `state_manager` unless I `unwrap` it incorrectly) can get dropped until the top-level loop ends. But, if this is too fishy, I should be able to take it out entirely without too much issue.

Sorry if I'm asking a lot of questions. I'm also trying to learn a bit of (good) game development strategies and advanced Rust, so I'd really appreciate it if you could answer them!

icefoxen:
The concept *arguably* shouldn't be a core part of ggez, and the implementation as written is a bit more clever than it needs to be.

The code I usually use is here: https://github.com/ggez/ggez-goodies/blob/master/src/scene.rs and an example of what it looks like for a simple game is here: https://github.com/icefoxen/ld42/blob/master/src/main.rs#L108 , with the scenes themselves in the `scene/` subdir of that same repo.  As you described, it just redirects `update()` and `draw()` calls to whatever scene is on top of the stack, and the switching code is handled by the library.  You are correct that some setup code gets used in basically every game that uses it, but the amount is pretty small.  The more complicated bits, which is mostly scene switching, are only written once.

For performance, as far as I'm concerned I don't particularly care if an extra allocation or two happens when a scene change occurs, or if an extra pointer dereference or bounds check happens once per game frame.  These are tiny performance costs that happen pretty rarely.  If these things happened once per sprite in a SpriteBatch, or once per vertex in MeshBuilder, then that would probably be different.  ggez in general prioritizes simplicity over performance, as one of my goals is to make it easy for newbies to use and tinker with (as you are doing!), hence the cost of fishy code is pretty high.

Which takes us back to the original question: *should* ggez add scene-changing functionality?  With any new feature, my default answer is going to be "no" unless a) there's an obvious hole in functionality, b) there's no very reasonable way to do it separate from ggez, or c) the maintenance burden for maintainers and learning curve for users is pretty small.  The currently open PR #733 is actually a real good example of a feature where all three of those things is true.

That said, this feature *is* something that almost any non-trivial game will want to have sooner or later, which is a pretty strong argument.  It fits with ggez's goal of giving you all the basic stuff you need in one package.  But it's one more thing to explain how to use, and it fails the test of "can you make this reasonably as an external library" as well as "would you ever decide it's too limiting and want to write your own anyway".

telastrus:
I don't quite understand why you feel that it would be another thing that beginners would have to learn, as the only thing that changes from a beginner's perspective is removing an ampersand from the `event::run` call (they don't need to understand state switching to continue using the game engine as is). Also, I don't quite know how this could be made reasonably as an external library (while still keeping the interface simple) without having to write a new `EventHandler` trait and a new event loop with it (which I'll probably have to do for my game).

But I understand all the other complaints, so I'll put it in your hands. If you don't think it's a good idea at all, close the PR.

icefoxen:
> they don't need to understand state switching to continue using the game engine as is

It's just another layer of hidden state and complexity that "magically" happens for them.  I dislike magic.

> Also, I don't quite know how this could be made reasonably as an external library (while still keeping the interface simple) without having to write a new EventHandler trait and a new event loop with it

In my experience, doing this is a little annoying, but also takes about twenty minutes to do once.  So the idea of adding more global mutable state to avoid it isn't very appealing.

Sorry!  Priorities.  :man_shrugging:   Everything in ggez is something that has to be maintained, explained, and fixed when it breaks.  It *would* be nice if `EventHandler` could be used transparently for a scene system, but I don't see a way to do that without making it more complicated.  Importing or writing a scene library and implementing a three-method trait for it is not really something I lose sleep over.

PR #752, opened 2020-03-07
Fix SpriteBatch link

j-browne:
There was a broken link in the documentation for the `spritebatch` module.


icefoxen:
Nice catch, thank you for the fix.

PR #766, opened 2020-04-17
Switched to anyhow error in handler and event run()

rdelfin:
Added anyhow as the return type for the handler update() and draw() as well as the event loop run.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
 c_dependencies = ["bzip2", "mp3"]
 
 [dependencies]
+anyhow = "1.0.28"
 bitflags = "1"
 zip = { version = "0.5", default-features = false }
 directories = "2"
--- a/src/error.rs
+++ b/src/error.rs
@@ -9,6 +9,7 @@ use gfx;
 use glutin;
 use winit;
 
+use anyhow;
 use gilrs;
 use image;
 use lyon;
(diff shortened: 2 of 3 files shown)


rdelfin:
...whoops

icefoxen:
...well now I have to hear more.

rdelfin:
@icefoxen well, if you want to follow what I'm doing: https://github.com/rdelfin/ggez/pull/1

TL;DR I'm switching the EventHandler `update()` and `draw()` to return an `anyhow::Result<()>` instead of a `ggez::GameResult` so my game can throw any custom error from my game loop.

I think it would be useful but I wasn't sure if you'd want the pull request since it could be an API change (would require people change return types and change their main to take an `anyhow::Result`) and I'm still not 100% sure if this is the right call, but I'll send the pull request when I'm done if you think it would be useful :D

icefoxen:
Hah, thanks.  I'm not sold on `anyhow` as an error handling crate, since for libraries in general I've found anything beyond pure rust error handling to not be terribly useful for the amount of trouble it gives.  `failure` for example adds like 10 seconds of compile time for... not a whole lot, and requires that the game maker use `failure` as well.  But it's interesting to see how ggez would handle it.

rdelfin:
I think that's a fair point on it. I imagined it was way too specific a use to warrant bringing in a new library. Out of curiosity, generally how have you seen people do error handling inside of the event loop?

icefoxen:
Generally if there's an error that involves returning from the event loop, it's pretty fatal anyway, so figuring out how to return it isn't terribly relevant.  Usually I end up either just exiting with the error, or trying to keep things running and displaying the error in a log of some kind.

PR #769, opened 2020-04-18
Added EventHandler support for anyhow errors

rdelfin:
One of the issues with ggez as it stands right now is you can only return ggez errors through the result in the `EventHandler`'s `update()` and `draw()`. This makes it so that any error handling code that you'd want to cause an error all the way through to main will have to either panic, or get somehow converted into a `GameResult` object, which will likely loose any context on the original error.

My proposed solution on this is to use the anyhow crate. This allows you to return any error that implements `std::error::Error`. It essentially acts like a `Box<dyn std::error::Error>`, with a couple of caveats and added features (like guaranteed backtraces and the ability to add context to an error). While this is generally not suited for libraries since it can be overly broad and partially boxes any user of the library to use `anyhow` for error handling, I believe this is appropriate here since there's not many other ways of allowing users to return any error result they want through the event handler.

That all being said, if there's any suggestions for what a better way of doing error handling here is (even if that suggestion is leaving it as is), I'm all ears.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
 c_dependencies = ["bzip2", "mp3"]
 
 [dependencies]
+anyhow = "1.0.28"
 bitflags = "1"
 zip = { version = "0.5", default-features = false }
 directories = "2"
--- a/README.md
+++ b/README.md
@@ -101,14 +101,15 @@ to common problems.
 ### Basic Project Template
 
 ```rust,no_run
-use ggez::{graphics, Context, ContextBuilder, GameResult};
+use anyhow;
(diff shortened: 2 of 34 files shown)


rdelfin:
See #766 for context

icefoxen:
Main arguments against: It adds another crate to the public API that users will have to figure out how to use, and frankly by now we've gone through enough error-handling crates (`error_chain`, `failure`, and now `anyhow`) in the last three years or so that I have little faith that `anyhow` won't be replaced by the next new shiny thing by mid 2021.

Being able to let users return their own errors is nice, but we could easily make the `EventLoop` type generic over the error type or make it return `dyn Error` without a third-party library.

rdelfin:
Actually... A `Box<dyn Error>` might actually be a good idea. Might send out a PR for that

PR #771, opened 2020-04-22
Added a .center() method to Rect to find center coords

mkhan45:
Well I'm not sure if there's a reason it's not added, but I use it fairly frequently so I thought I'd make a PR.


icefoxen:
Hey, that's pretty handy. Thanks!

Issue #640, opened 2019-06-03
Does not gracefully handle it if the main loop update is too slow


icefoxen:
**Describe the bug**
Described here: https://www.reddit.com/r/rust/comments/bvrkpx/a_guide_to_rust_2d_game_frameworks_2019/epv2f4q/?context=3

> Additionally, I've encountered situations where I've been able to destabilize the main loop; 
> i.e. where the update method runs too slow and ggez keeps running update, failing to 
> draw or handle inputs in a timely manner.

**To Reproduce**

At a GUESS, you make a loop with `event::check_update_time()` that takes longer than the given update time to execute.  I believe it is possible to make the main loop hang forever by doing that, and it won't ever get around to checking for events or handling quit requests.  Need to double check.

**Expected behavior**

Not sure!  But making everything hang forever is obviously wrong.   Options:

 * Check, and log a warning when this happens
 * Check, and if the main loop is taking too long then yield anyway to handle events before trying again.  Your game will lag, but won't hang.
 * Check, and return `Result` from `check_update_time` so we can distinguish between "time is caught up", "time is not caught up", and "it doesn't look like time will ever catch up".  Basically, incorporate a watchdog into it and force the user to handle the case where it takes too long.

**Screenshots or pasted code**

todo

**Hardware and Software:**
 - ggez version: 0.4.4 or 0.5.0-rc.X; `check_update_time()` shouldn't really be different between them

Wumpf:
Easiest (and I think typical) way to solve this is to have max number of times `check_update_time` returns true for every `update`. This is essentially how Unity solves the issue as well ([source](http://answers.unity.com/answers/11583/view.html)): An update step (== `check_update_time`) has a fixed dt, but the sum of those steps should never be over a given max timespan per update, otherwise one runs into a feedback loop with more and more update steps to catch up per update.

Alternatively `check_update_time` could clamp `residual_update_dt` to zero if "current time spend updating" + "last render timer" took already more than the available time budget... but sounds to me like this gets quite tricky and hard to control 🤔

icefoxen:
...That's honestly a great idea, and simple to implement.  Thank you!

icefoxen:
Possible link to add as a reference: https://joetsoi.github.io/fix-your-timestep-rust-ggez/

0----0:
The way examples in ggez devel currently treat #774 , which is to use `?` to unwrap `check_update_time()`, exiting the application on failure, behaves undesirably in two cases:
[...]

PR #774, opened 2020-05-11
Iss640

keatoooon:
Fix for issue #640 

Instead of returning a bool, check_update_time() will now return a GameResult<bool>. If called more than MAX_INCREMENTAL_UPDATES times per call to update(), it will return TimeCatchupError, which is a new GameError I added for this purpose. Otherwise, check_update_time() returns Ok(true/false) as it did before. All the examples and documentation should be updated accordingly.

Let me know if there are any issues/changes that you would like me to make


icefoxen:
Oooh, I actually quite like this.  Sorry it's taken me like five months to incorporate it.  I want to do a bit more checking but expect this to get merged into the 0.6 release.

PR #785, opened 2020-06-24
Add a `CustomError` variant to `GameError`.

gamma-delta:
When a game is running, there may be errors that ggez does not know how to handle. For example, the game may be in an incorrect state, or data it has loaded from a file may be malformed.

Add a `CustomError` variant to `GameError` so these custom errors can be handled in the same way as other ggez errors, instead of having to create your own error type.

Example:

```rs
fn update(&mut self, ctx: &mut Context) -> GameResult<()> {
  let mut player = self.player;
  if player.x < 0 {
    return GameError::CustomError(format!("Player was out of bounds at x = {}", player.x));
  }
// ... continue updating
}
```

Without a CustomError type, it is difficult to safely exit with an error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ addons:
 cache: cargo
 
 script:
-  - if [[ "$TRAVIS_OS_NAME" != "windows" ]]; then cargo test headless; cargo test --doc; cargo test --test skeptic; fi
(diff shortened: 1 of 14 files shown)


icefoxen:
Basically re-implemented in commit c4c056f19bdfb634c7aa1217d638f63ae42eb20e , since the branches with this one seem to have gotten a little tangly.  Thank you!

Issue #697, opened 2019-10-24
0.6.0 release checklist


icefoxen:
I think I'm going to make this a bugfix-and-update release, focusing on upgrading to winit 0.20... once that ever gets released.

Things that should get done:

 * [x] ~~Upgrade to winit 0.20.  `glutin` 0.21 makes some really annoying API changes, unfortunately, so I'm currently trying to avoid it... not sure how feasible that will be in the long run though.~~ -- ~~Can't be done without updating `gfx_window_glutin`, which uses winit 0.19 and hasn't had a new release.  We could try to make a `gfx_window_glutin` update, or we could vendor it all since it's pretty small and update it that way, or we could just leave it as it is and deal with old, somewhat-buggy winit 0.19.~~  Done in #837 , thanks to sapir and imbris for making this possible!
 * [x] Remove `nalgebra` in favor of ~~`euclid`~~ `glam`, if possible.  Lyon already uses `euclid`, so in THEORY, using only `euclid` will make compile times faster.  ...In practice, so far, `rodio` 0.9 uses `nalgebra` anyway, and the current git master of it removes `nalgebra` but adds `failure` which more than eats up the gained compile time.  I just can't win.  Let's just use `glam` which at least only takes half a second to compile.
 * [x] Remove rand from examples in favor of `oorandom` or something similarly minimal
 * [ ] Maybe try to make `gilrs` a little less of a pain
 * [x] Version bump rodio
 * ~~Go through the audio API and consistent-ify it with the new rodio.  Might need a total rewrite but it's relatively small, and should make things simpler since the new rodio API looks a bit better-organized than the old one.~~ I really can't handle a full rewrite right now.
 * [x] Version bump glyph_brush
 * [x] Version bump lyon
 * [x] Version bump whatever else presents itself

...In light of the `glutin` thing I'm really not sure this is worth the trouble.  Maybe just stick with the current version of `glutin`, fix everything else, then call it 1.0.  Meh.  We should try to make a big pile of bugfixes anyway though.

alexheretic:
If you're still aiming to use pre-ll gfx I'm moving Robo Instructus to [alexheretic/old-school-gfx-glutin-ext](https://github.com/alexheretic/old-school-gfx-glutin-ext) _(replaces gfx_window_glutin)_ which is much easier for me to keep up to date with glutin.

icefoxen:
[...]

PR #790, opened 2020-07-01
Bump `nalgebra` dependency

fosskers:
This bumps the `nalgebra` dependency in order to get away from an old version of `rand` that is injected. Currently 3 versions of `rand` are included in the `ggez` dependency graph, with the oldest coming via `skeptic`. Improving that is tracked here: https://github.com/budziq/rust-skeptic/pull/115.

`nalgebra` and `mint` seem to still be compatible, as marked in `Cargo.toml`.

Tests don't seem to pass locally - it freezes for me after a while when trying to open a `ggez` window and load some audio.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ toml = "0.5"
 log = "0.4"
 lyon = "0.13"
 smart-default = "0.5"
-nalgebra = {version = "0.18", features = ["mint"] }
+nalgebra = {version = "0.21", features = ["mint"] }
 # Has to be the same version of mint that nalgebra uses here.
 mint = "0.5"
 gilrs = "0.7"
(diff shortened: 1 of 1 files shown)


icefoxen:
Not sure about this one; `nalgebra` is still re-exported as `ggez::na` and so changing that is technically a breaking change.  See https://github.com/ggez/ggez/pull/722 as well.  On the other hand, removing `ggez::na` entirely is probably a good idea.  That said, this should get incorporated in a 0.6 release, see https://github.com/ggez/ggez/issues/697

icefoxen:
This is superceded by the latest `devel` branch, which does its best to remove `nalgebra` and `rand` completely.

fosskers:
Cool, thanks for the update.

PR #798, opened 2020-07-14
Remove cgmath dependency in example 02

### LINKED ISSUE #797: cgmath dependency in example 02
**Describe the bug**
the [example 02](https://github.com/ggez/ggez/blob/master/examples/02_hello_world.rs) needs the cgmath crate when it could simply use ggez's nalgebra.

**To Reproduce**
Run the copde without cgmath installed.

**Expected behavior**
Examples should be as painless as possible to run.

**Hardware and Software:**
 - ggez version: master

tlebrize:
changed cgmath's depency to use ggez's nalgebra instead. Fixes #797

### DIFF
--- a/examples/02_hello_world.rs
+++ b/examples/02_hello_world.rs
@@ -1,11 +1,11 @@
 //! Basic hello world example.
 
-use cgmath;
 use ggez;
 
 use ggez::event;
 use ggez::graphics;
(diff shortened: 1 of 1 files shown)


icefoxen:
Looks fine, but it's been taken over by more recent changes.  Thank you anyway!

PR #819, opened 2020-09-25
Replace gravity-sim-v2 on the projects list with SIMple Physics

mkhan45:
I renamed the repo so the link was broken, and I think it's better to link to the website anyway. Thanks!


icefoxen:
Great, thanks for the update!

PR #821, opened 2020-09-26
Add semeion to the list of projects

gliderkite:
All the examples of `semeion` are built with `ggez`, as well as the linked `formicarium` project.


icefoxen:
Awesome, thank you.

PR #822, opened 2020-10-01
Allow setting visibility of window

masterfool:
Allows the window to be shown or hidden. Useful for example to hide window on startup until initial resources have loaded.


icefoxen:
Nice, thank you.  Unfortunately changing the `WindowMode` struct is a breaking change, so this will have to get pulled into the next major release.

masterfool:
Makes sense. Have you considered changing structs to have all private fields so that additions like this would be minor and could be released on a shorter frequency?

PR #825, opened 2020-10-03
Get examples README displayed by default

mkesper:
By renaming 00_README it gets displayed by Github, improving user experience.

Also added a commit for reformatting the README for better editing and diffing.

Feel free to squash the commits etc.


icefoxen:
Hmmm.  The reason it was called 00_README was that when it was just called README lots of people missed that it was there.  But, it's a safe assumption people will be looking at it on github anyway, I wonder if showing the readme in non-root folders is a new feature.

Accepted, I just need to sort out some branching issues.

mkesper:
Thanks for taking care!

AaronM04:
I agree that the 00_ should be removed from the readme file name. If we want the example .rs files to appear after the readme maybe we could rename them as well. Either prepend a character or word that comes after capital letters in ASCIIbetical order, or just take off the numbers.

Issue #827, opened 2020-10-10
Accessing gfx_context or importing a texture from gstreamer


jabdoa2:
I'm maintaining a pinball framework which currently uses a python media controller. I want to convert that to rust with ggez. This turned out to be a great experience so far. Performance is much better than in python and it generally has been easy to work with ggez. Thanks for that :-).

To support video rendering we are using gstreamer. So far we have been using a standard pipeline which runs on the CPU (https://github.com/missionpinball/mpf-mc-rust/blob/dev/src/server.rs#L41). We then grab each frame, map it in the CPU and use `graphics::Image::from_rgba8` to upload and show it using ggez (https://github.com/missionpinball/mpf-mc-rust/blob/dev/src/scene.rs#L193). Works fine for most videos but it uses a lot of CPU which is an issue especially on embedded systems.

Instead we would like to move this work to the GPU. Render videos using gstreamer, then upload them via glupload to the GPU, grab the texture and draw it in ggez. Unfortunately, I fail to to that currently because I cannot convert my gstreamer texture to a gfx texture in ggez. Is there some way to access `gfx_context`? I found `raw_to_typed_texture` but it requires a backend which I also cannot get. I want to use something like this: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/examples/src/bin/glupload.rs but with ggez instead of with native gl commands. To share a gl context with gstreamer I also somehow need to get `gl_context`, `gl_display` and `platform`. Is that somehow possible?

icefoxen:
Good question.  https://docs.rs/ggez/0.5.1/ggez/graphics/fn.gfx_objects.html is probably the best path forward for "convert your texture to gfx texture", but it's still not great.  Other suggestions are welcome though!

jabdoa2:
I got everything working except instantiating a gfx texture and rendering it. I guess I could use an opengl context and draw the texture on my own but then I would also compile a shader and so on which feels wrong (and ggez also complains later in debug mode). Maybe I can transmute my integer texture id into a gfx raw texture somehow. Need to dig into that further.

icefoxen:
Not sure what the problem is, you should be able to create textures using the `gfx::Factory` type.

nobbele:
this shouldn't be an issue in 0.8

PR #828, opened 2020-10-11
allow getting raw window from ggez

jabdoa2:
partially enables #827


icefoxen:
Thank you!  Makes life simpler, at least.

PR #833, opened 2020-10-25
Add seeking support for ggez::filesystem::File

alula:


icefoxen:
Thank you!  I'm not 100% happy with the `vfs::VFile` `Seek` implementation for zip files, but... oh well.

PR #834, opened 2020-11-02
Upgrade glutin and winit

sapir:
Possibly I haven't fixed everything correctly, but at least it compiles and I can display some images with it.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -18,7 +18,7 @@ use winit::{self, dpi};
 // and look forward to ripping it all out and replacing it with newer winit.
 
 /// A mouse button.
-pub use winit::MouseButton;
+pub use winit::event::MouseButton;
 
 /// An analog axis of some device (gamepad thumbstick, joystick...).
 pub use gilrs::Axis;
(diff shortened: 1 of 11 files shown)


src/event.rs
```diff
@@ -277,9 +298,17 @@ where
                 }
             }
         }
-        state.update(ctx)?;
-        state.draw(ctx)?;
-    }
 
-    Ok(())
+        if let Err(e) = state.update(ctx) {
```
sapir, line 302: I think I should have moved this part into one of the event handlers, maybe MainEventsCleared?

icefoxen:
Dang!  I was thinking about doing this and was just about to conclude it wasn't worth the work, and yet here it is.  Thank you!

The downside is, well, needing updated versions of all the gfx libraries.  Vendoring all of `gfx-rs` is a bit extreme.  I wonder if they'd accept PR's and make new releases?  I'll test this out and see how it goes.

icefoxen:
Hmm.  The code works, at least from some basic tests.  If you want to continue on this could I ask you to fix the merge conflicts and do further work against the `devel` branch please?

icefoxen:
Ooooh, this looks like it might be a good path forward for this: https://crates.io/crates/old_school_gfx_glutin_ext

> imbris: We switched to it in veloren since we haven't finished wgpu transition yet

icefoxen:
Holy heck it works.  Thank you!  I've pulled this into  #837 and am carrying on development there.

sapir:
> The downside is, well, needing updated versions of all the gfx libraries. Vendoring all of gfx-rs is a bit extreme. I wonder if they'd accept PR's and make new releases? I'll test this out and see how it goes.

Well, I started to make a PR to gfx-rs and discovered that updating breaks more stuff there, like gfx-window-vulkan, similar to the breakage here, so I left it at that.

PR #842, opened 2020-11-22
[Chore] Fixes clippy warnings

TrveFenris:
This changeset aims to fix a few linting warnings/errors that `clippy` found while analyzing the whole repository. I did not address every warning since some of them might have side effects (like "redundant import" warnings, for example).

I'll be glad to add more changes (or remove some of the commits that go against code specs of the project) if it feels right, either as additional commits to this branch or as a brand new branch.

Please read each commit description to get more insight on the changes made.


src/graphics/mod.rs
```diff
@@ -290,10 +285,7 @@ impl BackendSpec for GlBackendSpec {
...
         depth_format: gfx::format::Format,
         window: &glutin::WindowedContext,
-    ) -> Option<(
-        gfx::handle::RawRenderTargetView<Self::Resources>,
-        gfx::handle::RawDepthStencilView<Self::Resources>,
-    )> {
+    ) -> MainTargetView<Self::Resources> {
         // Basically taken from the definition of
         // gfx_window_glutin::update_views()
         let dim = color_view.get_dimensions();
```
TrveFenris, line 291: If you check the `if let` statement below this line you'll notice that it can be removed and just call `update_views_raw`. Please let me know if this change is desirable.

examples/cube.rs
```diff
@@ -148,7 +148,7 @@ void main() {
             Vertex::new([1, -1, -1], [0, 1]),
         ];
 
-        #[cfg_attr(rustfmt, rustfmt_skip)]
+        #[rustfmt::skip]
```
TrveFenris, line 151: Stabilization of tool attributes is probably tied to a newer version of Rust. Let me know if it is a desirable change.

icefoxen:
Excellent, thank you so much!  Just make sure you work from and merge into the `devel` branch in the future.  :heart:

PR #851, opened 2020-12-08
Add Image::from_bytes constructor

17cupsofcoffee:
I was nerd-sniped on Discord, whoops--

Named to be consistent with `SoundData::from_bytes`, but I can change it if it's not clear enough.


icefoxen:
:D  Thank you

PR #852, opened 2020-12-13
Remove unused imports in examples/

Andy-Python-Programmer:
This pull request removes unused imports in `examples/` and removes statements like
```rust
use ggez;
```

As use statement with one path segment do nothing and extern crates is not needed in the latest versions of rust.


icefoxen:
Great, thanks!

Issue #853, opened 2020-12-15
Compile error on devel


jabdoa2:
**Describe the bug**
Compile error on `devel`:

```
warning: use of deprecated item 'winit::event::KeyboardInput::modifiers': Deprecated in favor of WindowEvent::ModifiersChanged
   --> src/context.rs:159:29
    |
159 | ...                   modifiers,
    |                       ^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated item 'winit::event::KeyboardInput::modifiers': Deprecated in favor of WindowEvent::ModifiersChanged
   --> src/event.rs:206:29
    |
206 | ...                   modifiers,
    |                       ^^^^^^^^^

warning: use of deprecated item 'winit::event::KeyboardInput::modifiers': Deprecated in favor of WindowEvent::ModifiersChanged
   --> src/event.rs:219:29
    |
219 | ...                   modifiers,
    |                       ^^^^^^^^^

error[E0599]: no method named `to_rgba8` found for enum `image::DynamicImage` in the current scope
   --> src/graphics/context.rs:344:24
    |
344 |     let image_data = i.to_rgba8();
    |                        ^^^^^^^^ help: there is an associated function with a similar name: `to_rgba`

error[E0599]: no method named `to_rgba8` found for enum `image::DynamicImage` in the current scope
   --> src/graphics/image.rs:152:44
    |
152 |             image::load_from_memory(&buf)?.to_rgba8()
    |                                            ^^^^^^^^ help: there is an associated function with a similar name: `to_rgba`

error: aborting due to 2 previous errors; 3 warnings emitted

For more information about this error, try `rustc --explain E0599`.
error: could not compile `ggez`.
```

**To Reproduce**
`cargo build`

**Expected behavior**
Compiles successfully.

**Hardware and Software:**
 - ggez version: devel
 - OS: Ubuntu 20.04

icefoxen:
Fixed in linked PR, thank you.

PR #854, opened 2020-12-15
Fix compile errors

jabdoa2:
Fix https://github.com/ggez/ggez/issues/853


icefoxen:
Thanks, not sure how this got missed.

PR #855, opened 2020-12-16
Fix keyboard documentation

danitrod:
I was wrapping my head around some errors running the documentation example until I noticed this small issue, decided to open a PR.


icefoxen:
Good catch, thank you

PR #856, opened 2020-12-17
Bump rustc version to 1.40

Andy-Python-Programmer:
This pull request bumps the minimum rustc version to 1.40 as `static_assertions` minimum version of rustc is 1.37 and `cast` requires rustc 1.38 and for `#[cfg(doctest)]` requires rustc 1.40. This should also make the `master` tree go green again!


icefoxen:
Awesome, thank you for your help!

Issue #861, opened 2020-12-25
Performance problem on devel (`update` and `draw` called twice per loop, probably)


AndrewRadev:
**Describe the bug**

On the current `devel` branch, I'm getting a slowdown. I think the problem is with #774, particularly this snippet: https://github.com/ggez/ggez/pull/774/files#diff-14da7c350de5742eaa248de110fefeae9b4b639c0358e26024d6040ccf174a41R309-R310. I think the `update` and `draw` functions are being called a second time in the loop.

On the latest `devel` commit there's a todo on top, so maybe it's a known problem, but I couldn't see a recent open issue about it (though I've only looked at the titles, sorry if this is a duplicate): https://github.com/ggez/ggez/blob/427661476d3d2133d74c7e7d73b6c6bc0a1cd40e/src/event.rs#L310-L311

Commenting those two lines out seems to fix my problem.

**To Reproduce**

I'm using the style of main loop shown in the examples, nothing fancier:

``` rust
const DESIRED_FPS: u32 = 60;

while timer::check_update_time(ctx, DESIRED_FPS)? {
    let seconds = 1.0 / (DESIRED_FPS as f32);
```

The project I'm testing with is a simple game for teaching purposes, tutorial-ish: https://github.com/AndrewRadev/rust-shooter/tree/update-ggez-2020. But I can reproduce it with astroblasto, too. When the cursor hovers on top of the window, the whole game freezes up.

**Expected behavior**

Tested on 5bbb5c37ea04eaa80f75480a2e977c0752c45d14 (before #774):

https://user-images.githubusercontent.com/124255/103132824-4d000480-46af-11eb-8d32-6b64060345d1.mp4

**Actual behaviour**

Tested on 1d9abaccae574e69b49693f03e269b980c210bd9 (after #774):

https://user-images.githubusercontent.com/124255/103132833-65701f00-46af-11eb-8943-2eacb8c24162.mp4

**Hardware and Software:**
 - ggez version: current git devel branch
 - OS: Linux (Arch), X11
 - Graphics card: Embedded Intel, also happens on Radeon RX 640, doesn't seem like it should matter if I'm right about the issue

nobbele:
that snippet is the problem, removing the last update/draw call fixes it without any issues. please fix upstream

icefoxen:
Fixed in #862, thank you.

AndrewRadev:
Awesome, thanks :+1:

PR #862, opened 2020-12-30
Update and draw only on MainEventsCleared.

0----0:
I think this was left over by some sort of merge mistake? d2a1fe9 moved
state.update() and state.draw() into the MainEventsCleared handler,
where they should be, but somehow they wound up outside again.  Anyway,
this fixes FPS issues in astroblasto on my machine.

The callback passed to event_loop.run() is called with every single event, so anything that's written after the match statement should also get executed with each event.  Code meant to run _after_ the event queue has finished processing should be in the MainEventsCleared handler, assuming I understand the way that the new winit event loop works.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -271,6 +271,25 @@ where
                 // internal state however necessary.
                 ctx.timer_context.tick();
 
+                // Handle gamepad events if necessary.
+                if ctx.conf.modules.gamepad {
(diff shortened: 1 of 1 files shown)


0----0:
Should fix #861

icefoxen:
Yep, it was left over from a bad/buggy merge.  Thank you so much for the PR, I only had time to sit down and look at this today.

PR #869, opened 2021-01-08
Invalidated text cached_metrics when returning a mutable slice of tex…

adam-bates:
…t fragments.


adam-bates:
Calling `text.width(ctx)` was giving cached values when I modified the content of a text fragment.

icefoxen:
Very good catch, thank you!  Doesn't look like it merges cleanly into the `devel` branch though, I just redid it myself in 856f71096e61ad7e53d7057dbe9e88e28d5bb183 .  :heart:

Issue #809, opened 2020-08-14
Creating and using a custom pipeline


arunchaganty:
For a game I'm building, I need to use two `sampler2D` textures in a shader (fragment shader code below); the TL;DR of this issue is a request for pointers on how to support such a use case.

As I understand it, this requires me to create a new gfx pipeline. I've been using [this post](https://wiki.alopex.li/LearningGfx) as a reference, and in an attempt to get something to work with minimal effort, I forked `ggez` and updated `ggez::graphics::pipe` to include an additional texture `t_Aux`, and accordingly updated `pipe::Data`. The code compiles, tests pass and shaders that do not use the new `t_Aux` work fine. However, when compiling a shader that does use the new `t_Aux` variables causes a panic:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: VideoError("Error constructing pipeline!\nThis should probably not be happening; it probably means an error in a shader or something.\nError was: DescriptorInit(ResourceView(\"t_Aux\", None))")', src/main.rs:60:21
```

A minimal set of edits that might be relevant is here, though the full version is probably easier to see [here](https://github.com/ggez/ggez/compare/master...arunchaganty:aux-texture?expand=1).

```rust
impl GraphicsContextGeneric<GlBackendSpec> {
    /// Create a new GraphicsContext
    pub(crate) fn new()
        // ...
        let typed_thingy_copy = backend.raw_to_typed_shader_resource(white_image.texture.clone());
        let sampler_copy = samplers.get_or_insert(sampler_info, &mut factory);

        let data = pipe::Data {
            vbuf: quad_vertex_buffer.clone(),
            tex: (typed_thingy, sampler),
            // Adding the aux texture data
            aux: (typed_thingy_copy, sampler_copy),
            rect_instance_properties: rect_inst_props,
            globals: globals_buffer,
            out: screen_render_target.clone(),
        };
       // ...
```

Accompanying fragment shader:

```glsl
/**
 * Advects flow 
 */
#version 150 core

/** The underlying texture we are advecting. **/
uniform sampler2D t_Texture;

/** The underlying texture we are advecting. **/
uniform sampler2D t_Aux;
/** Our coordinates **/
in vec2 v_Uv;
/** Final output **/
out vec4 Target0;

layout (std140) uniform ShaderState {
    vec4 u_Color;
    vec2 u_ScreenSize;
    vec2 u_Pos;
    uint u_Tick;
    float u_TimeDelta;
};

void main() {
[...]

PR #870, opened 2021-01-08
Remove the docs for pipe in graphics::pipe

Andy-Python-Programmer:
This pull request removes the docs of pipe in graphics::pipe and also makes pipe a `pub(crate)` not `pub`.

```rust
/// The pipe module is auto-generated by gfx_defines!. 
/// You shouldn't need to touch it, but alas we can't exclude it from cargo doc.
```
Fixes that


icefoxen:
Very nice!  Thanks a lot.

Andy-Python-Programmer:
No problem mate!

Andy-Python-Programmer:
#809

PR #871, opened 2021-01-08
Add rgb color constants

Andy-Python-Programmer:
I think if we are providing the color consts `BLACK` and `WHITE`. I think then it makes sense to provide the rbg color constants too.

### DIFF
--- a/src/graphics/types.rs
+++ b/src/graphics/types.rs
@@ -241,7 +241,7 @@ pub struct Color {
     pub a: f32,
 }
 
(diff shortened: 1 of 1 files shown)


icefoxen:
This just leads to a slippery slope where we provide *every* color.  There are crates for that: https://crates.io/crates/css-colors and https://crates.io/crates/chromatica .  :)

Andy-Python-Programmer:
Ah yes :)

PR #876, opened 2021-01-18
Glyph position

### LINKED ISSUE #738: Not enough font metrics data for `ggez::graphics::Text`.
Right now it seems impossible to draw text that has a cursor in it somewhere. All because `ggez::graphics::Text` only exposes the overall text size, but there's no way to get individual character positions (so there's no way to know where to draw a cursor or selection).

There could be something like:
```
text.with_each_char(start_pos, |index, ch, pos| { /* Whatever */ });
```

### LINKED ISSUE #849: Add a way to get the position of a specific fragment (or character)
Basically I need to do collision check with the text characters themselves to get links in text working, currently each character is a separate fragment so if it can't be done character-wise that's fine.
Related to #738 if it must be done manually.

nobbele:
Closes #849 and closes #738 with something like
```rust
self.glyph_positions(context)
    .iter()
    .for_each(|glyph_position| println!("There's a glyph at {:?}!", glyph_position));
```

### DIFF
--- a/src/graphics/text.rs
+++ b/src/graphics/text.rs
@@ -342,8 +342,7 @@ impl Text {
         let glyph_positions: Vec<mint::Point2<f32>> = {
             let varied_section = self.generate_varied_section(Point2::new(0.0, 0.0), None);
             use glyph_brush::GlyphCruncher;
-            gb
-                .glyphs(varied_section)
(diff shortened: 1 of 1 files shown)


icefoxen:
:+1:

Issue #879, opened 2021-01-22
DrawParam matrices are incorrect


PSteinhaus:
While implementing my BitmapFont ontop of SpriteBatch I stumbled upon this line in ```drawparam.rs```:
```
/// A [`DrawParam`](struct.DrawParam.html) that has been crunched down to a single matrix.
    fn to_na_matrix(&self) -> Matrix4 {
        // ...
        translate * offset * rotation * scale * offset_inverse     // this one
    }
```
And it baffled me, because it seemed wrong to me (though I use DrawParams in my own game and they seem to work just fine).
The problem is this:

The matrix should be calculated as:
```translate * rotation * scale * offset_inverse```

because this is exactly what we want to do:
- shift the sprite by an offset (called offset_inverse in the code)
- then scale it around that point
- rotate it around that point
- and finally move it to its destination

There simply is no need for multiplying offset onto the matrix after rotation.

Now this seems like a major problem, right? But why has noone ever noticed it, if it really is?
The reason is that most sprites are big and drawn inside of a large coordinate system, thereby making a shift of say 0.5 pixels practically unnoticable.

But if you use a smaller sprite and a small coordinate system as in the following example:
```
use ggez;

use ggez::event;
use ggez::graphics;
use ggez::{Context, GameResult};
use std::env;
use std::path;
use ggez::nalgebra::{Point2};
use ggez::graphics::{DrawParam, Image, Rect, FilterMode};

struct MainState {
    img: Image
}

impl MainState {
    fn new(ctx: &mut Context) -> GameResult<MainState> {
        let s = MainState { img: Image::new(ctx, "/testSprite.png").unwrap() };
        Ok(s)
    }
}

impl event::EventHandler for MainState {
    fn update(&mut self, _ctx: &mut Context) -> GameResult {
        Ok(())
    }

    fn draw(&mut self, ctx: &mut Context) -> GameResult {
        graphics::clear(ctx, [0.1, 0.2, 0.3, 1.0].into());

        let offset = Point2::new(0.5, 0.5);

        graphics::draw(ctx, &self.img, DrawParam::default().offset(offset))?;
        graphics::present(ctx)?;

        Ok(())
    }
}

pub fn main() -> GameResult {
    let resource_dir = if let Ok(manifest_dir) = env::var("CARGO_MANIFEST_DIR") {
        let mut path = path::PathBuf::from(manifest_dir);
        path.push("resources");
        path
    } else {
        path::PathBuf::from("./resources")
    };

[...]

Issue #736, opened 2020-01-16
Confusion about DrawParam `offset` field 


ischeinkman:
I'm working on a pixel art-ish game and using a `SpriteBatch` to render groups of sprites at a time. Sometimes these renders include transforms on the batch itself, such as flipping the batch horizontally via scaling it by `[-1.0, 1.0]` from the center of the image. According to the documentation for `DrawParam::offset`, this should be doable via `offset([0.5, 0.5])`, a la:

```rust
let params = DrawParam::new()
    .scale([-1.0, 1.0])
    .offset([0.5, 0.5]);
graphics::draw(&mut ctx, &batch, params)?;
```

 However, that does not work. Instead, I need to scale by the dimensions of the batch itself: 

```rust
let batch_dims = batch.dimensions(&mut ctx).unwrap_or_else(Rect::zero);
let params = DrawParam::new()
    .scale([-1.0, 1.0])
    .offset([0.5 * batch_dims.w, 0.5 * batch_dims.h]);
graphics::draw(&mut ctx, &batch, params)?;
```

Is this intended behavior? If so, can the documentation be updated to reflect that?

icefoxen:
Nope, that's not intended behavior.  Thanks for reporting it!

ischeinkman:
I actually ended up looking through the examples and found [this](https://github.com/ggez/ggez/blob/efdab624d346532f913bc8b9096dffacb3effbb5/examples/spritebatch.rs#L72), so I ended up assuming it was intended and just not yet documented. Interesting to hear that it isn't.

Plus, I also ended up looking through the `DrawParam` source and becoming confused as to what the `offset` parameter is even supposed to do. The documentation says that the transform matrix is supposed to be `translate * offset * rotation * scale * offset_inverse`, but this means that the value of the top-left corner ends up being drawn at `self.translate + (Matrix4::identity() - scale) * self.offset` instead of `self.translate` or `self.translate - scale * self.offset`, which would be expected. 

I included the notebook I used for the algebra and a simple example [here.](https://gist.github.com/ischeinkman/fd8979abd69a43803111f19d9eba288e)

What *is* the intended behavior of the `offset` parameter? Should I create a new, more general issue/PR for general `offset` quirks?

ischeinkman:
[...]

PR #880, opened 2021-01-22
fixed drawing transformation matrix

PSteinhaus:
This pull request fixes how the transformation matrix is calculated, making offset work as intended.
For more details and examples of the problematic behavior see #879.

### DIFF
--- a/examples/spritebatch.rs
+++ b/examples/spritebatch.rs
@@ -67,7 +67,7 @@ impl event::EventHandler for MainState {
                 ((time % cycle) as f32 / cycle as f32 * 6.28).sin().abs() * 2.0 + 1.0,
             ))
             .rotation((time % cycle) as f32 / cycle as f32 * 6.28)
-            .offset(Vec2::new(750.0, 750.0));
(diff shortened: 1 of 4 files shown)


icefoxen:
Thank you!  I'll check this out and try to make sure it doesn't break anything else.

PSteinhaus:
Testing it I found out that it breaks the spritebatch example a little.
It's still working, but it looks a little differently now. This is probably because the offset of a spritebatch currently works differently (being based not on a 1x1 size, but the size of the drawing surface that is), which amplyfies the bug which gets fixed in this PR. Therefore any countermeasures applied to the example to make it look nice are now no longer working as expected.

PSteinhaus:
While I was at it I also fixed #888 and #736. I hope that's ok.

Fixing #736 was a bit painful though. Now, when you use an offset with a SpriteBatch you'll have to pay a heavy performance cost, because SpriteBatch::dimensions needs to be calculated in every frame in this case. I made sure to include a check whether there actually is an offset, to safe most users from this cost.

icefoxen:
I'm still crossing my fingers merging this, but I can't find anything wrong with it yet...

Issue #865, opened 2021-01-05
Strange behavior when switching to true fullscreen without changing the drawable size


tesselode:
**Describe the bug**
If you switch to true fullscreen without changing the drawable size, the game appears as a borderless window in the top-left corner of the screen.

**To Reproduce**
Run this code and press F11:
```rust
use std::error::Error;

use ggez::{
	conf::FullscreenType,
	event::{EventHandler, KeyCode, KeyMods},
	graphics, Context, ContextBuilder, GameResult,
};
use graphics::{DrawMode, Mesh};

pub struct MainState {
	fullscreen: bool,
}

impl MainState {
	pub fn new() -> Self {
		Self { fullscreen: false }
	}
}

impl EventHandler for MainState {
	fn key_down_event(
		&mut self,
		ctx: &mut Context,
		keycode: KeyCode,
		_keymods: KeyMods,
		_repeat: bool,
	) {
		match keycode {
			KeyCode::F11 => {
				if self.fullscreen {
					graphics::set_fullscreen(ctx, FullscreenType::Windowed).unwrap();
				} else {
					graphics::set_fullscreen(ctx, FullscreenType::True).unwrap();
				}
				self.fullscreen = !self.fullscreen;
			}
			_ => {}
		}
	}

	fn update(&mut self, _ctx: &mut Context) -> GameResult {
		Ok(())
	}

	fn draw(&mut self, ctx: &mut Context) -> GameResult {
		graphics::clear(ctx, graphics::BLACK);
		let (drawable_width, drawable_height) = graphics::drawable_size(ctx);
		let circle = Mesh::new_circle(
			ctx,
			DrawMode::fill(),
			[drawable_width / 2.0, drawable_height / 2.0],
			drawable_height / 2.0,
			0.1,
			graphics::WHITE,
		)?;
		graphics::draw(ctx, &circle, ([0.0, 0.0],))?;
		graphics::present(ctx)?;
		Ok(())
	}
}

fn main() -> Result<(), Box<dyn Error>> {
	let (ctx, event_loop) = ContextBuilder::new("test", "test").build()?;
	let main_state = MainState::new();
	ggez::event::run(ctx, event_loop, main_state);
}
```

**Expected behavior**
The desktop resolution should switch to whatever the drawable size is, and the game should fill the entire screen (stretched out).

**Screenshots or pasted code**
Windowed mode:
[...]

Issue #883, opened 2021-01-29
(Bug) Mouse clicking/dragging adds an offset in 0.6.0-rc0


Manghi:
Props to @AaronM04 for first finding this bug.

**Describe the bug**
Mouse events are offset proportionally to the distance from the top-left corner of the window. The closer you are to (0,0) the less the error in offset is. The farther away from (0,0) in both positive directions increases the error.

**To Reproduce**
Run the `input_test` example on the head of the development branch. When the rectangle is dragged close to (0,0), there is virtually little-to-no offset error. The farther (+,+) you drag the rectangle to, the large the offset error is.

**Expected behavior**
The top-left corner of the rectangle follows the cursor on click/drag without an introduced offset.

**Screenshots or pasted code**
N/A but I was able to narrow down the offending commits these two back in Nov 2020:
```
The first bad commit could be any of:
599ab16bcd25e8db5a8c3fc8ab5d286cb348201e
6299b0b71662d642435968a83dd527dd0f02fdbc
We cannot bisect more!
```

Commit 599ab16bcd25e8db5a8c3fc8ab5d286cb348201e has build errors but looks like only a change in dependencies (`gfx_window_glutin`). 
Commit 6299b0b71662d642435968a83dd527dd0f02fdbc runs incredibly slow on my machine but it but does exhibit the issue in question. 

I scanned the code changes for the two commits but nothing stood out as an obvious introduction of the bug. The `input_test.rs` drawing code didn't change, and the mouse event handlers all look reasonable. I suspect it might be a dependency issue? Perhaps someone who is more familiar with how ggez/gfx interface might be able to solve this faster than me. (@sapir since he/she worked on these changes.)

**Hardware and Software:**
 - ggez version: Current git devel branch or tag `0.6.0-rc0`
 - OS: Linux Ubuntu 18.04
 - Graphics card: Nvidia GTX 1080
 - Graphics card drivers: 
`Kernel driver in use: nvidia`
`Kernel modules: nvidiafb, nouveau, nvidia_drm, nvidia`

PSteinhaus:
Thanks for the info!
I just tested it and I think I found the reason for this behavior.
It occurs when there is a mismatch between logical and physical screen size. For example setting scale from 125% to 100% made everything work as expected for me on Windows 10.

Thanks for narrowing it down. I'll try to look into this today.
Probably all that's necessary is changing some function to use physical size instead of logical size.

PSteinhaus:
[...]

PR #884, opened 2021-01-29
forced the window dimensions to be physical not logical

PSteinhaus:
The initial coordinate system was calculated based on the **logical** window size, while we actually want it to match the **physical** window size, so I changed it.
I also added some explanation to the input example, since users should be aware that they may need to think about such things.

inspired by / solving #883

### DIFF
--- a/examples/input_test.rs
+++ b/examples/input_test.rs
@@ -55,10 +55,6 @@ impl event::EventHandler for MainState {
         Ok(())
     }
 
-    fn resize_event(&mut self, _ctx: &mut Context, width: f32, height: f32)  {
-        println!("width: {}, height: {}", width, height);
-    }
-
     fn mouse_button_down_event(&mut self, _ctx: &mut Context, button: MouseButton, x: f32, y: f32) {
         self.mouse_down = true;
         println!("Mouse button pressed: {:?}, x: {}, y: {}", button, x, y);
--- a/examples/input_test.rs
+++ b/examples/input_test.rs
@@ -65,10 +65,15 @@ impl event::EventHandler for MainState {
         println!("Mouse button released: {:?}, x: {}, y: {}", button, x, y);
     }
 
-    fn mouse_motion_event(&mut self, _ctx: &mut Context, x: f32, y: f32, xrel: f32, yrel: f32) {
+    fn mouse_motion_event(&mut self, ctx: &mut Context, x: f32, y: f32, xrel: f32, yrel: f32) {
         if self.mouse_down {
-            self.pos_x = x;
-            self.pos_y = y;
+            // Mouse coordinates are PHYSICAL coordinates, but here we want logical coordinates
+            // so we need to divide through the scale factor.
+            // In more advanced/realistic scenarios you may need to calculate the logical coordinates
+            // using ggez::graphics::screen_coordinates.
+            let scale_f = graphics::window(ctx).scale_factor() as f32;
+            self.pos_x = x / scale_f;
+            self.pos_y = y / scale_f;
         }
         println!(
             "Mouse motion, x: {}, y: {}, relative x: {}, relative y: {}",
--- a/examples/input_test.rs
+++ b/examples/input_test.rs
@@ -55,6 +55,10 @@ impl event::EventHandler for MainState {
         Ok(())
     }
 
+    fn resize_event(&mut self, _ctx: &mut Context, width: f32, height: f32)  {
+        println!("width: {}, height: {}", width, height);
+    }
+
     fn mouse_button_down_event(&mut self, _ctx: &mut Context, button: MouseButton, x: f32, y: f32) {
         self.mouse_down = true;
         println!("Mouse button pressed: {:?}, x: {}, y: {}", button, x, y);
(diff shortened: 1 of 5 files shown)


PSteinhaus:
Hmm... having seen a bit more of the source code I'm now unsure, whether my fix is actually correct, as it seems that the WindowMode dimensions are actually supposed to be physical window dimensions and not logical ones (as they are now)?
Is that true?

icefoxen:
Oh look, a horrible coordinate bug caused by hidpi scaling.  Never seen that before.  :sigh:

The `WindowMode` dimensions *should* be physical dimensions and not logical ones, that is correct.  I really appreciate the help though.  Can I convince you to invert what this PR does?

PSteinhaus:
> Can I convince you to invert what this PR does?

Of course you can :)
I'll try to make sure it doesn't break anything, but I guess you'll still want to double check, since you probably know better where the horrible hidpi-bugs usually gather.

icefoxen:
Will do.  The unit tests don't all do anything useful currently, but *are* worth running.  Most of the real graphical tests basically consist of the example code, especially the more graphically complex ones like `shadows` and `text`.  I'll double-check everything I can before merging this though.

icefoxen:
:heart: 

Might be a day or two before I can test it out on a high-res display, but I'll try to get it done soon.

PSteinhaus:
Ok, I think I did it.

Changes:
- set_window_mode now divides all window dimensions through the scale factor, so that we get a window with the actual physical size we asked for (thereby practically disabling the scale factor)
- graphics::size and graphics::drawable_size now actually return physical sizes
- set_window_mode now also uses the monitor dimensions when toggling true fullscreen mode (fixing #865)

And I also left in the changes on input_test (but you can discard them if you think they just clutter it).

I tested most of the examples and everything seems to work just fine :)

What I'm unsure about is using the monitor dimensions when toggling true fullscreen mode. It makes perfect sense to me, but you may notice that I also changed the docs on WindowMode::dimensions, since the screen resolution in true fullscreen mode has now nothing to do with it anymore.
It fixes #865, but it also takes away the ability to set custom true fullscreen resolutions (which didn't work before anyway, so...)

Looking at winit::Window I don't see any way to change/specify the resolution, is there perhaps some other way?

PSteinhaus:
> Might be a day or two before I can test it out on a high-res display, but I'll try to get it done soon.

Actually it's quite easy to check, even on a low-res display (at least if you have Windows 10 running). Just go to screen options and set the scale to something else than 100% and you should be able to test just as well (but I'm also a creature of habit, so I understand if you want to do it the way that works best for you).

PSteinhaus:
By the way:
A more beautiful way to achieve what this PR does would be to simply change the scale factor of the window automatically to 1.

The thing is, my IDE tells me that there once was/ somewhere is a function in winit to do so, but as far as I can see it's just lying to my face... But if you happen to know/find one please let me know.

AaronM04:
@PSteinhaus After checking out your branch, I can confirm the `input_test` example works correctly for me. Nice work! 😁

PSteinhaus:
As a last refinement I changed true fullscreen behavior a bit:

Now setting true fullscreen results in getting the supported video modes of the current monitor. If the wanted resolution is found we now actually set the window to true fullscreen mode with the wanted resolution :)
But, this also means that setting to fullscreen can now fail and throw an error. I tried handling it as gracefully as I could though.

PSteinhaus:
Hmm... now the last thing I wonder about is whether we should add a function to ggez::conf which returns the supported resolutions of the current monitor, meaning something like:
```rust
pub fn supported_resolutions(ctx: &Context) -> impl Iterator<Item=winit::dpi::PhysicalSize<u32>> {
    let gfx = &ctx.gfx_context;
    let window = gfx.window.window();
    let monitor = window.current_monitor().unwrap();
    monitor.video_modes().map(|v_mode| v_mode.size())
}
```
Since the glutin window is already exposed through ```graphics::window```, we could also just let the users fend for themselves here. But I think providing a wrapper like this would make working with true fullscreen mode way easier for everyone.

icefoxen:
Hokay, I've tested this on a 4k laptop on Linux+X11 with NVidia drivers.  Without this patch, the following examples are broken: `canvas_subframe` `render_to_image` `shadows`.  With this patch, they all basically work, though the pointer location in the `shadows` example still doesn't quite match where it should be:

image

image

The yellow circle should always be centered on the pointer.  It appears to do this fine on an 1080p screen.

PSteinhaus:
Nice catch!
Luckily it was just a simple mistake in the example: It called ```graphics::size``` (giving it the window size including window borders), instead of ```graphics::drawable_size```.

PSteinhaus:
I added ```supported_resolutions```, since I think it's really nice to have.
If you think it would better be put somewhere else then ```ggez::conf```, or if you think it's unnecessary I can of course change things.

icefoxen:
`supported_resolutions()` function looks nice, I'd put it in the `graphics` module though.  That's where other query-graphics-settings things like `screen_coordinates` and `drawable_size` tend to live, so I think it makes sense there.

icefoxen:
Confirmed the bug in the `shadows` example is fixed, everything here looks good.  Do you consider this finished?  If so I'm ready to merge it.

PSteinhaus:
I'm absolutely done now.
Go ahead :)

PR #885, opened 2021-01-30
Update `Basic Project Template` for 0.6

Avishayy:
`event::run` now returns a `Never` type and doesn't get any references
therefore the old template doesn't compile.


icefoxen:
Thank you!

Issue #235, opened 2017-12-22
Road to 1.0


icefoxen:
For the purposes of fostering discussion.

When do we want to consider this 1.0?

Also, I'm intending to spend most of my time working on non-ggez things for 2018, so if there's features people seriously want added (mobile support, etc) I'm probably not going to be able to make them happen in the near future.  (One thing I do want to concentrate on is wasm support, but I'm hoping to do that by getting more into gfx-rs and various wasm tools.)

I'll still fix bugs and merge PR's and stuff, but I need to take a break from pushing things ahead.

Manghi:
Well since no one replied yet I though I would chime in.

* Resource management 
* Physics/collision engine
* Particle system
* ECS
* Scripting support
* Controller support (I know we have some already, not sure if it still needs work)
* Networking support
* Screen Management/UI 

The latter two are something we (myself and @AaronM04 ) are prioritizing so we have to roll our own. My goal is that anything we end up creating could be used by ggez, but to be frank I can't make any promises. 

The other items are things I know game engines tend to support. Up to team ggez if they fall under 1.0, or if they would be slated towards a future release. Since you're querying on an event-oriented release (what last things do we want before 1.0), I'll play devils advocate. 

Do you know when you all wanted 1.0 to be released?

icefoxen:
Good to see your list!  These are all things that are useful/necessary for a game but I also feel like are a layer above what I want to aim for with ggez.  I sort of want ggez to remain the low-level and simple-as-possible library so people can start with it and tinker with it fairly easily... not having to understand ECS concepts or a physics system or such.  What I personally **really want to do** with ggez is make it run on web, not add more functionality.

That said, a lot of those things are things I want myself, and some I've started implementing (very haphazardly) in the https://github.com/ggez/ggez-goodies repo.  I also have a template for starting a new game that I'm working on, which pulls together some of those things: resources with `warmy`, ECS with `specs`, scenes and cameras with `ggez-goodies`...  I've contemplated turning that into an actual engine, but haven't really gotten to that point.

I also have some semi-hacky GUI code in an incomplete game project that I should refactor out into its own thing someday, if people want.  It's not really designed to be general purpose though, and I, uh, don't actually know how to write GUI's...

[...]

PR #887, opened 2021-02-02
Physics

Andy-Python-Programmer:
**Warning**: Do not merge yet. Still working on it.

Context: https://github.com/ggez/ggez/issues/235#issuecomment-353841036
Also from me: Trying making easy to use physics with ggez. Got a lot of troubles with that. So making it super simple. Context for the overwhelming stuff: https://github.com/Andy-Python-Programmer/CallOfFerris/blob/master/src/screens/game/physics.rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,9 @@ default = ["c_dependencies"]
 bzip2 = ["zip/bzip2"]
 mp3 = ["rodio/mp3"]
 multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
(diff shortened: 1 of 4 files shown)


PSteinhaus:
I'm sorry, but in #235 it's stated, that physics belongs to the "[goodie stuff](https://github.com/ggez/ggez-goodies)" and is therefore not supposed to be part of the main engine.
Also [there are already many different working physics engines available](https://arewegameyet.rs/ecosystem/physics/), which can simply be imported and used in your game if you needs them.

Andy-Python-Programmer:
All good

PR #895, opened 2021-02-14
❌️ -> ✅️: Fix CI Failures due to Clippy Lints

Eosis:
I noticed that the CI is not passing due to Clippy, and thought I would try to get it green again. 😁

This will require a change to your Github Actions to use a slightly different clippy command (see below) so we don't pull in linting issues from dependencies, but when that is altered, I'm hoping the code changes I have made should allow it to pass.

I've made the changes pretty naïvely in order to get the lints passing (lots of `#[allow(...)]`). Perhaps you'd prefer to change the projects linting settings instead of this, but hopefully this is a good starting point. 🙂

---

The command to use in the Github Actions:
```
cargo clippy --all-targets --package ggez -- -D warnings
```

### DIFF
--- a/examples/01_super_simple.rs
+++ b/examples/01_super_simple.rs
@@ -1,4 +1,5 @@
 //! The simplest possible example that does something.
+#![allow(clippy::unnecessary_wraps)]
 
 use ggez::event;
 use ggez::graphics::{self, Color};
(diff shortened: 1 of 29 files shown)


icefoxen:
Cripes I need a newer version of clippy apparently.  Thank you so much!

PR #904, opened 2021-02-19
Run clippy just on ggez in github action

Eosis:
I realized that I hadn't actually changed the github action to only check ggez. 🤦‍♂️

I tested it out on my fork, so I'm more quite sure it will work this time around! 😅


PSteinhaus:
This looks good, merging it now.

Issue #694, opened 2019-10-20
The image displayed by the Canvas is corrupted


Athorus:
**Describe the bug**
The image displayed at the screen by the canvas is a little bit corrupted. When Image is displayed directly, everything is fine.

**To Reproduce**
See code below.
File [resources.zip](https://github.com/ggez/ggez/files/3747682/resources.zip) is provided too (contains 2 png).

**Expected behavior**
Canvas should display the same image as the one in input. No corrupted data is expected.

**Screenshots or pasted code**
If you look closely at the two first dark text lines, you can observe the second line is a little bit corrupted. There are black pixels around text.
With a light image (maybe because the contrast is higher), I cannot see any difference.
![canvas](https://user-images.githubusercontent.com/44225646/67157602-a1328a80-f32e-11e9-81d5-e9db35eb8268.png)

```rust
use ggez::ContextBuilder;
use ggez::conf::Conf;
use ggez::Context;
use ggez::event::{self, EventHandler};
use ggez::GameResult;
use ggez::graphics::{self, Canvas, Color, DrawParam, Image};

use mint::Point2;

struct Application {
    dark_image: Image,
    dark_canvas: Canvas,
    light_image: Image,
    light_canvas: Canvas,
    canvas_image: Image
}

impl Application {
    fn new(ctx: &mut Context) -> Self {
        let (dark_image, dark_canvas) = Self::image_canvas(ctx, "/dark-image.png");
        let (light_image, light_canvas) = Self::image_canvas(ctx, "/light-image.png");

        let canvas_image = Self::canvas(ctx, &dark_image).into_inner();

        Self {dark_image, dark_canvas, light_image, light_canvas, canvas_image}
    }

    fn image_canvas(ctx: &mut Context, file: &str) -> (Image, Canvas) {
        let image = Image::new(ctx, file).unwrap();
        let canvas = Self::canvas(ctx, &image);
        (image, canvas)
    }

    fn canvas(ctx: &mut Context, image: &Image) -> Canvas {
        let canvas = Canvas::with_window_size(ctx).unwrap();
        graphics::set_canvas(ctx, Some(&canvas));
        graphics::draw(ctx, image, DrawParam::new()).unwrap();
        graphics::set_canvas(ctx, None);

        canvas
    }
}

impl EventHandler for Application {
    fn update(&mut self, _: &mut Context) -> GameResult<()> {
        Ok(())
    }

    fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
[...]

PR #905, opened 2021-02-21
Add premultiplied alpha blend mode

### LINKED ISSUE #902: [0.6.0 rc1] egui integration
Hello, I'm trying to integrate `egui` into `ggez`, but I've run into some issues. Because the `egui::Mesh` doesn't guarantee the winding order of the vertices, it results in drawing text in differently sized rectangles.

I guess since the representation is so different, I'll have to use some lower-level things that I'm not sure `ggez` provides, since the `Context.gtx_context` is `pub(crate)`

You can take a look at my current, broken attempt here: https://github.com/Elinvynia/Enfaria/blob/ggez/enfaria-game/src/egui.rs

It results in a "Hello world!" string looking like this: 
image

What would be a better way of approaching this?

Elinvynia:
Note: I haven't been able to get any blend mode to work, so I'm very unsure if this does what I intend it to do. Regardless, setting the blend mode on either the `Image` or `Mesh`, in any combination, doesn't display any result at all.

This attempts to add the blend mode required to fix #902 

EDIT: Findings - setting the game mode globally via `ggez::graphics::set_blend_mode()` seems to work, so it's only the local  overrides that are broken. I guess that should go into a separate PR?

Anyway, now I have a way to test this, so it's time to play with values!

EDIT: Trying all combinations feels like shooting in the dark, I don't really understand what I'm doing, so any pointers would be helpul :(

### DIFF
--- a/src/graphics/shader.rs
+++ b/src/graphics/shader.rs
@@ -67,6 +67,8 @@ pub enum BlendMode {
     Lighten,
     /// When combining two fragments, choose the darker value
     Darken,
+    /// When using premultiplied alpha, use this.
+    Premultiplied,
 }
 
 impl From<BlendMode> for Blend {
@@ -113,6 +115,18 @@ impl From<BlendMode> for Blend {
                     destination: Factor::OneMinus(BlendValue::SourceAlpha),
                 },
             },
+            BlendMode::Premultiplied => Blend {
+                color: BlendChannel {
+                    equation: Equation::Add,
+                    source: Factor::One,
+                    destination: Factor::One,
+                },
+                alpha: BlendChannel {
+                    equation: Equation::Min,
+                    source: Factor::OneMinus(BlendValue::SourceAlpha),
+                    destination: Factor::OneMinus(BlendValue::SourceAlpha),
+                },
+            }
         }
     }
 }
--- a/src/graphics/shader.rs
+++ b/src/graphics/shader.rs
@@ -126,7 +126,7 @@ impl From<BlendMode> for Blend {
                     source: Factor::OneMinus(BlendValue::SourceAlpha),
                     destination: Factor::OneMinus(BlendValue::SourceAlpha),
                 },
-            }
+            },
         }
     }
 }
--- a/src/graphics/shader.rs
+++ b/src/graphics/shader.rs
@@ -122,8 +122,8 @@ impl From<BlendMode> for Blend {
                     destination: Factor::OneMinus(BlendValue::SourceAlpha),
                 },
                 alpha: BlendChannel {
-                    equation: Equation::Sub,
-                    source: Factor::One,
+                    equation: Equation::Add,
+                    source: Factor::OneMinus(BlendValue::DestinationAlpha),
                     destination: Factor::One,
                 },
             },
--- a/src/graphics/shader.rs
+++ b/src/graphics/shader.rs
@@ -123,7 +123,7 @@ impl From<BlendMode> for Blend {
                 },
                 alpha: BlendChannel {
                     equation: Equation::Add,
-                    source: Factor::OneMinus(BlendValue::DestinationAlpha),
+                    source: Factor::OneMinus(BlendValue::DestAlpha),
                     destination: Factor::One,
                 },
             },
--- a/src/graphics/shader.rs
+++ b/src/graphics/shader.rs
@@ -122,7 +122,7 @@ impl From<BlendMode> for Blend {
                     destination: Factor::One,
                 },
                 alpha: BlendChannel {
-                    equation: Equation::Min,
+                    equation: Equation::Add,
                     source: Factor::OneMinus(BlendValue::SourceAlpha),
                     destination: Factor::OneMinus(BlendValue::SourceAlpha),
                 },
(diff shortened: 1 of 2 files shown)


Elinvynia:
Some progress and more findings:
image

The `alpha` in `gfx::Blend` seems to have literally zero effect here, so it should all be about the `color`. Unfortunately I still haven't exactly figure what `(ONE, ONE_MINUS_SRC_ALPHA)` maps to, because there are three options - the equation (Min, Max, Sub, RevSub, Add) and then what maps to the `destination` and what to the `source`? If anyone knows, please let me know.

src/graphics/shader.rs
```diff
@@ -113,6 +115,18 @@ impl From<BlendMode> for Blend {
                     destination: Factor::OneMinus(BlendValue::SourceAlpha),
                 },
             },
+            BlendMode::Premultiplied => Blend {
+                color: BlendChannel {
+                    equation: Equation::Sub,
+                    source: Factor::One,
+                    destination: Factor::ZeroPlus(BlendValue::SourceAlpha),
```
emilk, line 122:
```suggestion
                    equation: Equation::Add,
                    source: Factor::One,
                    destination: Factor::OneMinus(BlendValue::SourceAlpha),
```

src/graphics/shader.rs
```diff
@@ -113,6 +115,18 @@ impl From<BlendMode> for Blend {
...
+                },
+                alpha: BlendChannel {
+                    equation: Equation::Sub,
+                    source: Factor::One,
+                    destination: Factor::One,
```
emilk, line 127: This matters much less, but technically this should be:

```suggestion
                    equation: Equation::Add,
                    source: Factor::OneMinus(BlendValue::DestinationAlpha),
                    destination: Factor::One,
```

Elinvynia:
That ends up looking like this: 
image

I think the issue must be in the way ggez handles blend modes, since only the global blend mode for everything works, and you can't set it per mesh/image (even though the docs say you can).-

PSteinhaus:
This effort seems like a good idea, but it's not functional yet, right?
If you could supply us with some demo code (your old links in #902 are broken) I'd try to look into it.

For now I'll convert this into a draft, ok?

PSteinhaus:
> only the global blend mode for everything works, and you can't set it per mesh/image (even though the docs say you can).-

That's strange, because for me setting blend modes per mesh/image works just fine... (try, for example, [this code](https://github.com/ggez/ggez/issues/694#issuecomment-850429958) with some slight canges (i.e. calling ```set_blend_mode``` on an image or canvas).

PSteinhaus:
@Elinvynia I think we might have stumbled upon a potential source of problems for you in #694 [here](https://github.com/ggez/ggez/issues/694#issuecomment-853320815).
Setting the blend mode on specific entities seems to work, but drawing on a transparent canvas using the ```Alpha``` blend mode will most certainly not yield the results you'd expect.

src/graphics/shader.rs
PSteinhaus: Why exactly does this have to be ```(ONE_MINUS_DEST_ALPHA, ONE)``` instead of just ```(ONE, ONE)```, like how usual Alpha is implemented:
```rust
pub const ALPHA: Blend = Blend {
        color: BlendChannel {
            equation: Equation::Add,
        },
        alpha: BlendChannel {
            equation: Equation::Add,
            source: Factor::One,
            destination: Factor::One,
        },
    };
```
I see why one might use ```(ONE_MINUS_DEST_ALPHA, ONE)``` instead of ```(ONE, ONE)```, since this is how transparency works in the real world. When you lay something that's half transparent (meaning half the light gets through) over something else that's half transparent then you'll get something where 75% of light passes, instead of something completely intransparent.
But I'm not yet fully convinced... wouldn't we need to change the definition of [ALPHA](https://docs.rs/draw_state/0.8.0/draw_state/preset/blend/constant.ALPHA.html) if this was true?

Thanks for helping btw.

src/graphics/shader.rs
```diff
@@ -113,6 +115,18 @@ impl From<BlendMode> for Blend {
                     destination: Factor::OneMinus(BlendValue::SourceAlpha),
                 },
             },
+            BlendMode::Premultiplied => Blend {
+                color: BlendChannel {
+                    equation: Equation::Add,
+                    source: Factor::One,
```
IGBC, line 121: Still wrapping my head around it but I think these two lines are still wrong. I think the equation should be ` col_target = col_src * a_src + col_dest * a_dest`

IGBC (changes requested):

src/graphics/shader.rs
PSteinhaus: No, it really shouldn't be. Just imagine having a white opaque destination.
If you used this equation then this destination would never change, it would always stay white, no matter what you draw upon it.

IGBC: OK I believe you

PSteinhaus: I think it's already right the way it is.
The new color value has to be the source color value plus whatever is left from the background/destination based upon how strong the alpha of the source is, meaning how much it covers up the background.

PR #912, opened 2021-03-11
Fix typo

cerebralcourt:
Missed a comma on line 110, resulting in error when trying to install from git.


dchiquito:
Clippy was complaining about quite a lot more stuff. I think this should be resolved by #917

cerebralcourt:
That's good, I came across those shortly after submitting this but didn't take the time to solve them.

PR #914, opened 2021-03-15
Replace to_rgba8 with to_rgba

Palladinium:
In newer versions of `image` (tested with `0.23.11`), `DynamicImage::to_rgba8()` has been removed and replaced with `to_rgba()`. This replaces two calls to it that cause compile errors.


Palladinium:
This actually seems to be a problem on `image`'s side, as updating it to `0.23.14` fixes this.

Issue #747, opened 2020-02-05
Behavior of ggez::graphics::Canvas when DrawParam::src is changed makes me confused


atsisy:
Hi. I'm developing a own game using ggez.
I found strange behavior of ggez::graphics::Canvas.

When ggez::graphics::Image is drawn with following code, drawing result will be like following.

```rust
let draw_param = ggez::graphics::DrawParam::default()
                                .src(ggez::graphics::Rect::new(0.0, 0.0, 1.0, 0.7));
ggez::graphics::draw(ctx, image_stuff, draw_param);

"Drawing Result of ggez::grapchics::Image"
---------------
|             |
|   visible   |
|-------------|
|  unvisible  |
---------------
```

But when ggez::graphics::Canvas is drawn with following code, drawing result will be like following.

```rust
let draw_param = ggez::graphics::DrawParam::default()
                                .src(ggez::graphics::Rect::new(0.0, 0.0, 1.0, 0.7));
ggez::graphics::draw(ctx, canvas_stuff, draw_param);

"Drawing Result of ggez::grapchics::Canvas"
---------------
|  unvisible  |
|-------------|
|   visible   |
|             |
---------------
```

This difference caused by flipping image when ggez::graphics::Canvas is drawn.
And I know the flipping is essential process.
But I want to draw ggez::graphics::Canvas same as ggez::graphics::Image.
So, do you know how to solve this problem?

Thank you.

icefoxen:
That's a good question.  Fixing this sort of inconsistency is one of the main problems with `Canvas`, but it involves fixing half-done hacks in a few different places, so right now fixing it for one case will just break it somewhere else.  :-(

Now that I think of it, the *right* answer might honestly be to copy a the `Canvas` to a new texture with the same byte ordering as `Image`...  Let me look into that.

PSteinhaus:
Hmm... I just tried fiddling with this a bit and to be honest I don't understand why we need to flip the image...
I tried how things look when we don't do it and therefore I see that it is necessary. But I don't see **why**...

Could you please try to explain that to me? I'd really like to give fixing this a try.

I guess the reason is that the ```RawRenderTargetView``` of a canvas is somehow different from the ```RawRenderTargetView``` of the screen, but... why? Can't we change that?

Btw: Not only DrawParam::src doesn't work because of this issue, but also DrawParam::rotation and DrawParam::offset...

nobbele:
fixed by #936

PR #936, opened 2021-05-21
Fixes canvas being upside down

### LINKED ISSUE #693: Canvas::into_inner(self)->Image returns a flipped image
**Describe the bug**
When Canvas::into_inner(self)->Image is called, a flipped image is returned.

**To Reproduce**
See code below. 
[resources.zip](https://github.com/ggez/ggez/files/3747672/resources.zip) files is provided too (contains 2 png).

**Expected behavior**
The same image as the one displayed should be returned.

The text at the bottom of the window is flipped. It should be the same as the text at the beginning of the window.
canvas

```rust
use ggez::ContextBuilder;
use ggez::conf::Conf;
use ggez::Context;
use ggez::event::{self, EventHandler};
use ggez::GameResult;
use ggez::graphics::{self, Canvas, Color, DrawParam, Image};

use mint::Point2;

struct Application {
    dark_image: Image,
    dark_canvas: Canvas,
    light_image: Image,
    light_canvas: Canvas,
    canvas_image: Image
}

impl Application {
    fn new(ctx: &mut Context) -> Self {
        let (dark_image, dark_canvas) = Self::image_canvas(ctx, "/dark-image.png");
        let (light_image, light_canvas) = Self::image_canvas(ctx, "/light-image.png");

        let canvas_image = Self::canvas(ctx, &dark_image).into_inner();

        Self {dark_image, dark_canvas, light_image, light_canvas, canvas_image}
    }

    fn image_canvas(ctx: &mut Context, file: &str) -> (Image, Canvas) {
        let image = Image::new(ctx, file).unwrap();
        let canvas = Self::canvas(ctx, &image);
        (image, canvas)
    }

    fn canvas(ctx: &mut Context, image: &Image) -> Canvas {
        let canvas = Canvas::with_window_size(ctx).unwrap();
        graphics::set_canvas(ctx, Some(&canvas));
        graphics::draw(ctx, image, DrawParam::new()).unwrap();
        graphics::set_canvas(ctx, None);

        canvas
    }
}

impl EventHandler for Application {
    fn update(&mut self, _: &mut Context) -> GameResult<()> {
        Ok(())
    }

        graphics::clear(ctx, Color{r: 0., g: 0.3, b: 0., a: 1.});

        graphics::draw(ctx, &self.dark_image, DrawParam::new()).unwrap();
        let point2 = Point2 {x: 0., y: 60.};
        graphics::draw(ctx, &self.dark_canvas, DrawParam::new().dest(point2)).unwrap();

        let point2 = Point2 {x: 0., y: 200.};
        graphics::draw(ctx, &self.light_image, DrawParam::new().dest(point2)).unwrap();
        let point2 = Point2 {x: 0., y: 260.};
        graphics::draw(ctx, &self.light_canvas, DrawParam::new().dest(point2)).unwrap();

        graphics::draw(ctx, &self.canvas_image, DrawParam::new()).unwrap();

        graphics::present(ctx).unwrap();
        Ok(())
    }
}

pub fn main() {
    let c = Conf::new();
    let (ctx, event_loop) = &mut ContextBuilder::new("canvas", "athorus").conf(c).build().unwrap();
    let app = &mut Application::new(ctx);
    event::run(ctx, event_loop, app).unwrap();
}
```

**Hardware and Software:**
 - ggez version: 0.5.1
 - OS: Windows 10
 - Graphics card: AMD Radeon(TM) R9 290X
 - Graphics card drivers: Radeon 19.9.2

nobbele:
This seems too obvious considering icefoxen marked it as HARD, does it not work as it should on some platforms? It seems to work for me.
closes #693
tested using code below modified from #693
```rust
use ggez::event::{self, EventHandler};
use ggez::graphics::{self, Canvas, Color, DrawParam, Image};
use ggez::Context;
use ggez::ContextBuilder;
use ggez::GameResult;
use ggez::{conf::Conf, graphics::Drawable};
use glam::vec2;

struct Application {
    dark_image: Image,
    dark_canvas: Canvas,
    canvas_image: Image,
}

impl Application {
    fn new(ctx: &mut Context) -> Self {
        let (dark_image, dark_canvas) = Self::image_canvas(ctx, "/download.png");

        let canvas_image = Self::canvas(ctx, &dark_image).into_inner();

        Self {
            dark_image,
            dark_canvas,
            canvas_image,
        }
    }

    fn image_canvas(ctx: &mut Context, file: &str) -> (Image, Canvas) {
        let image = Image::new(ctx, file).unwrap();
        let canvas = Self::canvas(ctx, &image);
        (image, canvas)
    }

    fn canvas(ctx: &mut Context, image: &Image) -> Canvas {
        let canvas = Canvas::with_window_size(ctx).unwrap();
        graphics::set_canvas(ctx, Some(&canvas));
        graphics::draw(ctx, image, DrawParam::new()).unwrap();
        graphics::set_canvas(ctx, None);

        canvas
    }
}

impl EventHandler for Application {
    fn update(&mut self, _: &mut Context) -> GameResult<()> {
        Ok(())
    }

        graphics::clear(
            ctx,
            Color {
                r: 0.,
                g: 0.3,
                b: 0.,
                a: 1.,
            },
        );

        graphics::draw(ctx, &self.dark_image, DrawParam::new()).unwrap();
        graphics::draw(ctx, &self.canvas_image, DrawParam::new().dest([400.0, 0.0])).unwrap();

        graphics::present(ctx).unwrap();
        Ok(())
    }
}

pub fn main() {
    let (mut ctx, event_loop) = ContextBuilder::new("canvas", "athorus").build().unwrap();
    let app = Application::new(&mut ctx);
    event::run(ctx, event_loop, app);
}
```

### DIFF
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -151,23 +151,7 @@ where
 impl Drawable for Canvas {
     fn draw(&self, ctx: &mut Context, param: DrawParam) -> GameResult {
         self.debug_id.assert(ctx);
-        // Gotta flip the image on the Y axis here
-        // to account for OpenGL's origin being at the bottom-left.
-        match param.trans {
-            Transform::Values { scale, dest, .. } => {
-                let new_scale = mint::Vector2 {
-                    x: scale.x,
-                    y: -scale.y,
-                };
-                let new_dest = mint::Point2 {
-                    x: dest.x,
-                    y: dest.y + f32::from(self.image.height()) * scale.y,
-                };
-                let new_param = param.dest(new_dest).scale(new_scale);
-                self.image.draw(ctx, new_param)
-            }
-            Transform::Matrix(_) => self.image.draw(ctx, param),
-        }
+        self.image.draw(ctx, param)
     }
     fn dimensions(&self, _: &mut Context) -> Option<Rect> {
         Some(self.image.dimensions())
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -207,25 +207,15 @@ impl Drawable for Canvas {
 
         // We have to mess with the scale to make everything
         // be its-unit-size-in-pixels.
-        let scale_x = param.src.w * f32::from(self.width());
-        let scale_y = param.src.h * f32::from(self.height());
+        let scale_x = param.src.w * self.width() as f32;
+        let scale_y = param.src.h * self.height() as f32;
 
-        let mat = glam::Mat4::from(param.trans.to_bare_matrix());
-        let new_param = param.transform(
-            mat * Matrix4::from_scale(glam::vec3(scale_x, scale_y, 1.0))
-                * glam::Mat4::from_scale_rotation_translation(
-                    glam::vec3(1.0, -1.0, 1.0),
-                    Quat::identity(),
-                    glam::vec3(0.0, 1.0, 0.0),
-                ),
+        let param = param.transform(
+            glam::Mat4::from(param.trans.to_bare_matrix())
+                * Matrix4::from_scale(glam::vec3(scale_x, scale_y, 1.0)),
         );
-        let new_src = Rect {
-            x: param.src.x,
-            y: (1.0 - param.src.h) - param.src.y,
-            w: param.src.w,
-            h: param.src.h,
-        };
-        let new_param = new_param.src(new_src);
+
+        let new_param = flip_draw_param_vertical(param);
 
         image::draw_image_raw(&self.image, ctx, new_param)
     }
@@ -240,44 +230,26 @@ impl Drawable for Canvas {
     }
 }
 
-/// Fast non-allocating function for flipping pixel data in an image vertically
-fn flip_pixel_data(rgba: &mut Vec<u8>, width: usize, height: usize) {
-    // cast the buffer into u32 so we can easily access the pixels themselves
-    // splits the pixel buffer into an upper (first) and a lower (second) half
-    let pixels: (&mut [u32], &mut [u32]) =
-        bytemuck::cast_slice_mut(rgba.as_mut_slice()).split_at_mut(width * height / 2);
-
-    // When the image has an uneven height, it will split the buffer in the middle of a row.
-    // This will decrease pixel count so that the x,y in the loop will never enter the split row since
-    // for uneven height images the middle row will stay the same anyway
-    let pixel_count = if height % 2 == 0 {
-        width * height / 2
+fn flip_draw_param_vertical(param: DrawParam) -> DrawParam {
+    let param = if let Transform::Matrix(mat) = param.trans {
+        param.transform(
+            glam::Mat4::from(mat)
+                * glam::Mat4::from_scale_rotation_translation(
+                    glam::vec3(1.0, -1.0, 1.0),
+                    Quat::identity(),
+                    glam::vec3(0.0, 1.0, 0.0),
+                ),
+        )
     } else {
-        width * height / 2 - width / 2
+        panic!("Can not be called with a non-matrix DrawParam");
     };
-    // Even though we removed uwidth / 2 from pixel_count,
-    // the second half of the buffer's size will still contain that data so
-    // we need to offset the index on that by the size of said data
-    let second_set_offset = if height % 2 == 0 {
-        // even height (evenness on width doesn't matter)
-        0
-    } else if width % 2 == 0 {
-        // uneven height but even width
-        width / 2
-    } else {
-        // uneven height and uneven width
-        width / 2 + 1
+    let new_src = Rect {
+        x: param.src.x,
+        y: (1.0 - param.src.h) - param.src.y,
+        w: param.src.w,
+        h: param.src.h,
     };
-    for i in 0..pixel_count {
-        let x = i % width;
-        let y = i / width;
-        let reverse_y = height / 2 - y - 1;
-
-        let idx = (y * width) + x;
-        let second_idx = (reverse_y * width) + x + second_set_offset;
-
-        std::mem::swap(&mut pixels.0[idx], &mut pixels.1[second_idx]);
-    }
+    param.src(new_src)
 }
 
 /// Set the `Canvas` to render to. Specifying `Option::None` will cause all
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -207,8 +207,8 @@ impl Drawable for Canvas {
 
         // We have to mess with the scale to make everything
         // be its-unit-size-in-pixels.
-        let scale_x = param.src.w * self.width() as f32;
-        let scale_y = param.src.h * self.height() as f32;
+        let scale_x = param.src.w * f32::from(self.width());
+        let scale_y = param.src.h * f32::from(self.height());
 
         let param = param.transform(
             glam::Mat4::from(param.trans.to_bare_matrix())
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -1,16 +1,19 @@
 //! I guess these docs will never appear since we re-export the canvas
 //! module from graphics...
+use std::path;
+
 use gfx::format::{Format, Swizzle};
 use gfx::handle::RawRenderTargetView;
 use gfx::memory::{Bind, Usage};
 use gfx::texture::{AaMode, Kind};
 use gfx::Factory;
+use glam::Quat;
 
-use crate::conf;
 use crate::context::DebugId;
 use crate::error::*;
 use crate::graphics::*;
 use crate::Context;
+use crate::{conf, filesystem};
 
 /// A generic canvas independent of graphics backend. This type should
 /// never need to be used directly; use [`graphics::Canvas`](type.Canvas.html)
@@ -117,9 +120,15 @@ where
         )
     }
 
-    /// Gets the backend `Image` that is being rendered to.
-    pub fn image(&self) -> &Image {
+    /// Gets the backend `Image` that is being rendered to. Note that this will be flipped but otherwise the same, use the [`to_image`](#method.to_image) function for the unflipped version.
+    pub fn raw_image(&self) -> &Image {
         &self.image
+    }
+
+    /// Creates an `Image` with the content of its raw counterpart but transformed to behave like a normal `Image`.
+    pub fn to_image(&self, ctx: &mut Context) -> GameResult<Image> {
+        let pixel_data = self.to_rgba8(ctx)?;
+        Image::from_rgba8(ctx, self.image.width, self.image.height, &pixel_data)
     }
 
     /// Gets the backend `Target` that is being rendered to.
@@ -127,7 +136,61 @@ where
         &self.target
     }
 
-    /// Get the filter mode for the image.
+    /// Dumps the flipped `Canvas`'s data to a `Vec` of `u8` RBGA8 values.
+    pub fn to_rgba8(&self, ctx: &mut Context) -> GameResult<Vec<u8>> {
+        let mut pixel_data = self.image.to_rgba8(ctx)?;
+        flip_pixel_data(
+            &mut pixel_data,
+            self.image.width as usize,
+            self.image.height as usize,
+        );
+        Ok(pixel_data)
+    }
+
+    /// Encode the `Canvas`'s content to the given file format and
+    /// write it out to the given path.
+    ///
+    /// See the [`filesystem`](../filesystem/index.html) module docs for where exactly
+    /// the file will end up.
+    pub fn encode<P: AsRef<path::Path>>(
+        &self,
+        ctx: &mut Context,
+        format: ImageFormat,
+        path: P,
+    ) -> GameResult {
+        use std::io;
+        let data = self.to_rgba8(ctx)?;
+        let f = filesystem::create(ctx, path)?;
+        let writer = &mut io::BufWriter::new(f);
+        let color_format = ::image::ColorType::Rgba8;
+        match format {
+            ImageFormat::Png => ::image::png::PngEncoder::new(writer)
+                .encode(
+                    &data,
+                    u32::from(self.width()),
+                    u32::from(self.height()),
+                    color_format,
+                )
+                .map_err(Into::into),
+        }
+    }
+
+    /// Return the width of the canvas.
+    pub fn width(&self) -> u16 {
+        self.image.width
+    }
+
+    /// Return the height of the canvas.
+    pub fn height(&self) -> u16 {
+        self.image.height
+    }
+
+    /// Returns the dimensions of the canvas.
+    pub fn dimensions(&self) -> Rect {
+        Rect::new(0.0, 0.0, f32::from(self.width()), f32::from(self.height()))
+    }
+
+    /// Get the filter mode for the canvas.
     pub fn filter(&self) -> FilterMode {
         self.image.filter()
     }
@@ -136,22 +199,35 @@ where
     pub fn set_filter(&mut self, mode: FilterMode) {
         self.image.set_filter(mode)
     }
-
-    /// Destroys the `Canvas` and returns the `Image` it contains.
-    pub fn into_inner(self) -> Image {
-        // TODO: This texture is created with different settings
-        // than the default; does that matter?
-        // Test; we really just need to add Bind::TRANSFER_SRC
-        // and change the Usage's to match to make them identical.
-        // Ask termhn maybe?
-        self.image
-    }
 }
 
 impl Drawable for Canvas {
     fn draw(&self, ctx: &mut Context, param: DrawParam) -> GameResult {
         self.debug_id.assert(ctx);
-        self.image.draw(ctx, param)
+
+        // We have to mess with the scale to make everything
+        // be its-unit-size-in-pixels.
+        let scale_x = param.src.w * f32::from(self.width());
+        let scale_y = param.src.h * f32::from(self.height());
+
+        let mat = glam::Mat4::from(param.trans.to_bare_matrix());
+        let new_param = param.transform(
+            mat * Matrix4::from_scale(glam::vec3(scale_x, scale_y, 1.0))
+                * glam::Mat4::from_scale_rotation_translation(
+                    glam::vec3(1.0, -1.0, 1.0),
+                    Quat::identity(),
+                    glam::vec3(0.0, 1.0, 0.0),
+                ),
+        );
+        let new_src = Rect {
+            x: param.src.x,
+            y: (1.0 - param.src.h) - param.src.y,
+            w: param.src.w,
+            h: param.src.h,
+        };
+        let new_param = new_param.src(new_src);
+
+        image::draw_image_raw(&self.image, ctx, new_param)
     }
     fn dimensions(&self, _: &mut Context) -> Option<Rect> {
         Some(self.image.dimensions())
@@ -161,6 +237,46 @@ impl Drawable for Canvas {
     }
     fn blend_mode(&self) -> Option<BlendMode> {
         self.image.blend_mode
+    }
+}
+
+/// Fast non-allocating function for flipping pixel data in an image vertically
+fn flip_pixel_data(rgba: &mut Vec<u8>, width: usize, height: usize) {
+    // cast the buffer into u32 so we can easily access the pixels themselves
+    // splits the pixel buffer into an upper (first) and a lower (second) half
+    let pixels: (&mut [u32], &mut [u32]) =
+        bytemuck::cast_slice_mut(rgba.as_mut_slice()).split_at_mut(width * height / 2);
+
+    // When the image has an uneven height, it will split the buffer in the middle of a row.
+    // This will decrease pixel count so that the x,y in the loop will never enter the split row since
+    // for uneven height images the middle row will stay the same anyway
+    let pixel_count = if height % 2 == 0 {
+        width * height / 2
+    } else {
+        width * height / 2 - width / 2
+    };
+    // Even though we removed uwidth / 2 from pixel_count,
+    // the second half of the buffer's size will still contain that data so
+    // we need to offset the index on that by the size of said data
+    let second_set_offset = if height % 2 == 0 {
+        // even height (evenness on width doesn't matter)
+        0
+    } else if width % 2 == 0 {
+        // uneven height but even width
+        width / 2
+    } else {
+        // uneven height and uneven width
+        width / 2 + 1
+    };
+    for i in 0..pixel_count {
+        let x = i % width;
+        let y = i / width;
+        let reverse_y = height / 2 - y - 1;
+
+        let idx = (y * width) + x;
+        let second_idx = (reverse_y * width) + x + second_set_offset;
+
+        std::mem::swap(&mut pixels.0[idx], &mut pixels.1[second_idx]);
     }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ glam = { version = "0.12", features = ["mint"]}
 mint = "0.5"
 gilrs = "0.8"
 approx = "0.4"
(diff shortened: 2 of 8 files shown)


nobbele:
@PSteinhaus can you test this?

PSteinhaus:
Will do! Looking at your changes it would at least make sense to me that it _could_ fix it.
But spontaneously I do wonder, because while this might give us the effect we want in canvases it will also flip all other images, won't it? I'll check it right away.

PSteinhaus:
Just checked, sadly it really wasn't as easy as that. All other images are flipped on the y axis now. Run the astroblasto example, for an example of this.

nobbele:
ooh it's image camvas, I know how to fix this then maybe

nobbele:
Now it should work

nobbele:
While testing this I found out that if you call .encode on an image coming from a canvas it's black? It appears to occur in normal devel branch too so it seems to be an existing issue

PSteinhaus:
I checked everything and it seems to just work now...
And it makes sense that it works... strange indeed. If that was all that it took to finally fix this issue I'd be absolutely delighted.

.encode indeed returns a black image...
We currently do not use it anywhere EXCEPT in astroblasto.rs (and some tests that currently do not run). There .encode seems to work (though it is used on screenshots). So .encode seems to work.

The issue might stem from the way canvas base images are created in Canvas::new. There is a comment in Canvas::into_inner hinting at it. We might need to do some fiddling around to solve this, or apply some amount of research to find out how Canvas::new actually works, preferably both...

We probably should start a new issue for this and merge this PR for now, since, as you said, it seems to be an existing issue.

PSteinhaus:
Though wait!
I just remembered that last time I looked at this (or #747 to be precise) there were also problems with DrawParam::src DrawParam::rotation and DrawParam::offset when used on a canvas. I vaguely remember that this kind of flipping stuff could be testet pretty well by drawing some test-meshes on a canvas. I'll try to run some tests on that later today.

nobbele:
flipping the image using the old method incurred a ~25% performance penalty but using this new method, it seems to barely affect the performance at all so that with this PR merged, Image will be just as fast as before

PSteinhaus:
Good idea. Yes, please wait a second. I'll get right to it.

PSteinhaus:
Ok, I tried out everything I wanted to try.

My initial fears did not manifest. The problems at play in #747 still persist, but they didn't get worse either. Actually they improved (but at a cost I'll come to in a moment). Also, drawing meshes works just as one would expect.

But then, when I thought everything was fine I ran the examples once more, side by side (devel vs PR), and I found this:
n_devel

which is very much different to this (devel):

devel

There are some off-by-one errors in the tile images that can probably be fixed by finding and fixing the corresponding off-by-one-error in ```flip_pixel_data```, but the much more problematic thing is that the approach of flipping all images and accounting for that by changing the draw function to flip the scale and move them, interferes with how DrawParam::offset and rotation work. In fact by doing this we have just moved all the nasty DrawParam problems from canvas to Image.

A similiar problem can be observed in transforms.rs. There it isn't the rotation, but the offset that causes the problem.

We could fix the offset problem by translating the image by ```f32::from(self.height()) * scale.y * (0.5 - offset.y) * 2.``` instead of the full height, but the problem of the rotation remains. And solving this that way seems to be much harder to me, since it would make it necessary to calculate where the bottom left of the Rect will be rotated to using trigonometric functions, just so that we could add another translation to account for that.

And at last I also want to mention that, though it seems to be currently the case that all images are somehow created by ```from_rgba8```, embracing this approach of flipping all images would mean that we'd have to be extra careful in the future to also apply this kind of flipping to all images stemming from possible other sources. It would thereby introduce a new convention that we'd have to remind ourselfs off.

nobbele:
> And at last I also want to mention that, though it seems to be currently the case that all images are somehow created by from_rgba8, embracing this approach of flipping all images would mean that we'd have to be extra careful in the future to also apply this kind of flipping to all images stemming from possible other sources. It would thereby introduce a new convention that we'd have to remind ourselfs off.

from_rgba8 doesn't do the flipping, make_raw does which still just takes normal unflipped pixel data

PSteinhaus:
> from_rgba8 doesn't do the flipping, make_raw does which still just takes normal unflipped pixel data

Sry, that's true, I mixed them up. I meant that the way images usually get loaded is through ```from_rgba``` which always calls ```make_raw```. I admit that right now ```make_raw``` is the only way to create an image and it might stay the only way forever, so this last point is not that important. The idea of having to flip every image is just a bit unattractive to me, though it has its benefits and making sure that images are always flipped when being created should in practice be no problem at all I think.

nobbele:
I actually don't know how offset is supposed to work normally

nobbele:
showcase

nobbele:
offset.y might just need to be flipped too?, [0.5, -0.5] centers it

nobbele:
fixes #747

icefoxen:
"Doesn't seem hard, you just flip a coordinate..."  Nine commits of fiddling with coordinates later, it works!  I applaud your persistance.  XD

The shadows example work with this change for me on Windows and Linux using AMD graphics, and on Linux using NVidia graphics.  It makes the `canvas_subframe` example scroll backwards on the y axis but that should be an easy fix when it's not 1 am.

PSteinhaus:
Take a look at the transforms example.
Change the angle DrawParam to something like this:
```rust
let param = graphics::DrawParam::new()
            .dest(Vec2::new(400.0, 400.0))
            .rotation(self.pos_x / 100.0)
            //.src((Rect::new(0.5,0.5,0.5,0.5))
            //.offset(Vec2::new(0.25, 0.25))
            .scale(Vec2::new(1.0, 2.0));
```
When the y-scaling is 2 it rotates as if the offset is (0, 0.5).
Also the src parameter doesn't work as intended. With the src param from above it shows the right part of the image, but at a wrong y position (that position being of by half the sprite height).

I haven't tried much more, but you get the idea. Could you maybe try to explain the mathematics behind your approach? Then we could try to find the errors together.

But before we do any of that I have to be the party pooper again and ask:

Are you sure that TODO on the ```Transform::Matrix(m)``` case can really be done? Because there we don't have any of the beautiful values we have in the ```Transform::Values``` case. If we look at [how these values get crunched into a matrix](https://github.com/ggez/ggez/blob/968dd8c09eb9c57ab18662d32532fe129189ac4b/src/graphics/drawparam.rs#L72-L90) I'm pretty sure we can't extract those values from the matrix itself.
The only way to somewhat fix this, that I can think of right now, is to change the computation of the matrix to include all the fixes that in this PR are currently done in the step before. But of course even that wouldn't completely fix the issue (apart from the fact that we would then have to give a flag to ```to_bare_matrix``` to let it differentiate between DrawParams of images and other entities), since it would practically remove the ability of users to write and use their own matrices, since trying to do this via the new ```transform``` function would not give the results they expect.

Overall, trying to go this way (meaning fiddling with the parameters like that) will inevitably lead to incompatability of matrices used for things like the projection (meaning "reasonable" matrices) and the hacky monstrum format we created, won't it?

PSteinhaus:
How about this:

Instead of flipping all images and accounting for it by flipping the pixels in ```make_raw``` we go back to the way things are handled as of now, meaning we only flip the canvas before drawing it, but we account for the flipped image inside of the canvas by applying the pixel flip only when ```into_inner``` is called.

That way we fix #693 and using the coordinate-fiddling you're currently applying to images we could possibly even fix #747, meaning we could get canvases to react to DrawParams as they're actually supposed to.

Of course this would mean that canvases would have "unreasonable" matrices. But since canvases are much less used than pure images I think this might be an acceptable tradeoff (one that is, might I add, already in play, since there already IS some fiddling with coordinates in the draw function of Canvas).

We'd have to add clear warnings to the documentation of Canvas, to inform users that they shouldn't throw their own handmade matrices into DrawParams that are used on Canvases and we'd have to add a warning to ```Canvas::image```, warning users that this image will be flipped on the y axis due to OpenGL, but that's about it.

## Digging deeper

If we don't want that then we might need to dig a bit deeper to solve this issue.
The fundamental reason for all the incorrect behaviour of canvases seems to be that their coordinate system is actually flipped on the y axis, starting from the bottom left, right? 

But when I look at how this can be I only get until the ```RawRenderTargetView<Spec::Resources>``` of CanvasGeneric. If I try to go deeper (looking at the GlBackend) all I run into are handles. If I instead look at how they are created differently I see that the ```RawRenderTargetView``` of our global context gets created though ```backend.init()``` while that of the Canvas gets created with ```factory.view_texture_as_render_target_raw()```, so the difference somehow has to stem from there right?
Then I ask myself what on earth we do in ```backend.init()``` to setup the coordinate system of ggez like it is? It has to be that glutin somehow configures the coordinate system to be like that, right? So in the end we might to start digging there to find out how they do it, so we can (hopefully) apply that to our canvases as well...

Maybe I'm completely on the wrong track here, but that's how I made sense of it.

nobbele:
I was also looking at a way to change the coordinate system but couldn't find anything

nobbele:
uhm what

nobbele:
this looks too simple but it works for everything I tested

PSteinhaus:
... Now I too applaud your persistence. 🥇 

I just tested it and it works like a charm.

Even the canvas subframe example now makes sense, and I mean actual sense!
If you update the code there to
```rust
        if timer::ticks(ctx) % 100 == 0 {
            println!("Delta frame time: {:?} ", timer::delta(ctx));
            println!("Average FPS: {}", timer::fps(ctx));
        }

        // Bounce the rect if necessary
        let (w, h) = graphics::size(ctx);
        if self.draw_pt.x + (w as f32 / 2.0) > (w as f32) || self.draw_pt.x < 0.0 {
            self.draw_vec.x *= -1.0;
        }
        if self.draw_pt.y + (h as f32 / 2.0) > (h as f32) || self.draw_pt.y < 0.0
        {
            self.draw_vec.y *= -1.0;
        }
        self.draw_pt += self.draw_vec;
        Ok(())
    }

        graphics::clear(ctx, [0.1, 0.2, 0.3, 1.0].into());
        self.draw_spritebatch(ctx)?;
        let dims = self.canvas.image().dimensions();
        let src_x = self.draw_pt.x / dims.w;
        let src_y = self.draw_pt.y / dims.h;
        graphics::draw(
            ctx,
            &self.canvas,
            graphics::DrawParam::new()
                .dest(self.draw_pt)
                .src(graphics::Rect::new(src_x, src_y, 0.5, 0.5)),
        )?;
        graphics::present(ctx)?;
        Ok(())
    }
```
then it works. And not only does it work, the canvas now even actually responds to DrawParam as one would expect!
I'm amazed.

To be honest, I just stared at your newest commit for 3 minutes before realizing **why** it is that it works.
But when it hit I immediately had to shake my head and smile. Man, why didn't we think of this earlier 😄

Great work.

To answer your question, I really think we should use this in Canvas::draw, since it does come with a tiny performance cost and and weighing the cost of doing this for potentially hundreds of sprites vs doing it only for a few canvases I absolutely prefer the latter. (Also the idea of **literally flipping every damn image ever used** just doesn't sit well with me, to be honest.)

On the other side this would leave us with the last of the problems that I mentioned when suggesting that we should do stuff like this on the canvas instead of the image: It would result in ```Canvas::image``` being mirrored, which people could trip over. Still, adding a warning to ```Canvas::image``` explaining why this is, should be enough to help users navigate this issue, I think.
You may disagree though, so I look forward to hearing your opinions.

But overall I'm enthralled by how easily this new approach actually and logically solves pretty much all our problems here.

nobbele:
Flipping both every image (*twice* even) *and* every canvas doesn't sit well with me either.
I think that if we rename `Canvas::image` to `Canvas::raw_image` or something it would look like something that might do what you don't expect (upside-down) and then `Canvas::into_image` would flip and return Image rather than &Image which is a considerably less intrusive signature change than changing from_rgba8 and make_raw from &[u8] to Vec<u8>.
What do you think of that
### TL;DR (compared to devel)
`Canvas::image(&self) -> &Image` becomes `Canvas::raw_image(&self) -> &Image` (unflipped)
`Canvas::into_inner(self) -> Image` becomes `Canvas::into_image(self) -> Image` (flipped)
add `Canvas::image(&self) -> Image` (flipped)

PSteinhaus:
> What do you think of that
> 
> ### TL;DR (compared to devel)
> `Canvas::image(&self) -> &Image` becomes `Canvas::raw_image(&self) -> &Image` (unflipped)
> `Canvas::into_inner(self) -> Image` becomes `Canvas::into_image(self) -> Image` (flipped)
> add `Canvas::image(&self) -> Image` (flipped)

YES.

nobbele:
Not that it matters *that* much and is probably overkill but would it be possible to move the flip code into a compute shader or something so you don't have to transfer the buffer from gpu, flip and transfer back.

PSteinhaus:
Agreed. That is probably overkill for now.
I'm no shader expert either, but using one to compute the flip would definetely be a bit faster, and shouldn't be too hard to implement. But considering that you already refined your approach to be pretty performant and especially considering that I cannot think of any use case right now that would make it necessary to create physical copies of canvases during a tight loop or such, I don't think this should have high priority until proven otherwise.

Placing a ```TODO``` somewhere near or inside ```flip_pixel_data``` should suffice for now ;)

PSteinhaus:
```encode``` not working properly on images coming from canvases, for example, is something I'd be more interested in solving than that. (though I haven't double checked whether this issue still persists now and to be honest it's not _that_ important either (though on second thought it might be useful for features like letting users draw images on surfaces that would then need to be exported to a file, to gain persistence between game sessions or something like that))

For now I'm just happy you managed to fix #693 and #747 (and maybe/probably other nasty things yet to be discovered in the tracker) all at the same time :D

This is exactly what I was talking about when (in the maintainer discussion) I said that I hoped we could fix some more rendering bugs and finally create a (more) consistent behavior that one could trust in and depend on. I just didn't think that we'd come this far in a mere few days 😅

**Comment by icefoxen** [comment]
> would it be possible to move the flip code into a compute shader or something so you don't have to transfer the buffer from gpu, flip and transfer back.

It's certainly possible, but IMO getting an image out of the GPU and back to the CPU should never be in the "hot path" of code, so optimizing it isn't necessary.

**Comment by nobbele** [comment]
You know actually, to_rgba8 flips the image already, why not just _not_ flip it

**Comment by nobbele** [comment]
I am pretty sure I did something like this before but something broke but I can't think of what

**Comment by icefoxen** [comment]
Hmmm, I'm not sure how feasible it is and I won't block the PR on it, but this functionality seems like a *great* thing to be able to unit test.  There's lots of moving parts interacting with each other and it's tedious to validate by hand that each one does what it should.  I might try adding a few later.

**Comment by nobbele** [comment]
We could split the drawparam flip code into a separate function to make it easier to test and re-use (flip_x, flip_y drawparams?)

**Comment by nobbele** [comment]
> encode not working properly on images coming from canvases, for example, is something I'd be more interested in solving than that. (though I haven't double checked whether this issue still persists now and to be honest it's not that important either)
yeah this needs to be solved too somehow, but despite having looked around the code I honestly have no idea what could cause it

**Comment by PSteinhaus** [comment]
> We could split the drawparam flip code into a separate function to make it easier to test and re-use (flip_x, flip_y drawparams?)

Probably a good idea, yes.

> Also are there visual tests yet? Rendering some test code to canvas and comparing with known good output image

Not to my knowledge... But I like the idea, basically. On the downside this would lead to some more maintenance, since we'd have to remember to create to new test images whenever an example is changed or added.

**Comment by PSteinhaus** [comment]
> yeah this needs to be solved too somehow, but despite having looked around the code I honestly have no idea what could cause it

Me neither. But I think we should discuss this in a new issue. I'll open one right now.

**Comment by nobbele** [comment]
So this should be ready to merge then (after I split the flip code I guess)

**Comment by nobbele** [comment]
Actually, is #938 fixed for to_image(but not raw_image) with the new change of recreating the image on call

**Comment by PSteinhaus** [comment]
Indeed, it seems fixed.

But testing that I just found out that ```graphics::screenshot``` now creates flipped images (which it didn't to before) both when you screenshot the whole screen and when you just screenshot a canvas (which apart from the flipping works fine too btw).

Calling encode on ```raw_image``` works as expected as well (expected meaning that it turns out flipped).

**Comment by PSteinhaus** [comment]
I guess we just have to... flip the image that is created inside ```graphics::screenshot``` before returning it too, right?

I mean, I'm not really sure why we have to do that though... After all the image created in this function is directly created from our current RawRenderTargetView. So why is it flipped both when this view belongs to our screen and when it belongs to some canvas? Wouldn't that imply that we'd need to to this flipping stuff even when not drawing a canvas? But... we don't have to... **why**?

**Comment by icefoxen** [comment]
I am super sympathetic, and I am also *really* glad that I'm not the only person who gets driven insane by canvases.  It's one of the leakiest damn abstractions in ggez, and for whatever reason any change to it gets *everywhere*.  That was honestly one of the main motivations behind the `ggraphics` attempted-rewrite.

**Comment by nobbele** [comment]
`graphics::screenshot` when drawn, displays nothing and when encoded, is solid black for me

**Comment by PSteinhaus** [comment]
...you mean... generally?
For me it works as intended, I just double checked, except from being flipped...

**Comment by nobbele** [comment]
Could you send me some test code?

**Comment by nobbele** [comment]
devel is same (was image::encode upside down before..?)

**Comment by nobbele** [comment]
hm that does work (except flipped) yeah

**Comment by PSteinhaus** [comment]
Ok... now in devel... screenshot and encode on Canvas::image work as expected... but when I then take a screenshot AFTER switching back to ```None``` as canvas I now get a black image as a screenshot... what?

**Comment by PSteinhaus** [comment]
So, using the PR I get same images as in devel, just flipped when encoded to a file. The screenshot of the whole screen is still working as a drawable image in-game and coming out black when encoded, just like in devel.

**Comment by PSteinhaus** [comment]
> You know actually, to_rgba8 flips the image already, why not just _not_ flip it.
> 
> EDIT: Image is flipped now

Ok. this is annoying.

If we choose to not flip it (PR) then:
- screenshots of the whole screen are: drawn flipped and encoded flipped (incorrect) (screenshot works (i.e. is not transparent black) in astroblasto but not render_to_image)
- screenshots of canvases are: both drawn flipped and encoded flipped (correct or incorrect?)
- raw images of canvases are encoded flipped (but that's how it's supposed to be according to the new documentation)
- images loaded from files are: drawn upright and encoded upright (correct)
- creating new images from existing ones using ```to_rgba8``` and then ```from_rgba8``` creates identical copies (as it should be)

If we choose to flip it (devel) then:
- screenshots of the whole screen are: drawn flipped and encoded upright (also incorrect) (screenshot works (i.e. is not transparent black) in astroblasto but not render_to_image)
- screenshots of canvases are: drawn flipped and encoded upright (definetely incorrect since inconsistent)
- raw images of canvases are encoded upright (incorrect)
- images loaded from files are: drawn upright and encoded flipped (incorrect)
- creating new images from existing ones using ```to_rgba8``` and then ```from_rgba8``` creates flipped copies (definetely incorrect)

So we can state that we should definetely not flip images in ```to_rgba8```.

Concerning screenshots we should probably flip every screenshot, since it's just a copy of our output texture and therefore not flipped yet in any case, ok?

Please note that I edited the code that I posted for ````render_to_image``` here. With this modification we can see that encode just seems to work. It seems that screenshot is the problem here.
@nobbele can you give me an example where encode on an image created by ```Canvas::into_inner``` or ```Canvas::image``` just gives black output (on devel)?

**Comment by PSteinhaus** [comment]
This PR, over the time of its lifespan, turns out to stir up and fix way more bugs than expected ^^

**Comment by nobbele** [comment]
```rust
use ggez::event::{self, EventHandler};
use ggez::graphics::{self, Canvas, Color, DrawParam, Image, ImageFormat};
use ggez::Context;
use ggez::ContextBuilder;
use ggez::GameResult;

struct Application {
    canvas_image: Image,
}

impl Application {
    fn new(ctx: &mut Context) -> Self {
        let image = Image::new(ctx, "/download.png").unwrap();

        let canvas = Canvas::with_window_size(ctx).unwrap();
        graphics::set_canvas(ctx, Some(&canvas));
        graphics::draw(ctx, &image, DrawParam::new()).unwrap();
        graphics::set_canvas(ctx, None);

        let canvas_image = canvas.into_inner();

        canvas_image
            .encode(ctx, ImageFormat::Png, "/canvas_image.png")
            .unwrap();

        Self { canvas_image }
    }
}

impl EventHandler for Application {
    fn update(&mut self, _: &mut Context) -> GameResult<()> {
        Ok(())
    }

        graphics::clear(
            ctx,
            Color {
                r: 0.,
                g: 0.3,
                b: 0.,
                a: 1.,
            },
        );

        graphics::draw(ctx, &self.canvas_image, DrawParam::new().dest([400.0, 0.0])).unwrap();

        graphics::present(ctx).unwrap();
        Ok(())
    }
}

pub fn main() {
    let (mut ctx, event_loop) = ContextBuilder::new("canvas", "athorus").build().unwrap();
    let app = Application::new(&mut ctx);
    event::run(ctx, event_loop, app);
}
```

black image on devel

nobbele:
Can you do one last check on this @PSteinhaus. I tested every code I have and it all works now.

src/graphics/canvas.rs
```diff
@@ -207,25 +207,15 @@ impl Drawable for Canvas {
 
         // We have to mess with the scale to make everything
         // be its-unit-size-in-pixels.
-        let scale_x = param.src.w * f32::from(self.width());
-        let scale_y = param.src.h * f32::from(self.height());
+        let scale_x = param.src.w * self.width() as f32;
+        let scale_y = param.src.h * self.height() as f32;
```
PSteinhaus: Why use ```as``` here instead of ```from```?

PSteinhaus:
> Can you do one last check on this @PSteinhaus. I tested every code I have and it all works now.

Will do.

PSteinhaus:
Tested everything and it all works beautifully now.

src/graphics/canvas.rs
nobbele: habit, always used as instead of from, should probably be changed to be consistent with the rest of the library

IGBC:
This PR introduces a Regression in the canvas API. For some reason committing to a canvas now requires calling `graphics::present(ctx)`, who's function is to move to the next frame in the swapchain. This is breaking real time rendering to canvases. Which is needed for pixel level screen effects like bloom, colour grain and other full screen after effects (I have a CRT emulation shader in my game)

Formal Request to revert merge until this can be corrected.

The following test should draw a black square to the screen using real time canvas rendering. It does not work with or without a `graphics::present(ctx)` before deselecting the canvas.

use ggez::event::{self, EventHandler};
use ggez::{graphics, Context, ContextBuilder, GameResult};
use graphics::Drawable;

fn main() {
    // Make a Context.
    let (mut ctx, mut event_loop) = ContextBuilder::new("my_game", "Cool Game Author")
        .build()

    // Create an instance of your event handler.
    // Usually, you should provide it with the Context object to
    // use when setting your game up.
    let mut my_game = MyGame::new(&mut ctx);

    // Run!
}

struct MyGame {
    // Your state here...
}

impl MyGame {
    pub fn new(_ctx: &mut Context) -> MyGame {
        // Load/create resources such as images here.
        MyGame {
		    // ...
		}
    }
}

impl EventHandler for MyGame {
        // Update code here...
        Ok(())
    }

        graphics::clear(ctx, graphics::Color::WHITE);
        // Draw code here...
        let canvas = graphics::Canvas::with_window_size(ctx)?;
        graphics::set_canvas(ctx, Some(&canvas));
        graphics::Mesh::new_rectangle(
            ctx,
            graphics::DrawMode::fill(),
            graphics::Rect::new(10., 10., 100., 100.),
            graphics::Color::BLACK
        )?
        .draw(ctx, graphics::DrawParam::new())?;
        graphics::present(ctx)?;
        graphics::set_canvas(ctx, None);
        let image = canvas.to_image(ctx)?;
        image.draw(ctx, graphics::DrawParam::new())?;
        graphics::present(ctx)
    }
}

PR #937, opened 2021-05-25
Update READ.ME

jean-lourenco:
First code example updated to keep parity with the official docs: https://docs.rs/ggez/0.6.0-rc1/ggez/index.html
The code example doesn't even compile out of the box without the `&mut` and the `Result` handling

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,17 +115,20 @@ use ggez::event::{self, EventHandler};
 
 fn main() {
     // Make a Context.
-    let (mut ctx, event_loop) = ContextBuilder::new("my_game", "Cool Game Author")
+    let (mut ctx, mut event_loop) = ContextBuilder::new("my_game", "Cool Game Author")
         .build()
         .expect("aieee, could not create ggez context!");
 
     // Create an instance of your event handler.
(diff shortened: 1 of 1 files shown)


nobbele:
Thank you, I think I fixed this on deve l though?

PSteinhaus:
> Thank you, I think I fixed this on deve l though?

You updated the README, but not in this way. I think [this commit](https://github.com/ggez/ggez/commit/e5094ae820c29cbb4b79b2a3e73865fbce8f4375) gave the example code there its current form.

@jean-lourenco 
I just tried it out and sadly the code from the official docs doesn't work with ggez 0.6.0-rc1 anymore.
Therefore we need to update the docs by releasing another rc crate or 0.6.0 in its final form.

@icefoxen Alternatively, you should be able to modify that example code in the docs directly, right?

icefoxen:
Nope, docs.rs pulls directly from what is published on crates.io .  Might be time to publish an rc2.

PSteinhaus:
Hmm, sad. But good to know. We should really publish 0.6.0-rc2.

PR #939, opened 2021-05-26
Snake example: Simplify non-negative modulo computation, by using rem_euclid() API

saveriomiroddi:
The functionality previously implemented manually, via ModuloSigned trait, is actually already existing in the Rust APIs - [rem_euclid()](https://doc.rust-lang.org/std/primitive.i16.html#method.rem_euclid).

While using traits is always cool™, it's best to reduce the cognitive overhead on the readers, especially considering that this is an example for beginners.


nobbele:
Thank you for your contribution, but could you open a new PR with these changes on the devel branch instead?

saveriomiroddi:
> Thank you for your contribution, but could you open a new PR with these changes on the devel branch instead?

Ah! Sorry, I didn't know the project structure :) Will do shortly.

saveriomiroddi:
@nobbele Done.

nobbele:
The modified comment is definitely an improvement (maybe mention negative input like `as the standard remainder function (%) returns a negative value for negative inputs`?).
You can decide whether we should have % or rem_euclid in the positive directions, doesn't really matter which way.

PR #943, opened 2021-05-30
Removes PhysicalFS create_root calls

### LINKED ISSUE #677: Empty resources directory is created when creating Context.
**Describe the bug**
When creating a `Context` using `ContextBuilder::build` an empty resource directory is created next to the application.

**To Reproduce**
Create an application with the following code and run it:
```rust
pub fn main() -> ggez::GameResult {
    let cb = ggez::ContextBuilder::new("resources-test", "Eran Cohen").window_setup(
        ggez::conf::WindowSetup::default()
    );
    let (ctx, event_loop) = &mut cb.build()?;
}
```
**Expected behavior**
Calling `ContextBuilder::build` should **not** create an empty resource directory.

**The source of the behavior**
By default, `ContextBuilder::load_conf_file` is set to `true`, so when `build` is called the following line in *context.rs* is run: (line 299)
```rust
fs.read_config().unwrap_or(self.conf)
```
`read_config` calls `is_file` (*filesystem.rs*, line 327) which then calls `vfs.metadata`(*filesystem.rs*, line 233) which then calls another `vfs.metadata` (*vfs.rs*, line 499) and brings us to *vfs.rs*, line 350: 

```rust
fn metadata(&self, path: &Path) -> GameResult<Box<dyn VMetadata>> {
    self.create_root()?;
    let p = self.to_absolute(path)?;
    p.metadata()
        .map(|m| Box::new(PhysicalMetadata(m)) as Box<dyn VMetadata>)
        .map_err(GameError::from)
}
```
which as can be seen, calls `create_root` and creates the directory.

**Hardware and Software:**
 - ggez version: *0.5.1*
 - OS: *Windows 10*, *Windows 7*
 - Graphics card: *Intel HD Graphics 530*
 - Graphics card drivers: *Intel*

**Workaround**
Creating a `Context` without trying to load from a configuration file doesn't create the directory:
```rust
let cb = ggez::ContextBuilder::new("life-rs", "Eran Cohen").with_conf_file(false);
```

nobbele:
I don't see why it even needs to do that.
Fixes #677

### DIFF
--- a/src/vfs.rs
+++ b/src/vfs.rs
@@ -239,6 +239,19 @@ impl PhysicalFS {
             Err(GameError::FilesystemError(msg))
         }
     }
+
+    /// Creates the PhysicalFS's root directory if necessary.
+    /// Idempotent.
+    /// This way we can not create the directory until it's
+    /// actually used, though it IS a tiny bit of a performance
+    /// malus.
+    fn create_root(&self) -> GameResult {
+        if !self.root.exists() {
+            fs::create_dir_all(&self.root).map_err(GameError::from)
+        } else {
+            Ok(())
+        }
+    }
 }
 
 impl Debug for PhysicalFS {
@@ -263,6 +276,10 @@ impl VFS for PhysicalFS {
             return Err(GameError::FilesystemError(msg));
         }
 
+        if open_options.create {
+            self.create_root()?;
+        }
+
         let p = self.to_absolute(path)?;
         open_options
             .to_fs_openoptions()
@@ -280,6 +297,8 @@ impl VFS for PhysicalFS {
                     .to_string(),
             ));
         }
+
+        self.create_root()?;
 
         let p = self.to_absolute(path)?;
         //println!("Creating {:?}", p);
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -9,6 +9,7 @@
   * [Resolution independence](#gfx_resolution)
 * **[Libraries](#libraries)**
   * [Can I use `specs`, `legion` or another entity-component system?](#library_ecs)
+  * [What is mint and how do I use `Into<mint::Point2<f32>>` and other `Into<mint::T>` types?](#library_mint)
 * **[Performance](#performance)**
   * [Image/sound loading and font rendering is slow!](#perf_slow1)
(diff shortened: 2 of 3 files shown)


icefoxen:
It does it so it never needs to check whether the various config/save/etc directories exist, they just always do.  Whether it *still* needs to do this, idk.

PSteinhaus:
> Whether it _still_ needs to do this, idk.

It does. I just tested it and the logging and files examples fail using this PR. The astroblasto example also throws an error when trying to create a screenshot.

I guess we need these checks.
But maybe we can find out why the empty directory is created even when there is no need for it.
Going through the code I don't see why ```create_root``` gets called at all when we build the context... maybe I'll try to debug it with gdb tomorrow.

src/vfs.rs
```diff
@@ -275,7 +262,7 @@ impl VFS for PhysicalFS {
             );
             return Err(GameError::FilesystemError(msg));
         }
-        self.create_root()?;
+
```
PSteinhaus, line 265:
```suggestion
        if open_options.create {
            self.create_root()?;
        }
```

src/vfs.rs
```diff
@@ -293,7 +280,7 @@ impl VFS for PhysicalFS {
                     .to_string(),
             ));
         }
-        self.create_root()?;
+
```
PSteinhaus, line 283:
```suggestion
        self.create_root()?;
```

src/vfs.rs
```diff
@@ -239,19 +239,6 @@ impl PhysicalFS {
...
-    /// This way we can not create the directory until it's
-    /// actually used, though it IS a tiny bit of a performance
-    /// malus.
-    fn create_root(&self) -> GameResult {
-        if !self.root.exists() {
-            fs::create_dir_all(&self.root).map_err(GameError::from)
-        } else {
-            Ok(())
-        }
-    }
```
PSteinhaus, line 254: If we want to use this in ```mkdir``` and ```open_options``` we need to keep it.

PSteinhaus (commented):
I just tried all the relevant examples with these changes and they seem to work fine now.

PSteinhaus:
I think this is ready to merge.

IGBC (approved):

PR #946, opened 2021-06-03
BlendChannel rules

PSteinhaus:
```BlendMode::Premultiplied``` uses a different rule for calculating the alpha channel when drawing. It's (IMO) clearly better, since it does what you'd expect from alpha. Instead of two objects with alpha 0.5 creating a pixel with an alpha of 1 this new rule creates a pixel with an alpha of 0.75 (which is what's to be expected when two objects that are each half transparent are laid ontop of each other).

To make everything consistent again i propose to change ```BlendMode::Alpha``` to also adapt this new rule. It shouldn't change the output _that_ much. At least I've found no example where this change clearly breaks something badly (though I'm sure such an example could be constructed, if one really wanted to do so).
In fact, it should make the output more consistent with what a user should expect, though I do realize that someone porting from 0.5 to 0.6 might notice it and not like it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,18 +45,18 @@ old_school_gfx_glutin_ext = "0.26"
 glutin = "0.26"
 winit = "0.24"
 image = {version = "0.23", default-features = false, features = ["gif", "png", "pnm", "tga", "tiff", "webp", "bmp", "dxt", ] }
-rodio = { version = "0.13", default-features = false, features = ["flac", "vorbis", "wav"] }
+rodio = { version = "0.14", default-features = false, features = ["flac", "vorbis", "wav"] }
 serde = "1"
 serde_derive = "1"
 toml = "0.5"
 log = "0.4"
 lyon = "0.17.5"
 smart-default = "0.6"
-glam = { version = "0.12", features = ["mint"]}
+glam = { version = "0.15", features = ["mint"]}
 # Has to be the same version of mint that our math lib uses here.
 mint = "0.5"
 gilrs = "0.8"
-approx = "0.4"
+approx = "0.5"
 bytemuck = "1.5.1"
 
 [dev-dependencies]
--- a/docs/BuildingForEveryPlatform.md
+++ b/docs/BuildingForEveryPlatform.md
@@ -2,11 +2,10 @@
 
 Greetings, one and all.  Today we shall explore how to build and
 deploy a `ggez` game for every possible platform.  For platforms like
-Linux it's pretty darn simple.  For ones like Android it gets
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Comment by PSteinhaus** [comment]
The main thing I'm thinking about is whether we might want to change the alpha rules of other blend modes as well to conform to this new standard.

For example ```BlendMode::Add``` currently simply adds the alpha of source and destination. How about redefining Add as:
```rust
BlendMode::Alpha => Blend {
    color: BlendChannel {
        equation: Equation::Add,
        destination: Factor::One,
    },
    alpha: BlendChannel {
        equation: Equation::Add,
        source: Factor::OneMinus(BlendValue::DestAlpha),
        destination: Factor::One,
    },
},
```
This also changes the rule for the color channel to respect alpha as well, making ```Add``` actually work as one coming from some other photomanipulation software or the like might expect.

PSteinhaus:
To be honest I'd really like to change Lighten and Darken this way as well:
```rust
BlendMode::Lighten => Blend {
    color: BlendChannel {
        equation: Equation::Max,
        destination: Factor::One,
    },
    alpha: BlendChannel {
        equation: Equation::Add,
        source: Factor::ZeroPlus(BlendValue::SourceAlpha),       // here I'm honestly not yet sure how/if we'd want to change it
        destination: Factor::OneMinus(BlendValue::SourceAlpha), // does someone understand why this rule is the way it is?
    },
},
BlendMode::Darken => Blend {
    color: BlendChannel {
        equation: Equation::Min,
        destination: Factor::One,
    },
    alpha: BlendChannel {
        equation: Equation::Add,
    },
},
```
As the comment says I'm a bit lost though as to why the rule for the alpha channel here is
```rust
    alpha: BlendChannel {
        equation: Equation::Add,
    },
```
which translates to ```new_A = src_A * src_A + dest_A * (1 - src_A)```. Does someone get the logic behind this?
IMO it should be ```new_A = max( src_A, dest_A )``` or something like that.
The best rule would probably be something more like
```
new_A = if max( src_R * src_A, dest_R * dest_A ) == src_R * src_A ||
           max( src_G * src_A, dest_G * dest_A ) == src_G * src_A ||
           max( src_B * src_A, dest_B * dest_A ) == src_B * src_A
           { src_A * (1 - dest_A) + dest_A } else { dest_A }
```
(for Lighten)
So that the alpha of a pixel is only increased when we actually lighten it.

IGBC:
This PR has all the changes @icefoxen and I made yesterday now bundled up into it. Could you delete and recreate the PR from your branches head when you get a chance? Thanks.

PSteinhaus:
My bad.

PR #948, opened 2021-06-04
BlendChannel rules

PSteinhaus:
For context see #946.

### DIFF
--- a/examples/blend_modes.rs
+++ b/examples/blend_modes.rs
@@ -24,7 +24,7 @@ impl MainState {
             ctx,
             graphics::DrawMode::fill(),
             Vec2::new(0.0, 0.0),
-            50.0,
+            45.0,
             0.5,
             Color::WHITE,
         )?;
@@ -49,7 +49,7 @@ impl MainState {
             Color::new(0., 0., 0.8, 0.8),
         ];
 
-        const OFFSET: f32 = 30.;
+        const OFFSET: f32 = 24.;
         const REL_POSITIONS: [[f32; 2]; 3] = [
             [-OFFSET, -OFFSET / 2.],
             [OFFSET, -OFFSET / 2.],
@@ -83,7 +83,7 @@ impl MainState {
     fn draw_venn_diagrams(&mut self, ctx: &mut Context) -> GameResult<()> {
         let (w, h) = graphics::drawable_size(ctx);
         let y = h / 4.;
-        const MODE_COUNT: usize = 6;
+        const MODE_COUNT: usize = 7;
         let x_step = w / (MODE_COUNT + 1) as f32;
 
         // draw with Alpha
@@ -102,13 +102,17 @@ impl MainState {
         self.circle.set_blend_mode(Some(BlendMode::Multiply));
         self.draw_venn(ctx, [x_step * 4., y].into(), "Multiply")?;
 
+        // draw with Invert
+        self.circle.set_blend_mode(Some(BlendMode::Invert));
+        self.draw_venn(ctx, [x_step * 5., y].into(), "Invert")?;
+
         // draw with Darken
         self.circle.set_blend_mode(Some(BlendMode::Darken));
-        self.draw_venn(ctx, [x_step * 5., y].into(), "Darken")?;
+        self.draw_venn(ctx, [x_step * 6., y].into(), "Darken")?;
 
         // draw with Lighten
         self.circle.set_blend_mode(Some(BlendMode::Lighten));
-        self.draw_venn(ctx, [x_step * 6., y].into(), "Lighten")?;
+        self.draw_venn(ctx, [x_step * 7., y].into(), "Lighten")?;
 
         Ok(())
     }
--- a/examples/blend_modes.rs
+++ b/examples/blend_modes.rs
@@ -161,7 +161,8 @@ impl EventHandler for MainState {
                 .dest(Vec2::new(8., 4.))
                 .color(Color::WHITE),
         )?;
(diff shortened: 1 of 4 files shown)


PSteinhaus:
As stated in the older PR I kinda also want to change `Add`, `Darken` and `Lighten`, but I'd need your opinion on that.
The least I'd like to do is review the alpha rules for `Darken` and `Lighten`, because they don't seem to make any sense to me...

IGBC (commented):
can we get some before and after screenshots from some of the example demos so we can see how they are different?

PSteinhaus:
I tested all relevant examples and they all look completely the same as far as I can see.
The reason for this is that all current examples either feature no alpha images at all, or alpha images drawn upon an opaque background (where this changed blend rule doesn't affect anything).

We'd need to create a new example where transparent objects are drawn ontop of each other.

PSteinhaus:
> It's probably a good idea to create a more general example showcasing the different blend modes next to each other.

Done. Tell me if you think it does everything it needs to do.
Now that we have it we can use it to discuss changing blending modes more practically.

I also fixed a small, but potentially quite disastrous bug that @Elinvynia first mentioned in her PR about the premultiplied blending mode, but that we simply ignored back then: Meshes currently do not respect the blend mode you give them. There simply was no code in their draw function yet to even use it.

PSteinhaus:
I saw that I should probably also add the `Invert` blend mode, so I added it.
I also went ahead and changed `Add` and `Subtract` to behave more reasonably.
What's reasonable for me hasn't to be reasonable for you though:

canvas using premultiplied blend mode:
new

To be honest I have pretty much no idea what `Invert` is supposed to do, but intuitively it seems pretty broken/useless to me as well. It's another preset though, so there _might_ be some idea behind this that I just can't fathom.

PSteinhaus:
Ok, so here is my final example with the semi-final blend modes:

Canvas using alpha blend mode:
blend

Canvas using premultiplied blend mode:
blend_pre

And a short explanation of each ruleset:

```rust
            BlendMode::Alpha => Blend {
                color: BlendChannel {
                    equation: Equation::Add,
                },
                alpha: BlendChannel {
                    equation: Equation::Add,
                    source: Factor::OneMinus(BlendValue::DestAlpha),
                    destination: Factor::One,
                },
            },
```
So the resulting color is `src_RGB * src_A + dest_RGB * (1 - src_A)`.
The source color only contributes as much as its transparency allows. The dest color contributes the rest (i.e. what the transparency of the source lets through).

The resulting alpha is `src_A * (1 - dest_A) + dest_A`.
The second summand is obvious, since the alpha of the destination is what we want to add upon. The first summand can best be rationalized by an example: Imagine laying two half transparent objects on top of each other. What you'd now see as a result would be that 75% of light goes through, since each object only lets half of what enters through. This is how light behaves in reality and that's why we cannot simply add `src_A + dest_A`, but have to multiply the source component with the "leftover" transparency `1 - dest_A`.

```rust
            BlendMode::Add => Blend {
                color: BlendChannel {
                    equation: Equation::Add,
                    destination: Factor::One,
                },
                alpha: BlendChannel {
                    equation: Equation::Add,
                    source: Factor::OneMinus(BlendValue::DestAlpha),
                    destination: Factor::One,
                },
            },
```
The resulting color is `src_RGB * src_A + dest_RGB`.
The dest_RGB is what we want to add upon. But we don't just add all of src_RGB. Instead we take its transparency into consideration and only take as much as it actually gives.

The resulting alpha is `src_A * (1 - dest_A) + dest_A`.
This is the same as in `Alpha` and the reasoning is again the same.

```rust
            BlendMode::Subtract => Blend {
                color: BlendChannel {
                    equation: Equation::RevSub,
                    destination: Factor::One,
                },
                alpha: BlendChannel {
                    equation: Equation::Add,
                    source: Factor::Zero,
                    destination: Factor::One,
                },
            },
```

The resulting color is `dest_RGB - src_RGB * src_A`.
The dest_RGB is what we want to subtract from. But we don't just subtract all of src_RGB. Instead we take its transparency into consideration and only take as much as it actually gives.

The resulting alpha is simply `dest_A`.
This is a bit of a tricky case, but I decided to handle it like this because it has the benefit of no longer creating black images as shown here in the images of previous versions when used on a transparent (black) canvas. As a user I'd naively assume that using `Subtract` on something shouldn't be able to increase alpha (especially when the previous alpha was zero).

```rust
            BlendMode::Darken => Blend {
                color: BlendChannel {
                    equation: Equation::Min,
                    destination: Factor::One,
                },
                alpha: BlendChannel {
                    equation: Equation::Add,
                    source: Factor::OneMinus(BlendValue::DestAlpha),
                    destination: Factor::One,
                },
            },
            BlendMode::Lighten => Blend {
                color: BlendChannel {
                    equation: Equation::Max,
                    destination: Factor::One,
                },
                alpha: BlendChannel {
                    equation: Equation::Add,
                    source: Factor::OneMinus(BlendValue::DestAlpha),
                    destination: Factor::One,
                },
            },
```
For these two modes I changed the alpha rule to be consistent with those of `Alpha`, `Premultiplied` and `Add`. It just makes sense to me that the same reasoning applies here.

I also changed the color rule to take source alpha into consideration as well, but I'm still unsure whether this is actually the right way to go. In the end it's probably a matter of taste. At least for consistency it may be nice to do it like this.

Note that the lights in the shadows example now look a bit dimmer. That is probably due to the change in the alpha rule of `Alpha` and the changes for `Add` and can probably be mostly fixed by changing some parameters.

Issue #638, opened 2019-06-02
Scaling issue when resizing window on macOS


raybritton:
**Describe the bug**
When any ggez program is run and the window is resized the graphics appear incorrectly.

**To Reproduce**
Run this:
```
use ggez::{Context, ContextBuilder, GameResult, graphics};
use ggez::event::{self, EventHandler};
use ggez::conf::WindowMode;
use ggez::graphics::{MeshBuilder, DrawMode, Rect};
use ggez::mint::Point2;

fn main() {
    let (ctx, event_loop) = &mut ContextBuilder::new("resize_test", "Ray Britton")
        .window_mode(WindowMode {
            resizable: true,
            ..WindowMode::default()
        })
        .build()
        .expect("aieee, could not create ggez context!");

    let mut my_game = MyGame {};

    match event::run(ctx, event_loop, &mut my_game) {
        Ok(_) => println!("Exited cleanly."),
        Err(e) => println!("Error occured: {}", e)
    }
}

struct MyGame {}

impl EventHandler for MyGame {
    fn update(&mut self,_ctx: &mut Context) -> GameResult<()> { Ok(()) }

    fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
        graphics::clear(ctx, (0.,0.,0.,1.).into());

        let mesh = MeshBuilder::new()
            .rectangle(DrawMode::fill(), Rect::new(0.,0.,8.,8.), (1.,1.,1.,1.).into())
            .build(ctx)?;

        let size = graphics::window(ctx).get_inner_size().expect("failed to get window size");
        let dpi = graphics::window(ctx).get_hidpi_factor();

        let sw = (size.width * dpi) as f32;
        let sh = (size.height * dpi) as f32;

        graphics::draw(ctx, &mesh, (Point2 {x: 0., y: 0.},))?;
        graphics::draw(ctx, &mesh, (Point2 {x: sw - 8.,y: sh - 8.},))?;

        graphics::present(ctx)?;

        Ok(())
    }
}
```
and then resize the window

**Expected behavior**
The boxes should continue to appear in the corners of the window

**Screenshots or pasted code**
Initial window: 
<img width="512" alt="initial" src="https://user-images.githubusercontent.com/1632774/58760788-b7cd7600-8534-11e9-9c18-2fb466aa94f7.png">
If made bigger:
[...]

PR #949, opened 2021-06-04
don't resize windows when ScaleFactorChanged events happen

PSteinhaus:
fixes a bug discovered [here](https://github.com/ggez/ggez/issues/638#issuecomment-854378569)

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -354,12 +354,14 @@ pub fn process_event(ctx: &mut Context, event: &mut winit::event::Event<()>) {
                 ctx.keyboard_context.set_key(*keycode, pressed);
             }
             winit_event::WindowEvent::ScaleFactorChanged { new_inner_size, .. } => {
-                // actively set the new_inner_size to be the desired size
-                // to stop winit from resizing our window
-                **new_inner_size = winit::dpi::PhysicalSize::<u32>::from([
-                    ctx.conf.window_mode.width,
-                    ctx.conf.window_mode.height,
-                ]);
+                if !ctx.conf.window_mode.resize_on_scale_factor_change {
+                    // actively set the new_inner_size to be the desired size
+                    // to stop winit from resizing our window
+                    **new_inner_size = winit::dpi::PhysicalSize::<u32>::from([
+                        ctx.conf.window_mode.width,
+                        ctx.conf.window_mode.height,
+                    ]);
+                }
             }
             _ => (),
         },
--- a/src/event.rs
+++ b/src/event.rs
@@ -354,9 +354,12 @@ pub fn process_event(ctx: &mut Context, event: &mut winit::event::Event<()>) {
                 ctx.keyboard_context.set_key(*keycode, pressed);
             }
             winit_event::WindowEvent::ScaleFactorChanged { new_inner_size, .. } => {
-                // actively set the new_inner_size to be the current inner size
+                // actively set the new_inner_size to be the desired size
                 // to stop winit from resizing our window
-                **new_inner_size = ctx.gfx_context.window.window().inner_size();
+                **new_inner_size = winit::dpi::PhysicalSize::<u32>::from([
+                    ctx.conf.window_mode.width,
+                    ctx.conf.window_mode.height,
+                ]);
             }
             _ => (),
         },
--- a/src/event.rs
+++ b/src/event.rs
@@ -164,7 +164,7 @@ where
 {
     use crate::input::{keyboard, mouse};
 
-    event_loop.run(move |event, _, control_flow| {
+    event_loop.run(move |mut event, _, control_flow| {
         if !ctx.continuing {
             *control_flow = ControlFlow::Exit;
             return;
@@ -175,7 +175,7 @@ where
         let ctx = &mut ctx;
         let state = &mut state;
 
-        process_event(ctx, &event);
+        process_event(ctx, &mut event);
         match event {
             Event::WindowEvent { event, .. } => match event {
                 WindowEvent::Resized(logical_size) => {
@@ -311,7 +311,7 @@ where
 /// state it needs to, such as detecting window resizes.  If you are
 /// rolling your own event loop, you should call this on the events
 /// you receive before processing them yourself.
-pub fn process_event(ctx: &mut Context, event: &winit::event::Event<()>) {
+pub fn process_event(ctx: &mut Context, event: &mut winit::event::Event<()>) {
     match event {
         winit_event::Event::WindowEvent { event, .. } => match event {
             winit_event::WindowEvent::Resized(physical_size) => {
@@ -353,8 +353,10 @@ pub fn process_event(ctx: &mut Context, event: &winit::event::Event<()>) {
                 };
                 ctx.keyboard_context.set_key(*keycode, pressed);
             }
-            winit_event::WindowEvent::ScaleFactorChanged { .. } => {
-                // Nope.
+            winit_event::WindowEvent::ScaleFactorChanged { new_inner_size, .. } => {
+                // actively set the new_inner_size to be the current inner size
+                // to stop winit from resizing our window
+                **new_inner_size = ctx.gfx_context.window.window().inner_size();
             }
             _ => (),
         },
--- a/examples/eventloop.rs
+++ b/examples/eventloop.rs
@@ -21,7 +21,7 @@ pub fn main() -> GameResult {
     let mut position: f32 = 1.0;
 
     // Handle events. Refer to `winit` docs for more information.
-    events_loop.run(move |event, _window_target, control_flow| {
+    events_loop.run(move |mut event, _window_target, control_flow| {
         if !ctx.continuing {
             *control_flow = ControlFlow::Exit;
             return;
@@ -33,7 +33,7 @@ pub fn main() -> GameResult {
 
         // This tells `ggez` to update it's internal states, should the event require that.
         // These include cursor position, view updating on resize, etc.
-        event::process_event(ctx, &event);
+        event::process_event(ctx, &mut event);
         match event {
             Event::WindowEvent { event, .. } => match event {
                 WindowEvent::CloseRequested => event::quit(ctx),
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -52,6 +52,7 @@ pub enum FullscreenType {
 ///     max_height: 0.0,
 ///     resizable: false,
 ///     visible: true,
+///     resize_on_scale_factor_change: false,
 /// }
 /// # , WindowMode::default());}
 /// ```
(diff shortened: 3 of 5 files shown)


jiyinyiyong:
going to try this on my macbook...

jiyinyiyong:
could you offer some help with this error:

```bash
# cargo run calcit/compact.cirru
   --> src/main.rs:133:31
    |
133 |     event::process_event(ctx, &event);
    |                               ^^^^^^ types differ in mutability
    |
    = note: expected mutable reference `&mut winit::event::Event<'_, ()>`
                       found reference `&winit::event::Event<'_, ()>`
```

[my code](https://github.com/calcit-lang/calcit-paint/blob/main/src/main.rs#L132) was copied from https://github.com/ggez/ggez/blob/devel/examples/eventloop.rs#L36 but got this error after switch to this branch locally.

PSteinhaus:
I also just realised that this change breaks the example 😅 , so I fixed it with the latest commit.

jiyinyiyong:
now I made it running. but what is supposed to happen? it looks just like I tried before. Dragging window from Dell monitor to mac screen, it's still the size.

PSteinhaus:
What? Moving it from Dell to Mac doesn't shrink it down to the size it has when created on the mac screen?

jiyinyiyong:
Something made me worried was that  the `ScaleFactorChanged` is triggered multiple times if I move window slowly from the lower screen to upper screen. The slower, the more events:

```text
Other window event fired: Moved(PhysicalPosition { x: 574, y: 6 })
Other window event fired: ScaleFactorChanged { scale_factor: 1.0, new_inner_size: PhysicalSize { width: 1000, height: 600 } }
Other window event fired: Moved(PhysicalPosition { x: 288, y: -35 })
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 2.0, new_inner_size: PhysicalSize { width: 2000, height: 1200 } }
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 1.0, new_inner_size: PhysicalSize { width: 1000, height: 600 } }
Other window event fired: Moved(PhysicalPosition { x: 289, y: -46 })
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 2.0, new_inner_size: PhysicalSize { width: 2000, height: 1200 } }
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 1.0, new_inner_size: PhysicalSize { width: 1000, height: 600 } }
Other window event fired: Moved(PhysicalPosition { x: 291, y: -57 })
Other window event fired: Moved(PhysicalPosition { x: 291, y: -58 })
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 2.0, new_inner_size: PhysicalSize { width: 2000, height: 1200 } }
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 1.0, new_inner_size: PhysicalSize { width: 1000, height: 600 } }
Other window event fired: Moved(PhysicalPosition { x: 294, y: -65 })
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 2.0, new_inner_size: PhysicalSize { width: 2000, height: 1200 } }
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 1.0, new_inner_size: PhysicalSize { width: 1000, height: 600 } }
Other window event fired: Moved(PhysicalPosition { x: 294, y: -67 })
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 2.0, new_inner_size: PhysicalSize { width: 2000, height: 1200 } }
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 1.0, new_inner_size: PhysicalSize { width: 1000, height: 600 } }
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
Other window event fired: ScaleFactorChanged { scale_factor: 2.0, new_inner_size: PhysicalSize { width: 2000, height: 1200 } }
unkwnon event: RedrawRequested(WindowId(Id(140465592682624)))
event: ({} (:x 127.89776611328125) (:y 105.34829711914063) (:clicks 1) (:type :mouse-move))
Other window event fired: CursorEntered { device_id: DeviceId(DeviceId) }
Other window event fired: CursorLeft { device_id: DeviceId(DeviceId) }
Other window event fired: CursorEntered { device_id: DeviceId(DeviceId) }
event: ({} (:y 335.4372253417969) (:clicks 1) (:type :mouse-move) (:x 128.5228271484375))
event: ({} (:type :mouse-move) (:x 130.94482421875) (:clicks 1) (:y 347.73980712890625))
```

when I move window from upper screen to lower screen, only one event is triggered.

jiyinyiyong:
> What? Moving it from Dell to Mac doesn't shrink it down to the size it has when created on the mac screen?

tried multiple times. nope, the size never changed.

PSteinhaus:
Ahh, because the eventloop example uses a copy of the process_event function not the function itself...
Let me just update it real quick.

src/event.rs
```diff
@@ -353,8 +353,10 @@ pub fn process_event(ctx: &mut Context, event: &winit::event::Event<()>) {
                 };
                 ctx.keyboard_context.set_key(*keycode, pressed);
             }
-            winit_event::WindowEvent::ScaleFactorChanged { .. } => {
-                // Nope.
+            winit_event::WindowEvent::ScaleFactorChanged { new_inner_size, .. } => {
+                // actively set the new_inner_size to be the current inner size
+                // to stop winit from resizing our window
+                **new_inner_size = ctx.gfx_context.window.window().inner_size();
```
PSteinhaus, line 359: How does this not change new_inner_size for you?
I just don't get it...

jiyinyiyong:
also tried 2 default demos, size not changing:

```rust
cargo run --example logging
cargo run --example eventloop
```

src/event.rs
jiyinyiyong: any logs to capture on my side?

jiyinyiyong: something like?

image

image

PSteinhaus:
Hmm, maybe it's because the window was already resized by the OS itself?
Let me try something.

PSteinhaus:
Now the `new_inner_size` is based upon the `window_mode` or our Context, so that we can be sure it'll be set to the size we actually want.

jiyinyiyong:
while the code is working now... personally I don't like this feature, well, mac..

https://youtu.be/FnuikSafxdo

PSteinhaus:
The many resized are probably due to your OS trying to resize the window immedeately, so that us trying to stay at the same window size results in changing window size twice whenever you move the window inbetween the screens.

That's a bit sad, but I have no idea how to avoid this...
Of course we could just not set `new_inner_size` to our desired value, but then we'd have the problem of inconsistency between windows created on that screen and windows moved there. But the bigger problem is that we'd then suddenly have a window where actual physical window size and what is stated in `window_mode` differ, which could lead to all different kinds of problems.

IGBC (approved):

jiyinyiyong:
> The many resized are probably due to your OS trying to resize the window immedeately, so that us trying to stay at the same window size results in changing window size twice whenever you move the window inbetween the screens.

https://github.com/ggez/ggez/pull/949#issuecomment-854707123

just triggered too many events.

I can use the tricks in previous thread. However, speaking in the angle of better experience, I would prefer the window not being resized, because other windows, Chrome, native windows in macos,(Windows, I'm not using...) their apparent sizes are always stable when dragged to a screen with a different DPI.

PSteinhaus:
Hmm, I understand that.
I mean... we could also do it the other way around, meaning we update our `window_mode` to just reflect the changed window size.

But that would still break all games which didn't plan for resizable windows...

jiyinyiyong:
yep, like it always hard to support different kinds of customers using same piece of program in business. normally people add options to give users the control, but I'm not sure if that's a choice or not for ggez.

PSteinhaus:
Hmm that might be a good idea actually. But this would be a very very specific option (something like `resize_on_screen_scale_factor_changed`)... But I guess to avoid cases like these it might be sensible to add it anyway.

@nobbele, @IGBC what do you think about that?

nobbele:
Can't the resize spam be fixed by buffering the scale factor change (storing it in a variable in context or something) and only applying it once the window has been dropped (not dragging).
As for the scaling itself, I would prefer the positions/sizes/scales being preserved over the different monitors and not breaking existing games than the window size being preserved. If anything, it should be implemented by the game itself, maybe using a helper method or an option
If I understand this issue correctly that is.
How do other applications do a smooth transition between monitors though? (they do right?)

PSteinhaus:
> Can't the resize spam be fixed by buffering the scale factor change (storing it in a variable in context or something) and only applying it once the window has been dropped (not dragging).

That too is a good idea! It's just the question whether we want to go this far (adding a otherwise useless variable to context (or more likely one of the sub-contexts)), to solve this problem. But it could be a pretty decent solution.

> As for the scaling itself, I would prefer the positions/sizes/scales being preserved over the different monitors and not breaking existing games than the window size being preserved. If anything, it should be implemented by the game itself, maybe using a helper method or an option

True. We could somewhat implement it for our users directly by calling `set_screen_coordinates` for them with adapted width and height calculated from the new scale though. But this would not account for mouse coordinates or other things that might break when the screen coordinates are changed...

> How do other applications do a smooth transition between monitors though? (they do right?)

It's simple. They leave this task to the OS and the OS manages to do it... somehow.
We, in constrast, fight the OS by calling functions to ensure that the sizes we specify actually reflect the physical size of our window.

PSteinhaus:
Having thought a bit more about it I now think that nobbeles solution for #949 is the best we can do right now.
My suggestion for implementing it:
- in `process_event` in the `ScaleFactorChanged` event case check whether the window is currently being dragged; if true don't apply the immediate correction (that being setting `new_inner size` to the current value)
- instead, in this case, we let `new_inner_size` be whatever it wants to be, so that we don't get a resize spam
- then, whenever a dragged window is released, we just always check whether `drawable_size` equals what is set in `window_mode` and if not we resize the window to again be as big as stated in `window_mode`

The problem with this: How do we know whether the window is currently being dragged? As far as I can see there is no way to check for that in winit, or did I miss something?

nobbele:
that's true, looking at the logs though it seems like changing the scale on a move event should at least better it.

PSteinhaus:
Ok, so the last problem that we'd want to solve here (that being this annoying resize spam) is definetely a winit issue (or even something deeper).
So before we go ahead and fix it I do need to ask: Do we really want to implement a workaround for this? Or do we just want to lable it a dependency bug and move on?

PSteinhaus:
> [#949 (comment)](https://github.com/ggez/ggez/pull/949#issuecomment-854707123)
> 
> just triggered too many events.

The thing is, all these events should in fact do nothing at all, since we set the new size to be the old size, so ideally no resizing should happen at all.

Apart from that it's of course also true that there's no reason for so many events to be created. That's another bug in itself.

src/event.rs
```diff
@@ -353,8 +353,13 @@ pub fn process_event(ctx: &mut Context, event: &winit::event::Event<()>) {
...
             }
-            winit_event::WindowEvent::ScaleFactorChanged { .. } => {
-                // Nope.
+            winit_event::WindowEvent::ScaleFactorChanged { new_inner_size, .. } => {
+                // actively set the new_inner_size to be the desired size
+                // to stop winit from resizing our window
+                **new_inner_size = winit::dpi::PhysicalSize::<u32>::from([
+                    ctx.conf.window_mode.width,
+                    ctx.conf.window_mode.height,
+                ]);
```
PSteinhaus, line 362: > Do we really want to implement a workaround for this? Or do we just want to label it a dependency bug and move on?

I, for one, think we should add a configuration option like `resize_on_screen_scale_factor_changed` to some sub-context which is checked before executing this.

If set to true, then this fix doesn't get executed and ggez will show the old behavior (i.e. ggez will give you a more "pleasing" screen change behavior, which might break your game if you don't plan for it).
Therefore this value should by default be set to false.

nobbele (approved):

PR #950, opened 2021-06-04
Add MacOS explaination to building guide.

### LINKED ISSUE #900: Mac Developers - ensuring all requirements are met
While working on a simple process (I *love* this library, by the way; the class documentation is fantastic), I noticed that my TravisCI instance was failing its builds:

`error: failed to run custom build command for alsa-sys v0.1.2`

However, I also realized I didn't directly specify this crate/lib in my Cargo.toml, and am using:

`[dependencies]

  rand = "0.7.0"
  structopt = "0.3.21"
  ggez = "0.5.1"
  glam = { version = "0.12", features = ["mint"]}
`

If I add alsa-sys to my dependencies, a cargo build freaks out.  

The page about building for every platform does not mention Mac OS developers: https://github.com/ggez/ggez/blob/master/docs/BuildingForEveryPlatform.md

Do we have any advice so I can fix travis (I can apt install before running in the config, that should not be a problem) but more importantly, ensure that I have all the tools necessary for development on a Mac?  

I'm happy to document my process/progress and keep the mentioned blob updated for every platform, too!

Thanks,

-Thomas

icefoxen:
Resolves #900 ; issues like that are best considered bugs in the documentation.  They looked for info in the correct spot and it wasn't there, so the program we wrote for humans to execute was incomplete.

### DIFF
--- a/docs/BuildingForEveryPlatform.md
+++ b/docs/BuildingForEveryPlatform.md
@@ -81,6 +81,13 @@ Should just build.  We recommend using the MSVC toolchain whenever possible, the
 
 Just copy-paste the exe and resource directory to the destination computer.
 
+# MacOS/iOS
+
+At the moment we unfortunately cannot provide support for developing on
+mac (none of the developers have hardware). While `ggez` tries not to do
+anything that would make it *not* work on Mac, it is a DIY target. If
+you resolve this issue feel free to open a PR.
+
 # Android
 
 Not officially supported yet. ;_; See https://github.com/ggez/ggez/issues/70


jiyinyiyong:
sad to hear that. I thought that many developers prefer using mac(me, js developers) for POSIX with stable GUIs, you all hate mac?

nobbele:
I like mc os generally despite its major flaws. I just don't want to spend 10x the money on a device that actually runs it, hackintoshes are too buggy.

jiyinyiyong:
unfortunately Rust survey didn't collect OS usages https://blog.rust-lang.org/2020/12/16/rust-survey-2020.html .

I should try Linux GUI somehow, after 6+ years.

icefoxen:
I dislike the closed hardware, the way Apple treats third party developers, and the arbitrary decisions they make about what APIs to support. 

Despite all that I'd do my best to support Mac, and did for a long time, but things kept being broken due to decisions Apple made, so I stopped having the time and energy to support a platform that was uninterested in supporting me back.

PR #952, opened 2021-06-20
Fixes typo and clippy warnings

remilauzier:


PSteinhaus:
Hi, thanks for giving this a try.
But to be honest I'm sorry, but I don't think we'll merge this, since these fixes (not sure about typo, but clippy for sure) have already been applied to the devel branch, [which is where all the work happens usually](https://github.com/ggez/ggez/blob/master/CONTRIBUTING.md).

Issue #800, opened 2020-07-21
EventHandler Error handling


OvermindDL1:
The EventHandler hardcodes a GameError error type even though `event::run` does nothing more then return the error if it happens from the (very few, why not all?) callbacks on EventHandler.  The GameError trait, though useful for errors from the engine, are useless for user generated errors, and currently the user seems to not be able to delegate to a main error handler, but instead has to pepper error handling in every single one of their EventHandler callbacks.

It would be far more simple for users, and for more usual Rust, to just use the `std::error::Error` trait for error handling.  That way GameError can be passed in to it, as can any other user error type, then they can be uniformly handled in a single place when a fatal error has occurred.  This would significantly reduce the amount of needed error handling code used.

GameError already implements `std::error::Error` so it's already half-way there, just need to fix up the callbacks as well such as EventHandler.

This would significantly improve usability around error handling, and would fix one of the biggest nits that I keep encountering that forces me to write excess code.

OvermindDL1:
As a secondary aside, the `GameError`'s implementation of `std::error::Error` is significantly lacking (most error details are just vanished, which is really really bad for debuggability...).  I would recommend using the `thiserror` library to simplify its implementation while ensuring that it is comprehensive and correct (plus it doesn't leak into the public API, it's just a derive).

icefoxen:
I'd really like to do this, but it's surprisingly tricky to coalesce multiple error types together in a way that the `std::error::Error` trait likes.  I've tried to handle it a bit in the simple cases in commit cef04ec30572e649b67b2f5824f6c765f9be906b

nobbele:
Can we just add `eprintln!("Error on EventHandler::update(): {:?}", e);` for update and draw beside the error! call?

PSteinhaus:
I'm not sure I understand the issue here.
Is the idea that we want to return something more generic instead of GameResults, so that users can handle their own custom Errors together with any GameErrors that update or draw may throw? I haven't yet gotten into touch with Rusts error handling really, sry.

nobbele:
Currently errors just fail silently unless you attach a logger

nobbele:
Actually this issue might be unrelated lol

PSteinhaus:
Now that error handling inside EventHandler has been improved the last thing to address is probably whether to use `thiserror` to rewrite most of the `GameError` implementation.

[...]

PR #957, opened 2021-07-01
EventHandler Error handling

PSteinhaus:
This is a first shot at fixing #800.
Instead of having to write error handling wrappers for each EventHandler function one can now just implement `EventHandler::on_error` to achieve customized error handling.

This still doesn't give users much freedom to handle their own kinds of errors though. Their only option for this, at this point, is to create a `GameError::CustomError(String)`, which is much better than nothing, but not ideal yet.

### DIFF
--- a/examples/01_super_simple.rs
+++ b/examples/01_super_simple.rs
@@ -17,7 +17,7 @@ impl MainState {
     }
 }
 
-impl event::EventHandler for MainState {
+impl event::EventHandler<ggez::GameError> for MainState {
     fn update(&mut self, _ctx: &mut Context) -> GameResult {
         self.pos_x = self.pos_x % 800.0 + 1.0;
         Ok(())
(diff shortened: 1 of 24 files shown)


nobbele:
Should the error! and eprintln! calls be kept in the event loop? It's already called in the default impl of on_error so it would give double prints.

IMO error! should be kept in the event loop and removed from default on_error so only eprintln! is in on_error and not the event loop.

PSteinhaus:
The only reason I split them up into two calls, one in the event loop and one in `on_error`, is that I thought it would be nice to still know where the error is coming from, which can only be done in the event loop, while one may not want to print the whole error (for some reason, but maybe that's not the case), so I moved that part into `on_error`. My thought was basically "if we want to give the user full control over error handling, then why not".

So you would leave error! in the event loop, but remove it from `on_error`? And if I understand correctly you want to remove eprintln! from `on_error`, so that it's only in the event loop? Could you explain why?

nobbele:
You correct about my thoughts on all users being logged but user decides on printing. User does have full control of it if it's split like I said, the logger can be implemented however they want.

As for the origin, I am not sure. Does it even need to be known if it already says it in the log? We could also add another parameter with an enum Origin { Update, Draw }.

PSteinhaus:
Hmm, I like that idea!

nobbele (approved):

Issue #958, opened 2021-07-05
Possible memory leak on to_rgba8


fairbairn:
Grabbing a snapshot of canvas seems to create a memory leak over time, as if something internal to the function is not released.

The loss seems to be around 50-60k per second.

We stripped back our use case to the following...
update does nothing
draw does the following..

```rust
    fn draw(&mut self, ctx: &mut Context) -> GameResult {
        self.capture(ctx);
        self.render(ctx);
        graphics::present(ctx);
        Ok(())
    }
```

We have the following struct..our state which holds a stage.. the state maintains the canvas that the stage draws into..
We grab a capture of that canvas before allowing it to draw..
If we don't perform the capture function, everything works  If we do, we leak memory.

```rust
pub struct State {
    pub stage: Stage,
}

impl State {
    pub fn new(ctx: &mut Context) -> State {
        let canvas = Canvas::with_window_size(ctx).unwrap();

        State {            
            canvas,                  // off-screen canvas
            stage: Stage::new(ctx), 
        }
    }
    
    fn capture(&mut self, ctx: &mut Context) {
        &self.stage.capture_stage(ctx, &self.canvas);
    }

    /// renders the off-screen canvas to the on-screen canvas
    fn render(&self, ctx: &mut Context) {
        let params = DrawParam::new().dest([0.0, 0.0]);
        graphics::set_canvas(ctx, None);
        graphics::draw(ctx, &self.canvas, params).unwrap();
    }
}

pub struct Stage {
    capture_video: Arc<RwLock<Vec<u8>>>,
}

impl Stage {
    pub fn new(ctx: &mut Context) -> Stage {
        let canvas = graphics::Canvas::with_window_size(ctx).unwrap();
        let capture_video = Arc::new(RwLock::new(canvas.image().to_rgba8(ctx).unwrap()));

        Stage {
            capture_video,
        }
    }

    pub fn capture_stage(&mut self, ctx: &mut Context, canvas: &graphics::Canvas) {
        let screen = canvas.image().to_rgba8(ctx).unwrap();    // <<<<<<<<<<<<<<<<<<<<<<<<< THIS CAUSES A LEAK

        // let mut bytes = self.capture_video.write().unwrap();
        // *bytes = screen;
        // drop(bytes);

    }
}
```

This should just work all day long, not really do anything, and `screen` should drop the used memory.

If we allow the `let screen = canvas.image().to_rgba8(ctx).unwrap();` to run without our update or draw event, then memory leaks continuously.

[...]

PR #961, opened 2021-07-09
Fixes memory leak in to_rgba8 and screenshot

PSteinhaus:
This fixes the memory leak by just using the global command buffer instead of creating a "local" new one, which is leaky.

I also added a download buffer handle to the gfx context, to avoid possibly recreating it all the time.
This is probably not important, but it's the first thing I tried and it works and it probably improves performance for use cases like #958, so why not keep it.

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -28,6 +28,7 @@ where
     pub(crate) projection: Matrix4,
     pub(crate) white_image: ImageGeneric<B>,
     pub(crate) screen_rect: Rect,
+    pub(crate) to_rgba8_buffer: gfx::handle::Buffer<B::Resources, u8>,
     color_format: gfx::format::Format,
     depth_format: gfx::format::Format,
     srgb: bool,
@@ -219,6 +220,8 @@ impl GraphicsContextGeneric<GlBackendSpec> {
 
         quad_slice.instances = Some((1, 0));
 
+        let to_rgba8_buffer = factory.create_download_buffer::<u8>(1)?;
+
         let globals_buffer = factory.create_constant_buffer(1);
         let mut samplers: SamplerCache<GlBackendSpec> = SamplerCache::new();
         let sampler_info =
@@ -283,6 +286,7 @@ impl GraphicsContextGeneric<GlBackendSpec> {
             projection: initial_projection,
             white_image,
             screen_rect: Rect::new(left, top, right - left, bottom - top),
+            to_rgba8_buffer,
             color_format,
             depth_format,
             srgb,
--- a/src/graphics/image.rs
+++ b/src/graphics/image.rs
@@ -199,20 +199,18 @@ impl Image {
         let w = self.width;
         let h = self.height;
 
-        // Note: In the GFX example, the download buffer is created ahead of time
-        // and updated on screen resize events. This may be preferable, but then
-        // the buffer also needs to be updated when we switch to/from a canvas.
-        // Unsure of the performance impact of creating this as it is needed.
-        // Probably okay for now though, since this probably won't be a super
-        // common operation.
-        let dl_buffer = gfx
-            .factory
-            .create_download_buffer::<u8>(usize::from(w) * usize::from(h) * 4)?;
+        let format = gfx.color_format();
 
-        let factory = &mut *gfx.factory;
-        let mut local_encoder = GlBackendSpec::encoder(factory);
+        let dl_buffer = &mut gfx.to_rgba8_buffer;
+        // check if it's big enough and recreate it if not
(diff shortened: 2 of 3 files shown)


nobbele (commented):
Wouldn't it be better to have a second encoder in the gfx context rather than using the global one?

PSteinhaus:
But what for?
Is there any situation where flushing the "global" command buffer (meaning the one stored in the gfx context) too early can have any unwanted side effects?

nobbele:
encoding a canvas mid-render would produce an unfinished image, but now that I think about it that might be the expected behavior rather

nobbele (approved):

fairbairn:
Not yet certain the leak is fixed.  We're going to re-run our test bench now with that one line removed and evaluate.

nobbele:
@fairbairn you can try their branch
`ggez = { git = "https://github.com/PSteinhaus/ggez" }`

fairbairn:
@nobbele That repo seems to be compatible with v5 instead of v6.rc2/master.  It had conflicts suggesting we'd have to backport.
We'll wait until it merges into master.

PSteinhaus:
I just realized, since we now flush whenever we use `to_rgba8` we can remove that extra flush from `Canvas::to_image`.

fairbairn:
Testing now...

We still use raw_image() to bypass the flush and manual flip.  The canvas to_rgba8() call is still too slow for our use case...likely because of the flip.

If we draw to an off-screen canvas, and capture before present(), or a flush (which we may have to do manually) are we out of sync with the latest draw?

Could we do a present() while our canvas is active and before the capture to insure a current snap?

We haven't tried this yet, but it would be nice to assure that we're capturing the most recent draw and not the one 33ms prior.  We may have to do some single frame tests to understand the current behavior. 

Be nice if there was a separate flush method on canvas though.

We'll report back with findings.

fairbairn:
Memory leak resolved.  Great work @PSteinhaus.

PSteinhaus:
> If we draw to an off-screen canvas, and capture before present(), or a flush (which we may have to do manually) are we out of sync with the latest draw?

With this PR merged capturing a canvas (via `to_rgba8` on the canvas itself, its internal image, or an image created from it) means adding a texture copy command to the (global) command buffer and then flushing it.
So if you use `to_rgba8` (or `screenshot`) then your capture will always be up to date to the latest draw IF you place it at the end of your draw function.
If you capture before that it will of course be up to date to the previous draw. And if you place the capture code somewhere inbetween other draw calls you'll get a middle-stage/a half drawn image.

> Could we do a present() while our canvas is active and before the capture to insure a current snap?

If you mean "could we add a second `present` call?", then yes, but please don't do that.
It would work to ensure a current snap, but having more than one `present` call in your draw function creates flickering images.

You don't actually need to do anything to make sure the snap is current. The reason for this is: Creating a copy of a texture always means putting additional commands in the draw command buffer and then (at some point) flushing it to actually get it done.

So to get the current canvas image and avoid a second flush the ideal solution from my point of view would be [to go this way](https://github.com/ggez/ggez/issues/958#issuecomment-876712273). Reimplement `to_rgba8`, but without the flush and put that code at the end of your draw function, but before the `present` call.
The effect: Your capture command is treated like any other draw call and processed together with them when `present` is called. The capture command being the last thing in the command buffer ensures the creation of a current image.

PSteinhaus:
> We still use raw_image() to bypass the flush and manual flip. The canvas to_rgba8() call is still too slow for our use case...likely because of the flip.

If you're open to sharing your (apparently more performant) flipping code, so we can replace ours with it, that would be great.

On the other hand: How sure are you that it's really the flipping code and not the cost of an additional flush?
Honestly I don't think having two or three flushes instead of one should be very costly, but `Canvas::to_rgba8` currently still has an unnecessary flush at the beginning (which I'll remove right now) and then there's another flush inside `Image::to_rgba8` (which is called by `Canvas::to_rgba8`), so you actually end up with two additional flushes right now when using `Canvas::to_rgba8` versus one additional flush when using `Image::to_rgba8` on `raw_image` directly.

fairbairn:
Thanks, we'll re-pull and test again with latest.

In terms of our flipping code, I do not think it is necessarily more performant.  We actually do that downstream in our encoder, so there's no code in Rust to share.  It could very well be the flush, hard to know.

We were going to just write a simple routine of our own and use the flush calls that were clearly documented in both screenshot and to_rgba8, but the gfx-related methods are private and we'll have to do some more work to build and test a suitable routine with an optimized flush.

nobbele:
gfx related methods should be publicly usable with `graphics::gfx_objects()`

PR #963, opened 2021-07-17
Allow SoundSources to be skipped by a Duration.

a2aaron:
This functionality is useful if you need to skip ahead in an audio file some amount before playing it (for example, maybe there is a cutscene in your game that can be skipped, so the music must also skip ahead by some amount). There isn't a lot to this PR, I'm mostly just exposing more of Rodio's functions.

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -165,6 +165,9 @@ pub trait SoundSource {
     /// Sets the speed ratio (by adjusting the playback speed)
     fn set_pitch(&mut self, ratio: f32);
 
+    /// Sets the amount of time to skip in the sound
+    fn set_skip_amount(&mut self, dur: time::Duration);
+
     /// Gets whether or not the source is set to repeat.
     fn repeat(&self) -> bool;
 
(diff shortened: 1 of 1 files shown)


nobbele:
I am unavailable to test or merge snything currently. Does elapsed time work?

a2aaron:
After some investigation, it seems that this might not actually be a great PR? It seems that the way that rodio handles skipping ahead in a track is by consuming samples in `current_frame_len` units until reaching the desired duration. In practice that means skipping ahead ~30 seconds means a noticeable lag when playing something (you usually end up consuming millions of samples in 1000 sample chunks). Hence maybe its not appropriate for a real-time application like a game? I'm gonna close this for now I think.

IGBC:
I have been lurking on this one, and I think one way of achieving it would be to add some functions to support creating a sound sample from a file with a N sample offset. 

that sample will be a reference that is loaded and consumed by N samples at load time and then can be cloned and played as needed.

PR #968, opened 2021-08-03
Make most of the `filesystem` functions take `&Context`.

### LINKED ISSUE #967: Most of the functions in `filesystem` should take `&Context`, not `&mut Context`
It seems that a bunch of the functions in the `filesystem` module take `&mut Context` as a parameter, even though that this isn't actually necessary. Among them are:

- `create`
- `create_dir`
- `delete`
- `delete_dir`
- `read_dir`
- `open`
- `open_options`
- `print_all`
- `log_all`
- `read_config`
- `write_config`

I do not see a reason why these functions should require an `&mut Context`, since the underlying `vfs` methods it calls also do not require `&mut OverlayFS`. It would be nicer if they all too `&Context`, since this means it will work in places where you need multiple reference to the same context (for example, if you have a bunch of closures needing to use `open` or `read_dir`)

a2aaron:
Most of these functions do not need to take `&mut Context` and can simply
take `&Context` instead. This is helpful since there are some scenarios
where you want `&Context`s, such as when passing the same context to a
bunch of closures that all want to read a file and don't need mutability.

Fixes #967

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -183,7 +183,7 @@ impl Filesystem {
 
     /// Opens the given `path` and returns the resulting `File`
     /// in read-only mode.
-    pub(crate) fn open<P: AsRef<path::Path>>(&mut self, path: P) -> GameResult<File> {
+    pub(crate) fn open<P: AsRef<path::Path>>(&self, path: P) -> GameResult<File> {
         self.vfs.open(path.as_ref()).map(|f| File::VfsFile(f))
     }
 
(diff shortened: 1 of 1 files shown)


nobbele:
can you base it on and merge into devel instead?

a2aaron:
Yes, I can do that.

PSteinhaus:
It seems you need to fix some tests to fit your changes, but then it should be good to go.

a2aaron:
Ah, looks like Clippy is complaining about unnecessary `mut`s. Fixed I think!

nobbele:
Looks fine, thank you!

PR #970, opened 2021-08-07
Allow renaming of resources and resources.zip

MatchaChoco010:
Hi!

This is a PR for an additional feature as the title suggests.

I wanted to change the name of `resources.zip` to `data`.
I personally think it's cooler to have `data` with no extension next to the exe than to have a file with a `.zip` extension next to it.
Of course, I understand that hiding the extension does not lead to any kind of encryption.
It's just to make it look cool, nothing more.

I added `resources_dir_name` and `resources_zip_name` as arguments when building the filesystem.
I added a method to ContextBuilder to change this setting.
The default values are set to `resources` and `resources.zip`, which are the same as before, respectively.

As a limitation, unfortunately, this setting cannot be written in `conf.toml`, because `conf.toml` must be in resources, and can only be used after the resources directory has been determined.
Therefore, this setting can only be changed via the `ContextBuilder` method.

I understand that more configuration increases complexity and may threaten "make good games, easily".
However, I believe that adding this setting will not add too much complexity since the default setting is to use the existing `resources` ans `resources.zip`.
If there are any problems, I will withdraw this PR.

Thank you for your patience.
Finally, thanks for the awesome crate!

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -130,6 +130,8 @@ pub struct ContextBuilder {
     pub(crate) game_id: String,
     pub(crate) author: String,
     pub(crate) conf: conf::Conf,
+    pub(crate) resources_dir_name: String,
+    pub(crate) resources_zip_name: String,
     pub(crate) paths: Vec<path::PathBuf>,
     pub(crate) memory_zip_files: Vec<Cow<'static, [u8]>>,
     pub(crate) load_conf_file: bool,
@@ -142,6 +144,8 @@ impl ContextBuilder {
             game_id: game_id.to_string(),
             author: author.to_string(),
             conf: conf::Conf::default(),
+            resources_dir_name: "resources".to_string(),
+            resources_zip_name: "resources.zip".to_string(),
             paths: vec![],
             memory_zip_files: vec![],
             load_conf_file: true,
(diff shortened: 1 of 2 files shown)


PSteinhaus:
Thanks for this PR!

This might be somewhat of a niche feature, but spontaneuosly I don't see any reason to disallow it.

I also don't have a problem with bloating up the ContextBuilder a little to allow for it. It's shortlived enough and since you honored the usual syntax of setting options for it I don't think it increases complexity much. Sure, the documentation becomes a bit longer, but your functions are named pretty self explanatory and the documentation looks nice and short, so why not.

The only thing that confuses me a little are the clippy failures...
Clippy doesn't have any problems with devel, but for some reason it suddenly feels very annoyed about a bunch of stuff you didn't even touch...

nobbele:
maybe remove the change_ prefix?

PSteinhaus:
> maybe remove the change_ prefix?

Right, that's a bit redundant / incoherent with the other method names.

Nobbele was so kind to fix the clippy errors on devel, so if you update your PR to the latest devel version they should be gone here as well.

MatchaChoco010:
OK! I'll merge the latest devel branch and remove the change prefix :+1:

nobbele (approved):

PR #971, opened 2021-08-07
Fixed an issue that prevented opening resources.zip files deeper than one level in Windows environments.

MatchaChoco010:
Fixed a problem that prevented opening files deeper than one level in resources.zip on Windows.

When reading a Zip file, it was using the `to_str()` string of `PathBuf` as index.
However, on Windows, `to_str()` on a `PathBuf` would use `\\` as the separator.
For example, `foo\\bar.txt`.

This prevented us from reading the contents of the resources.zip file more than one level deep under Windows, which would require a separator.

In this PR, when reading zip files, instead of using `to_str()` in `PathBuf`, using  `sanitize_path_for_zip`, which a modified version of `sanitize_path`, to generate a `/` delimited path using a string.

This makes it possible to use resources.zip in Windows environments.

### DIFF
--- a/src/vfs.rs
+++ b/src/vfs.rs
@@ -213,6 +213,38 @@ fn sanitize_path(path: &path::Path) -> Option<PathBuf> {
     Some(accm)
 }
 
+/// We need to return a string.
+/// The reason is that the path in zip is `/` delimited, but the path
+/// delimiter in rust is environment dependent.
+/// For example, on Windows, `PathBuf` to str in rust make "foo\\bar.txt".
+fn sanitize_path_for_zip(path: &path::Path) -> Option<String> {
+    let mut c = path.components();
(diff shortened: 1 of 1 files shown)


nobbele:
I don't get what the problem you are having is, using backslash works fine for me on windows `graphics::Image::new(ctx, "\\g\\t\\tile.png")` PathBuf also works

MatchaChoco010:
Hmmm.
Here's a repository that reproduces it minimally at my end.
https://github.com/MatchaChoco010/ggez-windows-path

I'm sorry if this is a problem with my personal environment.

nobbele:
ah wait it's zip, I misunderstood the issue

nobbele:
could you add some test cases to the function?

Issue #910, opened 2021-03-09
Animation


PSteinhaus:
The discussion in #875 about the direction of ggez made me remember what I'd really like ggez to have one day!
### Animation
Animation is a crucial feature for most games (and one asked for in #235 by multiple people), so I think it's important to finally provide some sort of framework for it.
This could be an expansion of the API, but it could also (maybe) just be a simple example showcasing how to achieve animation through Spritebatches and DrawParams with source rects shifting for every animation step (but only if we come to the agreement that this is how animation should be handled in ggez).

I personally favor the latter approach, since it would keep ggez simple to learn while giving those who seek for them the necessary tools for handling simple animation, instead of forcing them to invent the wheel anew.

If you agree I would start a PR adding such an example. But I'm really interested in other opinions here, since this is just my kind of naive approach to animation in ggez and there might be better.

nobbele:
The way I did animations in my ggez game was have like
struct Game {
    fading_box: Tween<Sprite<Rect>>, // ggez doesnt have a type with both a drawable and a drawparam for some reason so thats what sprite is
}

Game {
    fading_box: Tween {
        current: Rect { x: 0, y: 0, w: 100, h: 100 }
        time: 10.0,
        func: Box::new(|target: &mut Rect, progress: f32| { target.x = 100*progress }) //progress goes from 0.0 to 1.0
}

makes x go from 0 to 100 in 10 seconds

PSteinhaus:
> Couldn't such a thing be implemented with tween though? |current: &mut DrawParam, progress: f32| { *current = sourceRects[(progress * sourceRects.len() as f32) as usize] } or something

You're right. It's probably a good idea to just use a tween for that. We would also need some kind of helper function though, to scale the source rects according to the image used in the spritebatch, so I'd still create and use a dedicated Animation struct as shown above, but passing a Tween to an Animation to control how progress gets mapped to source rects is probably a good idea.

> about the tween implementation you mentioned, would that be like iterator in that you could do tween::linear::().sin() to get Sin<Linear>

Hmm. Yes, we could combine them that way, but honestly I'd rather just have ```tween::linear()```, ```tween::sin()```, ... , and then a ```tween::custom(|t| sin(t) /* or whatever combination you want here */)``` for more complex cases.

nobbele:
I just coded something up an idea I had
```rust
let mut tweener = Tweener {
    data: 0.5,
[...]

PR #975, opened 2021-08-17
Added an example showcasing animation with keyframe

PSteinhaus:
Based on the discussion in #910 I finally made an example showcasing both tween and frame-by-frame animations using keyframe.

It... works, but there's some code duplication, that I'd really like to get rid off.
But after fiddling with generics for some time I now think the only option would be to use macros to do that, but I'm unsure as to whether we'd want to do that, since we probably don't want to force new (Rust) users to read macro code in order to understand the example.

But maybe I'm missing something.


PSteinhaus:
Oh, and it seems I also need a workaround for all these enum variants never being "officially" constructed anywhere...

PSteinhaus:
Ok, now that I was able to get rid of the code duplication I'm satisfied with the example.
It's still somewhat long, but it's also a bit shorter than examples like `shadow.rs` or `snake.rs`, so I think it's alright.

nobbele:
what's with the transmutes, doesn't `my_enum as i32` work if you just assign a number to the variants

PSteinhaus:
> what's with the transmutes, doesn't `my_enum as i32` work if you just assign a number to the variants

It does, but the other way around doesn't work. So I thought it's a bit more explicit to use transmute both ways.
I could also change it to `as`` and refactor the transmute into a `From<i32>` implementation, for both enums. Or we could use the `num_derive` and `num_traits` crates.

nobbele:
adding num_traits and num_derive to dev dependencies should be fine

PSteinhaus:
Ok, done.

nobbele (approved):

PR #982, opened 2021-09-10
Add methods to obtain references to internal sprite data in SpriteBatch

Rapptz:
There is currently no way, as far as I know, to remove elements from the sprite batch and this would give a decent way of doing that.

This is modeled after the same methods in MeshBatch, docstring and all. The naming could probably use some bikeshedding.

### DIFF
--- a/src/graphics/spritebatch.rs
+++ b/src/graphics/spritebatch.rs
@@ -89,9 +89,9 @@ impl SpriteBatch {
 
     /// Returns a mutable reference to the sprites.
     ///
-    /// Please note that manually altering items in this slice
-    /// will not automatically invalidate the buffer, you will
-    /// have to manually call `flush()` or `flush_range()` later.
+    /// Unlike with `MeshBatch`, manually calling `flush` after altering sprites
+    /// in this slice is currently unnecessary, as `SpriteBatch` flushes automatically
+    /// on every draw call. This might change in the future though.
     pub fn get_sprites_mut(&mut self) -> &mut [DrawParam] {
         &mut self.sprites
     }
--- a/src/graphics/spritebatch.rs
+++ b/src/graphics/spritebatch.rs
@@ -82,6 +82,20 @@ impl SpriteBatch {
         }
     }
 
+    /// Returns a reference to the sprites.
+    pub fn get_sprites(&self) -> &[DrawParam] {
+        &self.sprites
+    }
+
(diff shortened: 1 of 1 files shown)


src/graphics/spritebatch.rs
```diff
@@ -82,6 +82,20 @@ impl SpriteBatch {
...
     }
 
+    /// Returns a reference to the sprites.
+    pub fn get_sprites(&self) -> &[DrawParam] {
+        &self.sprites
+    }
+
+    /// Returns a mutable reference to the sprites.
+    ///
+    /// Please note that manually altering items in this slice
```
nobbele, line 92: Is this note really necessary? It automatically gets flushed in the draw() call and clear() doesn't say anything about it either.

Rapptz: I copy pasted this docstring from MeshBatch, so I figured might as well do it for consistency there. I can remove it if it's unnecessary.

PSteinhaus: Manually flushing the spritebatch is really not necessary, as spritebatch currently does not feature a dirty bit like MeshBatch. Then again, we recently thought about adding one, so it might be worth to keep this in, just so that people don't stumble over this later.

I'm a bit torn. The best thing might be to change the comment, so that it says that manually flushing the spritebatch is currently not necessary, but might become necessary in the foreseeable future, due to a dirty bit that might be added to spritebatch internally.

nobbele (approved):

PSteinhaus:
> There is currently no way, as far as I know, to remove elements from the sprite batch and this would give a decent way of doing that.
> 
> This is modeled after the same methods in MeshBatch, docstring and all. The naming could probably use some bikeshedding.

I just realized: This doesn't give you a way to remove elements from the sprite batch, does it? I mean, you can't shorten the slice, you can't delete elements from it, so it still really doesn't give you any option to remove something, right?

Rapptz:
> I just realized: This doesn't give you a way to remove elements from the sprite batch, does it? I mean, you can't shorten the slice, you can't delete elements from it, so it still really doesn't give you any option to remove something, right?

Actually you're right here. Since the backing store is dynamically owned by the `Vec`, the slice doesn't provide a way to remove elements from it.

PSteinhaus:
> Though truthfully, this has no bearing on the PR IMO. A method to remove via `SpriteIdx` can be added later.

I agree.

PR #984, opened 2021-09-17
Fix conversion of linear color to sRGB

0----0:
Per https://en.wikipedia.org/wiki/Srgb#From_CIE_XYZ_to_sRGB , you have to subtract 0.055 from the color value after multiplying and exponentiating in order to properly invert the sRGB gamma function.  You can verify that the conversion is correct here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c2aedead960001921875514d371bea5e


PSteinhaus:
Thank you, somehow missed that one!

PR #986, opened 2021-09-18
Re-export winit

gorilskij:
Actually add a re-export of `winit` as this was previously missing. It's mentioned in `context.rs` that "we re-export winit" but this currently isn't the case. This PR just adds `ggez::winit`.


PSteinhaus:
Hah, thanks!
I remember thinking the other day "maybe we should just reexport winit". Didn't know that this has been already intended 😅

PR #991, opened 2021-10-31
implemented MSAA on canvases via a fragment shader workaround

PSteinhaus:
I followed the inspiration from https://github.com/gfx-rs/gfx/issues/1024 to implement a fragment shader workaround for MSAA on canvases.

It doesn't work on GLES though, as it doesn't have `sampler2DMS`, so the shader won't work. But it compiles and it doesn't crash. On my system multi-sampled canvases simply do not get drawn on GLES.

To write `Canvas::resolve` (which applies the fragment shader workaround) I had to lose `CanvasGeneric`, because `resolve` now needs to be called before `to_rgba8` to make sure the image is up to date and `resolve` needs to change the `out` pipe, which is part of `gfx_device_gl::Resources`, but not of `S::Resources` yet.

As we'll probably move on to `wgpu` in the future it shouldn't matter much though. When we do so we'll probably get rid of the generics anyway (as `wgpu` in itself is really generic enough already and noone wants to write a second graphics backend for ggez anyway, I think; if someone desires to do so, forking ggez might really be the more sane option anyway IMO).

If there was some elegant way though to keep `CanvasGeneric` and still implement `resolve` I'd of course prefer that.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -41,6 +41,8 @@ pub enum GameError {
     GamepadError(String),
     /// Something went wrong with the `lyon` shape-tesselation library.
     LyonError(String),
+    /// You tried to use MSAA on canvases with GLES, which isn't supported.
+    CanvasMSAAError,
     /// A custom error type for use by users of ggez.
     /// This lets you handle custom errors that may happen during your game (such as, trying to load a malformed file for a level)
     /// using the same mechanism you handle ggez's other errors.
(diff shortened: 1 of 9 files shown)


PSteinhaus:
I'm unsure whether the automatic calls to `resolve` are worth losing `CanvasGeneric`.

Maybe it's ok to not have them be automatic. The most important automatic call to `resolve` would still remain: the one inside `Canvas::draw`. For `Canvas::to_rgba8` (and the methods depending on it) it might be ok to force users to call `resolve` manually, if they need to update the texture manually at all that is, after all `Drawable`s do get drawn a lot usually anyway...

nobbele (approved):

PR #995, opened 2021-11-21
Touch event

ssciarini:
I needed access to the touch event for a project I am doing with a raspberry pi. Rather than writing a custom event loop, this PR will give developers access to the touch events in the same way that we can with mouse & keyboard events.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -198,6 +198,12 @@ where
         Ok(())
     }
 
+    /// An event from a touchscreen has been triggered; it provides the x and y location
+    /// inside the window as well as the state of the tap (such as Started, Moved, Ended, etc)
+    fn touch_event(&mut self, _ctx: &mut Context, _phase: TouchPhase, _x: f64, _y: f64) -> GameResult {
+        Ok(())
+    }
+
     /// A gamepad button was pressed; `id` identifies which gamepad.
     /// Use [`input::gamepad()`](../input/fn.gamepad.html) to get more info about
     /// the gamepad.
@@ -423,6 +429,12 @@ where
                 WindowEvent::Touch(touch) => {
                     let res =
                         state.touch_event(ctx, touch.phase, touch.location.x, touch.location.y);
+                    if catch_error(ctx, res, state, control_flow, ErrorOrigin::TouchEvent) {
+                        return;
+                    };
+                }
+                WindowEvent::Touch(touch) => {
+                    let res = state.touch_event(ctx, touch.phase, touch.location.x, touch.location.y);
                     if catch_error(ctx, res, state, control_flow, ErrorOrigin::TouchEvent) {
                         return;
                     };
--- a/src/event.rs
+++ b/src/event.rs
@@ -198,12 +198,6 @@ where
         Ok(())
     }
 
-    /// An event from a touchscreen has been triggered; it provides the x and y location
-    /// inside the window as well as the state of the tap (such as Started, Moved, Ended, etc)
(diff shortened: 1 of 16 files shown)


PSteinhaus:
I really like this PR. Adding a touch event is definitely a good idea.

Yet, I assume most of our users won't really use it much, as ggez currently doesn't have any mobile support. Still, some laptops do feature touchscreens so it still makes sense to have this. It will also become much more necessary once/if we take the step towards mobile support one day.

But for now I assume most users just want touch events to be handled like the mouse events they probably designed their application around. Due to this it would be nice to change the default implementation of `touch_event` to simulate mouse input, [like in good-web-game](https://github.com/ggez/good-web-game/blob/3ae0d23cd84b921647f1984643b71dab5cfc9f62/src/event.rs#L49-L77), I think.

For ggez the code we'd need to replicate should be [this](https://github.com/ggez/ggez/blob/50918f133785e1ee585c022d94a0176ea0e73887/src/event.rs#L352-L382).

ssciarini:
I can absolutely look in to making the touchscreen events emulate the mouse functionality!

PSteinhaus:
Thank you :)

It's not a strict necessity though. I plan on merging your PR anyway.
But if you could add that extra bit of functionality it would of course be very welcome ;)

src/event.rs
```diff
@@ -273,6 +277,9 @@ where
                     let delta = mouse::last_delta(ctx);
                     state.mouse_motion_event(ctx, position.x, position.y, delta.x, delta.y);
                 }
+                WindowEvent::Touch(touch) => {
+                    state.touch_event(ctx, touch.phase, touch.location.x, touch.location.y);
+                }
```
PSteinhaus, line 282: You were a bit unlucky with your timing though, as we just merged a PR extending error handling to all methods of `EventHandler`.

To make the new touch event fit that scheme you'd have to adapt this code like this:
```suggestion
                WindowEvent::Touch(touch) => {
                    let res = state.touch_event(ctx, touch.phase, touch.location.x, touch.location.y);
                    if catch_error(ctx, res, state, control_flow, ErrorOrigin::TouchEvent) {
                        return;
                    };
                }
```

Of course, you'd also have to add `TouchEvent` to the `ErrorOrigin` enum.

src/event.rs
```diff
@@ -131,6 +131,10 @@ where
     /// This is the intended way of facilitating text input.
     fn text_input_event(&mut self, _ctx: &mut Context, _character: char) {}
 
+    /// An event from a touchscreen has been triggered; it provides the x and y location
+    /// inside the window as well as the state of the tap (such as Started, Moved, Ended, etc)
+    fn touch_event(&mut self, _ctx: &mut Context, _phase: TouchPhase, _x: f64, _y: f64) {}
```
PSteinhaus, line 136: Also, this method should now return a `GameResult` as well:
```suggestion
    fn touch_event(&mut self, _ctx: &mut Context, _phase: TouchPhase, _x: f64, _y: f64) -> GameResult {
        Ok(())
    }
```

PSteinhaus (changes requested):

ssciarini:
Updated this with your suggestions.

I'll start a new branch with mouse emulation enabled after I do a bit more testing here and after updating the new code with the error handling changes. What are your thoughts on keeping the touch_event function in place? In theory it shouldn't be needed if the preference is having it trigger mouse events, but since it is getting merged, I wouldn't want to do something that would break functionality.

Thanks!

PSteinhaus (approved):

nobbele:
Seems nice

PR #998, opened 2021-12-10
Modified the input module somewhat

Ulfr26:
I've added some functions to the keyboard and mouse contexts that can detect whether a button is pressed/released on a certain frame (i.e., pressed this frame but released last frame and vice versa). I also changed the mouse state to be represented with a HashSet, just to keep it consistent with the keyboard.

This requires an extra variable in `KeyboardContext` and `MouseContext` that represents the state of the keyboard/mouse on the previous frame. Alternatively you could represent the state as a `HashMap<KeyCode, (bool, bool)>`, but I don't know if changing this each frame would be more or less efficient than just cloning to a new variable.

### DIFF
--- a/examples/eventloop.rs
+++ b/examples/eventloop.rs
@@ -79,6 +79,14 @@ pub fn main() -> GameResult {
                 // necessary because it's calculated cumulatively each cycle
                 ctx.mouse_context.reset_delta();
 
+                // Copy the state of the keyboard into the KeyboardContext and
+                // the mouse into the MouseContext.
+                // Not required for this example but important if you want to
+                // use the functions keyboard::is_key_just_pressed/released and
+                // mouse::is_button_just_pressed/released.
+                ctx.keyboard_context.save_keyboard_state();
+                ctx.mouse_context.save_mouse_state();
+
                 ggez::timer::yield_now();
             }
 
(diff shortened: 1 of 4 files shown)


PSteinhaus:
Thank you for this PR!

Glancing over it it looks fine to me. I'll try to take a closer look tomorrow ;)

I'm not 100% sure yet whether this change is really worth it though, as this dynamic kind of information is already kind-of supplied to the user through the EventHandler input callbacks. On the other hand we now have this kind of duplication in many places and I personally like it, as it gives me greater flexibility in how I use the input.

nobbele:
I like this

PSteinhaus:
I like it too.

One small thing that would need to be changed though is that `save_keyboard_state()` and `save_mouse_state()` need to be public, sadly, because we need to give users who are running their own event loop the ability to call them; same situation as with `reset_delta`.
Then, these calls would need to be added to the eventloop example as well, to keep it in sync with our default one.

What I really like is the change from HashMap to HashSet, that's simply better, I think 👍 

I don't really think that going through a `HashMap<KeyCode, (bool, bool)>`, copying each 'current' bool onto the 'previous' one, would be more efficient than just cloning the HashSet, as currently done, but I also hope that the difference wouldn't be too big.

My mind tries to convince me that there has to be some clever buffer-swapping-idea to get around the copying completely, but I guess that'd never work, as both states need to be equal at the start of a frame.

Ulfr26:
No worries, I've made those public now :+1:

nobbele (approved):

PR #999, opened 2021-12-12
Fix zip `read_dir` deeper than one level in Windows

MatchaChoco010:
Windows uses `\\` instead of `/` as the delimiter for Path, but since the index of zip returns the `/` delimiter, it did not detect the existence of the directory correctly when compared with `starts_with()` in `read_dir()`.
By using `sanitize_path_for_zip()`, `\\` is converted to `/` and the path is sanitized.

Also fix that the directory passed as an argument to `read_dir()` is not included in the iterator to make it work the same as NonZip.

In NonZip's read_dir, `read_dir("/foobar/")` returns `/foobar/file.ext` and so on.
In Zip's read_dir, when `read_dir("/foobar/")` is used, it returns `foobar/file.ext` without the leading `/`.
Zip and NonZip don't match, so the Zip process is now adapted to NonZip and returns the file with `/` at the beginning.

### DIFF
--- a/src/vfs.rs
+++ b/src/vfs.rs
@@ -811,12 +811,15 @@ impl VFS for ZipFS {
     /// Zip files don't have real directories, so we (incorrectly) hack it by
     /// just looking for a path prefix for now.
     fn read_dir(&self, path: &Path) -> GameResult<Box<dyn Iterator<Item = GameResult<PathBuf>>>> {
-        let path = convenient_path_to_str(path)?;
+        let path = sanitize_path_for_zip(path).ok_or_else(|| {
+            let errmessage = format!("Invalid path format for resource: {:?}", path);
+            GameError::FilesystemError(errmessage)
+        })? + "/";
         let itr = self
             .index
             .iter()
-            .filter(|s| s.starts_with(path))
-            .map(|s| Ok(PathBuf::from(s)))
+            .filter(|&s| s.starts_with(&path) && s != &path)
+            .map(|s| Ok(PathBuf::from("/").join(s)))
             .collect::<Vec<_>>();
         Ok(Box::new(itr.into_iter()))
     }


PSteinhaus:
Thank you so much for this PR.

Ever since you started addressing this in your previous PR I had the nagging feeling that the whole Zip path problematic might be worse than I thought, but I just couldn't find the time and motivation to actually look into it properly. Therefore, thank you : )

I'll try to test it soon.

PSteinhaus:
Ok, I tested it and it works.
But the thing is... it already works with current devel as well...

Or am I missing something here? I tried to load resources from directories deeper than one level in both normal directories and zip files and in both cases it worked just fine (here on Windows 10).

MatchaChoco010:
Oh, my God, I'm sorry about that.
Maybe I was not using the latest devel?
Or maybe I'm using Windows 11 and that's the problem.
I'll check it again in my environment over the weekend and report back.
Please wait until the weekend.
Thank you very much for your confirmation.

MatchaChoco010:
By the way, I have confirmed that the `open()` method of filesystem can retrieve resources from directories deeper than one level, even with devel prior to this PR.

What did not work well in my environment is the following `read_dir()` method.
https://github.com/MatchaChoco010/ggez/blob/de0a4a04adede047909cbd0f61e81f4a63d48a82/src/filesystem.rs#L260
https://github.com/MatchaChoco010/ggez/blob/de0a4a04adede047909cbd0f61e81f4a63d48a82/src/vfs.rs#L813

I remember that the `read_dir()` method did not return the path correctly...

Whatever it is, I will investigate it again over the weekend and report back.

MatchaChoco010:
Hello!

I have verified on my Windows 11 that `read_dir()` is not working correctly for zip files.
I uploaded a project to GitHub that reproduces the issue.
https://github.com/MatchaChoco010/ggez_zip_read_dir_test
I hope this repository will be useful in your investigation.

PSteinhaus:
Thank you. I'm sorry, now I can reproduce the issue and can confirm that your fix works. I misread your original description.
Thank you for putting in the extra work :)

PSteinhaus:
@nobbele I currently don't have access to a linux machine to try this out on.
Could you quickly run the repro to confirm that everything works fine on linux as well?

nobbele:
I am getting the same as win11 in the README on linux

PR #1006, opened 2022-01-16
Fix misnaming of `EventLoop` in HelloGgez guide

IndigoLily:
The text and link were referring to `EventsLoop`, which seems to have been its name prior to winit 0.20.0.


PSteinhaus:
thanks!

Issue #1012, opened 2022-02-10
WGPU transition


LechintanTudor:
I have some free time in the following weeks, I could help with transitioning to WGPU if nobody is working on it at the moment.

PSteinhaus:
Thank you! That's good to hear. @nobbele [startet to work on it](https://github.com/nobbele/ggez/tree/oblivion), some time ago, but we haven't yet found the time and energy to complete it since then. Feel free to take a look!

For me it will be a few weeks until I'll be able to really pull my weight in this again, but if you're interested and come across anything you'd like a quick explanation or second opinion on feel free to open an issue (probably on the [repo of the renderer behind the effort](https://github.com/nobbele/Oblivion)), or to just hit me up on Discord (where we can probably best be found over the Rust Game-Dev server).

nobbele:
Oblivion currently has some major issues (Text / Mesh blinking). 

And I think that the API of the WGPU library should be the graphics module rather than ggez::graphics being wrappers for graphics types like Oblivion is currently integrated. For this to work while being framework-agnostic all the functions have to take a generic context, my ideas for this is to either take `&mut impl HasContext`, or add a type parameter `struct Text<T: HasContext>` and re-export `pub type Text = wgpu_library::Text<Context>`.

Also due to the lack of lifetimes and mixed retained/immediate-mode drawing, there's no simple way to get wgpu to render your objects. Oblivion uses a queue system with Arcs but other ideas could be to use arenas.

LechintanTudor:
Thanks for the replies! For now, I'll try to port the existing code to WGPU with as few changes as possible and see if it works out.

nobbele:
Good luck!

aleokdev:
@jazzfool over at the [wgpu branch on my fork](https://github.com/aleokdev/ggez/tree/wgpu) has practically finished with the interface and is now squashing the few bugs that are remaining. One thing to note is that we are changing the interface drastically, taking a data-driven approach. The entire interface will change, and it'll result in a completely different (but still simple and easy to use) crate.

nobbele:
@aleokdev I just saw your implementation, it looks very good!
Could you make Canvas and Text only contain owning types though? Using lifetimes (in public API) is against the ggez way of doing things.

aleokdev:
> Using lifetimes (in public API) is against the ggez way of doing things.

[...]

PR #1019, opened 2022-03-12
Use subcontexts directly instead of free functions

aleokdev:
Deprecates the old free functions and uses the subcontexts directly, as mentioned in #1012 and discussed in the Matrix room. WIP.

Note: The graphical context hasn't been refactored since the changes would be overwritten by the WGPU PR anyways.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,11 +31,13 @@ name = "ggez"
 path = "src/lib.rs"
 
 [features]
-default = ["c_dependencies"]
+default = ["c_dependencies", "audio", "gamepad"]
 bzip2 = ["zip/bzip2"]
 mp3 = ["rodio/mp3"]
 multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
 c_dependencies = ["bzip2", "mp3"]
+audio = ["rodio"]
+gamepad = ["gilrs"]
 
 [dependencies]
 bitflags = "1"
@@ -49,7 +51,7 @@ old_school_gfx_glutin_ext = "0.28"
 glutin = "0.28"
 winit = "0.26"
 image = {version = "0.23.12", default-features = false, features = ["gif", "png", "pnm", "tga", "tiff", "webp", "bmp", "dxt", ] }
-rodio = { version = "0.14", default-features = false, features = ["flac", "vorbis", "wav"] }
+rodio = { version = "0.14", optional = true, default-features = false, features = ["flac", "vorbis", "wav"] }
 serde = "1"
 serde_derive = "1"
 toml = "0.5"
@@ -59,7 +61,7 @@ smart-default = "0.6"
 glam = { version = "0.20", features = ["mint"]}
 # Has to be the same version of mint that our math lib uses here.
 mint = "0.5"
-gilrs = "0.8"
+gilrs = { version = "0.8", optional = true }
 approx = "0.5"
 bytemuck = "1.7"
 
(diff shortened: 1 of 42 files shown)


aleokdev:
Completely forgot about Filesystem, sorry.

src/audio.rs
```diff
@@ -17,29 +18,17 @@ use std::sync::Arc;
...
 /// of your `Context` object.
-pub(crate) struct RodioAudioContext {
+pub struct AudioContext {
     _stream: rodio::OutputStream,
     stream_handle: rodio::OutputStreamHandle,
 }
 
-impl RodioAudioContext {
+impl AudioContext {
     /// Create new `RodioAudioContext`.
```
PSteinhaus, line 32: /// Create new `AudioContext`.
(or something similar)

src/context.rs
```diff
@@ -35,20 +35,21 @@ use crate::timer;
...
+    pub time: timer::TimeContext,
+    /// Audio context.
+    #[cfg(feature = "audio")]
+    pub audio: audio::AudioContext,
+    /// Keyboard input context.
+    pub keyboard: input::keyboard::KeyboardContext,
+    /// Mouse input context.
+    pub mouse: input::mouse::MouseContext,
+    /// Gamepad input context.
+    pub gamepad: input::gamepad::GamepadContext,
```
PSteinhaus, line 52: This needs the same treatment as audio, as it only compiles as long as `#[cfg(feature = "gamepad")]` is true.
```suggestion
    /// Gamepad input context.
    #[cfg(feature = "gamepad")]
    pub gamepad: input::gamepad::GamepadContext,
```

Cargo.toml
```diff
@@ -31,11 +31,13 @@ name = "ggez"
 path = "src/lib.rs"
 
 [features]
-default = ["c_dependencies"]
+default = ["c_dependencies", "audio", "gamepad"]
 bzip2 = ["zip/bzip2"]
 mp3 = ["rodio/mp3"]
 multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
 c_dependencies = ["bzip2", "mp3"]
```
PSteinhaus, line 38: Currently opting out of rodio by disabling "audio" would still fail, as "mp3" still links to rodio and is part of the default features through "c_dependencies", leading to rodio being loaded and compiled anyway.

PSteinhaus (changes requested):
Overall, it looks good. The central problematic is that the ripple effects of disabling either of the new features haven't been taken care yet. So there needs to be a little (in the case of audio), or even a bit more (in the case of gilrs) conditional compilation to make things work. Luckily, these passages are easy to find by disabling the two features (temporarily taking them out of the default feature list for example) and then bumping your head until everything compiles.

Cargo.toml
aleokdev: What's `c_dependencies` for?

aleokdev: > Currently opting out of rodio by disabling "audio" would still fail

Also, this is not possible; You can't disable a single feature. You need to disable all default features to enable one specifically.

PSteinhaus: `c_dependencies` groups together all libraries that at some point depend on a C-compiler, which can be undesirable, and luckily what they provide is all non-essential so they may be disabled.

You’re right, it’s a silly thought. Who would disable default features and then reenable `c_dependencies` without `audio`? And if you only need `bzip` or `mp3` but not the other there’s already features for that so it’s fine.

PSteinhaus:
Meh, mistook `RodioAudioContext` for `RodioAudioDevice` in the commit description...
I have a feeling that more similar minor comment issues are still hidden in this, but these would be really minor, as the general sense behind them should remain logical, so I'm good with merging this now.

PSteinhaus (approved):

nobbele (approved):

Issue #921, opened 2021-04-06
Issue using graphics::clear


KyleGBC:
I'm just started with using this library and ran into an issue with my project almost immediately. I was able to replicate my problem with a very simple program, leading me to believe it is a result of my own misunderstanding of the library. 

The basic program I set up to help me demonstrate is as follows: 
```
use ggez::{graphics, Context, ContextBuilder, GameResult};
use ggez::event::{self, EventHandler, KeyCode, KeyMods};

fn main() {
    let (mut ctx, mut event_loop) = ContextBuilder::new("Game", "Kyle").build().expect(">:(");
    let mut my_game = MyGame::new(&mut ctx);

    match event::run(&mut ctx, &mut event_loop, &mut my_game) {
        Ok(_) => println!("Exited cleanly."),
        Err(e) => println!("Error occured: {}", e)
    }
}

struct MyGame {
    image: graphics::Image,
    x: f32,
    y: f32,
}

impl MyGame {
    pub fn new(_ctx: &mut Context) -> MyGame {
        let image = graphics::Image::new(_ctx, "/red.png").unwrap();
        MyGame {
		    image, x: 10.0, y: 10.0
		}
    }
}

impl EventHandler for MyGame {
    fn update(&mut self, _ctx: &mut Context) -> GameResult<()> {
		Ok(())
    }

    fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
	graphics::clear(ctx, graphics::WHITE);

        let drawparams = graphics::DrawParam::default().dest([self.x, self.y]);
        graphics::draw(ctx, &self.image, drawparams)?;

        println!("Drawing at {}, {}", self.x, self.y);

	graphics::present(ctx)
    }
    fn key_down_event(&mut self, _ctx: &mut Context, keycode: KeyCode, _keymods: KeyMods, _repeat: bool){
        match keycode {
            KeyCode::Up => self.y -= 1.0,
            KeyCode::Down => self.y += 1.0,
            KeyCode::Right => self.x += 1.0,
            KeyCode::Left => self.x -= 1.0,
            _ => ()
        }
    }
}
```

As you can see, all this does is load an image, in this case, a 1x1 red png, and draw it at the stored coordinates, which are changed by pressing the arrow keys. 

However, running this doesn't work, and the image does not move around with the keypresses (although the `println!()` within the `MyGame::draw function` _does_ change accordingly).

[...]

Issue #560, opened 2019-01-20
Move blend mode from Drawable to DrawParams


ozkriff:
Why is (`set_`)`blend_mode` a part of the `Drawable` trait and not just a field of `DrawParams`? I can't see any complex logic in any `Drawable::set_blend_mode` implementation.

icefoxen:
Check out where the `blend_mode` is used now that I think of it.  I vaguely recall gfx-rs wanting it that way.

If we can put it in `DrawParams`, it might be nice.  Though I wonder if it would work in `SpriteBatch` in that case, since idk if you can have instanced drawing with different blend modes per instance.

PSteinhaus:
Blend modes are now no longer set per `Drawable`. Now, with the wgpu-PR merged, they're part of `Canvas`' state.
We could have also moved them to `DrawParam`, leading to them being included in the queue maintained inside of `Canvas`, but for now I think having blend mode be one of the two values in the internal-state-queue of `Canvas` (together with the shader used by it) makes sense.

Issue #368, opened 2018-05-04
Canvas vs. shader API's are inconsistent


icefoxen:
The shader functions returns a handle that pops to the old shader when it goes out of scope.  Canvas's have to be manually reset via `set_canvas(ctx, None);`.  The inconsistency irks me.  Which is better?

 * Shaders are strictly scope-bound because of this (I think?) unless you pass the handle around.  RAII style.
 * Canvas's are not, but don't have a stack structure either.  More C style.

Rust is generally more RAII-style but returning handles for a resource that are only used for freeing them somehow doesn't feel terribly Rust-y.  It happens, but shaders are sorta weird there 'cause you then don't DO anything with the handle, unlike say Mutex or such.

icefoxen:
notttttt going to fix this right now, there's too much stuff to do already.

icefoxen:
The RAII structure is better, because it restores the state automatically on errors or early returns... such as calling a function and using `?` after it.

icefoxen:
Should be solved by the `ggraphics` branch.

PSteinhaus:
`Shader` and `Canvas` APIs have been changed massively through the WGPU-PR.
Both choosing a canvas and choosing a shader are now strictly explicit:
- You either call `draw` directly on a canvas or hand over a reference to it when calling `Drawable::draw`.
- And you set the shader used for the following draw calls via `Canvas::set_shader` and probably `Canvas::set_default_shader`.

Issue #1008, opened 2022-01-17
Placing text in ggez 0.7


fenhl:
I wrote [a wrapper](https://github.com/dasgefolge/sil/blob/abd4d42025859574ec7cfb1ba00b29fb776ba081/src/text.rs#L59-L90) for drawing text with a given size and alignment in ggez 0.6. It handles things like keeping a half-font-size margin from the edges of the window.

After upgrading to ggez 0.7, this broke, presumably due to [the changes to `DrawParam::offset`](https://github.com/ggez/ggez/blob/devel/docs/FAQ.md#offsets). Now text that's supposed to be centered on the window is instead centered on the right edge of the window so that only half of it is visible.

I've tried various fixes, but none seem to restore the intended behavior. My most recent attempt (from a different project, so some of the details are different):

```rust
pub(crate) fn draw(mut self, handler: &crate::Handler, ctx: &mut Context) -> GameResult {
    let Rect { w, h, .. } = graphics::screen_coordinates(ctx);
    for fragment in self.text.fragments_mut() {
        fragment.color.get_or_insert(self.color.unwrap_or(Color::WHITE));
    }
    self.text.set_font(handler.dejavu_sans, PxScale::from(self.size));
    self.text.set_bounds([w - self.size, h - self.size], self.halign);
    let dim = self.text.dimensions(ctx);
    self.text.draw(ctx, DrawParam::default().dest([ // upper left corner
        match self.halign {
            HorizontalAlign::Left => self.size / 2.0,
            HorizontalAlign::Center => (w - dim.w) / 2.0,
            HorizontalAlign::Right => w - self.size / 2.0 - dim.w,
        },
        match self.valign {
            VerticalAlign::Top => self.size / 2.0,
            VerticalAlign::Middle => (h - dim.h) / 2.0,
            VerticalAlign::Bottom => h - self.size / 2.0 - dim.h,
        },
    ]))
}
```

With this, text that should be centered looks like it's right-aligned instead.

Is there a way to make this work in ggez 0.7?

PSteinhaus:
Yes, there is. In fact you already link to [the solution](https://github.com/ggez/ggez/blob/devel/docs/FAQ.md#offsets):

> What this means for you is: If you want `DrawParam::offset` to be a relative offset (i.e. [1,1] means "bottom right", [0.5,0.5] means "centered", etc.) for any of the types mentioned as "absolute interpretations" above, then you'll have to adapt your offset like this:
```rust
[...]

PR #1020, opened 2022-03-22
WGPU refactor

### LINKED ISSUE #441: TextCached doesn't scale on resize
I'm going to preface this by saying that great work has been put into this game engine, and I'm really enjoying using it.

---

When resizing a window, everything is automatically scaled to fill the window, but it seems that TextCached instances aren't.

This can be seen in the following example:
```rust
extern crate ggez;

use ggez::conf::Conf;
use ggez::event;
use ggez::event::EventHandler;
use ggez::graphics;
use ggez::graphics::DrawMode;
use ggez::graphics::Drawable;
use ggez::graphics::Point2;
use ggez::graphics::Rect;
use ggez::graphics::Text;
use ggez::graphics::TextCached;
use ggez::Context;
use ggez::GameResult;

struct MainState {
    text_cached: TextCached,
    text: Text,
}

impl EventHandler for MainState {
        Ok(())
    }

        graphics::clear(ctx);

        graphics::set_color(ctx, graphics::BLACK)?;
        let rect = Rect {
            x: 0.0,
            y: 0.0,
            w: self.text_cached.width(ctx) as f32,
            h: self.text_cached.height(ctx) as f32,
        };
        graphics::rectangle(ctx, DrawMode::Fill, rect)?;
        graphics::set_color(ctx, graphics::WHITE)?;
        self.text_cached.draw(ctx, Point2::new(0.0, 0.0), 0.0)?;

        graphics::set_color(ctx, graphics::BLACK)?;
        let rect = Rect {
            x: 0.0,
            y: self.text_cached.height(ctx) as f32 + 10.0,
            w: self.text.width() as f32,
            h: self.text.height() as f32,
        };
        graphics::rectangle(ctx, DrawMode::Fill, rect)?;
        graphics::set_color(ctx, graphics::WHITE)?;
        let dest = Point2::new(0.0, self.text_cached.height(ctx) as f32 + 10.0);
        self.text.draw(ctx, dest, 0.0)?;

        graphics::present(ctx);
        Ok(())
    }
}

fn main() {
    let mut conf = Conf::new();
    conf.window_setup.resizable = true;
    conf.window_mode.width = 200;
    conf.window_mode.height = 50;

    let ctx = &mut Context::load_from_conf("issue", "reujab", conf).unwrap();
    let default_font = ctx.default_font.clone();

    let text_cached = TextCached::new("This doesn't scale").unwrap();
    let text = Text::new(ctx, "This does scale", &default_font).unwrap();
    let state = &mut MainState { text_cached, text };
    event::run(ctx, state).unwrap();
}
```

Default window:
small

Resized window:
big

What is the benefit of using `TextCached` over `Text`, and why doesn't it scale like everything else on a window resize? Could this be better explained in the documentation?

### LINKED ISSUE #562: Offset + Text in 0.5
It seems that setting offset on the draw params of a simple Text object does not affect the rendering at all.

### LINKED ISSUE #594: Shader Generic over multiple Uniforms?
Currently it seems like there's only the ability to pass in one uniform (unless I'm reading the examples and code incorrectly, in which case feel free to close this immediately!), but it would nice to be able to gfx_defines!  multiple uniforms and send them in one shot to the shader, instead of requiring multiple passes. I know I could drop to the gfx context in the Context object, is there a nice-ish way to achieve this in ggez?

### LINKED ISSUE #751: Can't take screenshots of antialiased render target
**Describe the bug**
Taking a screenshot of a window created with `.samples(conf::NumSamples::Eight)` (or any number of samples) fails with:
`VideoError("Memory copy error: OutOfSrcBounds { size: [1920, 1080, 1], copy_end: [1920, 1080, 0] }")`

**To Reproduce**
* use `.samples(conf::NumSamples::Eight)` on window_setup
* run `graphics::screenshot(ctx)`

**Expected behavior**
Ideal: GGEZ is able to copy the msaa buffer, img.encode does anti alias resolve for me.
Or alternatively resolve happens when taking the screenshot (might be easier to implement)

**Hardware and Software:**
 - ggez version: 0.5.1
 - OS: Windows 10
 - Graphics card: Nvidia

### LINKED ISSUE #951: Add custom blend mode
Changing the blend modes, like we're now doing, will probably lead to users having problems when trying to port their code to 0.6.0.

Therefore (and because it would be nice to have regardless) we should implement another "custom" blend mode variant, which can be freely configurated to hold whatever ruleset you want. Any number of these custom variants could then be handed over to the context builder, so that they'd then be included in our list of useable blend modes.

This may be a bit more difficult than it seems though. Currently, the magic of shaders mostly seems to reside in the so called `PsoSet`, which differentiates between different rendering pipelines (one for each blend mode) via a HashMap of BlendModes. I'm unsure whether this would work as well if BlendMode was a truly algebraic enum.

### LINKED ISSUE #962: Discussion on replacing gfx-rs
`ggez` has used `gfx-rs` since version 0.3.0, and there have not been *major* changes to the basic structure since 0.5.0.  In that time `gfx-rs` has actually held up surprisingly decently, but currently has a number of downsides:

 * It is unmaintained.  Bugs will generally not be fixed and new features will not be added.
 * It was intended as a cross-API framework but as currently used, it essentially targets OpenGL only.  Binding to multiple API's efficiently required a different design, which is now `gfx-hal`.  Notably, it can't effectively target mobile or web platforms.
 * It has some overhead that doesn't really need to be there.  `SpriteBatch` and `MeshBatch` are artifacts of this.  It would be nice to have a lower level API and just do automatic batching of draw calls all the time, the way `tetra` does.
 *  The current graphics engine in ggez is honestly kinda oogly and I now know how to do things better.

Below here is the tale of my own experiences.  I fought with this quite a lot since ggez 0.5.0 but didn't quite get to a minimum viable product.  Since we're handing off to new maintainers, here's the chronicle of my experiences and decisions.  Hopefully it's useful.

So the first question for replacing `gfx-rs` is, what to replace it with?  The options are:

 * A high level drawing API such as bevy, kiss3d, etc.  I personally think doing the low level design is half the fun, so I never really investigated these.
 * A lower-level portability API such as `gfx-hal` or `rendy`.  I love these in theory, but I spent a long time playing with these in like 2019 and was not terribly impressed tbh.  They are big, complicated, have a large dependency surface, and still have bugs.  Maybe they've gotten better since then.  (I kind of mentally put `wgpu` in this category; in my experience `wgpu` is still in heavy enough development that it is not ready for prime time.)
 *  A low level GPU API: OpenGL, DirectX, Metal, or Vulkan.  In theory these are the most pain in the ass, but in practice have the fewest layers of pain-in-the-ass between your code and the actual GPU driver.

I eventually decided to use a low-level GPU API, so the next question was, which one?  I didn't want to write my own portability layer with multiple backends, so which API to use was dictated by what platforms I wanted to support.  As of mid 2021, the compatibility matrix looks something like this:

```
┌───────────────────────┬────────────┬─────────┬────────────────────┬─────────────────────┬────────────────────┐
│           .           │   Linux    │ Windows │       MacOS        │         Web         │  Mobile/embedded   │
├───────────────────────┼────────────┼─────────┼────────────────────┼─────────────────────┼────────────────────┤
│ OpenGL 2.1/ES2/WebGL1 │   ✓        │   ✓     │ ✓ (for now)        │  ✓                  │    ✓               │
│ OpenGL 4.3/ES3/WebGL2 │   ✓        │   ✓     │ ✓ (newer versions) │  ~ (newer versions) │    ✓ (new devices) │
│ Vulkan                │   ✓        │   ✓     │ ? (MoltenVK)       │  ✗ (prolly never)   │    ???             │
│ Metal                 │   ✗        │   ✗     │ ✓                  │  ✗                  │    ✗               │
│ DirectX               │   ? (DXVK) │   ✓     │ ✗                  │  ✗                  │    ✗               │
└───────────────────────┴────────────┴─────────┴────────────────────┴─────────────────────┴────────────────────┘
```

Web support info taken from https://caniuse.com/?search=webgl2 .  OpenGL version equivalences figured out by digging through the specs.  WebGL1 is kinda a subset of ES2 which is kinda a subset of OpenGL 2.1 with some extra things added.  WebGL2 is a direct subset of ES3 which is a direct subset of OpenGL 4.3.

At the time I was considering this, my main goal was supporting the existing Windows+Linux OS's, plus web and old mobile devices.  There's actually quite a lot of fun little Linux-based computers out there such as Raspberry Pi's < 4, the Pinebook Pro, quite high-quality [retro handheld devices](https://www.rghandhelds.com/), and other stuff like that.

So I decided I would work with OpenGL ES2.  ES2 is kinda ancient tech from the Bad Old Days of OpenGL, but that means there's *tons* of cheap as dirt chips out there that support it and it's often the first thing that new device drivers implement.  For example the Pinebook Pro has a GPU that supports ES3, but the Mesa Panfrost driver for it still only *reliably* supports ES2.  I would have much preferred Vulkan or ES3, but using ES2 meant that ggez would be able to run almost literally anywhere.  And frankly once you get above a basic level of usability, it doesn't make a whole lot of difference one way or another.

My work on an OpenGL ES2 renderer for ggez lives in the `ggraphics` branch, here: <https://github.com/ggez/ggez/tree/ggraphics/ggraphics>.  An older attempt using `rendy` lives here: <https://github.com/ggez/ggraphics>.  Both use a much more "retained mode" style of drawing, where you fill a buffer with things to draw and then hand the GPU whole buffers at a time.  Layered on top of that are further pieces of state, shader pipeline and render target.  The ES2 `ggraphics` branch readme still probably describes it as well as possible.  You should be able to make about 80% of ggez's immediate mode graphics API work with only minimal changes, it will make automatic draw batching possible so `SpriteBatch` and `MeshBatch` will become unnecessary, and it will forcefully clear up some of the places where `ggez`'s drawing API conflates different stages of the process (such as blend mode, some of the weirder things involving text drawing, etc).

### LINKED ISSUE #989: Memory leak in set_screen_coordinates()
**Bug description**
My game is leaking memory (at the speed of about 1MB in 5 seconds?) and while I was debugging I realized that after changing this line:
```rs
graphics::set_screen_coordinates(context, new_screen_view).unwrap();
```
to this:
```rs
// graphics::set_screen_coordinates(context, new_screen_view).unwrap();
```
... I could get the leak to disappear.

**To Reproduce**
I was able to reproduce the issue by adding a call to the `set_screen_coordinates` function in the `update` implementation of the [basic example](https://docs.rs/ggez/0.6.1/ggez/#basic-project-template) for this ggez version (ggez version number below).

**Expected behavior**
This is kinda obvious, but the expected behavior is just the function not causing a memory leak.

As I said before, adding a call to the `set_screen_coordinates` function in the basic template reproduces the issue:
```rs
use ggez::{Context, ContextBuilder, GameResult};
use ggez::graphics::{self, Color, Rect};
use ggez::event::{self, EventHandler};

fn main() {
    let (mut ctx, event_loop) = ContextBuilder::new("my_game", "Cool Game Author")
        .build()
    let my_game = MyGame::new(&mut ctx);
}

struct MyGame;

impl MyGame {
    pub fn new(_ctx: &mut Context) -> MyGame {
        MyGame {}
    }
}

impl EventHandler<ggez::GameError> for MyGame {
    fn update(&mut self, ctx: &mut Context) -> GameResult<()> {
        let new_screen_view = Rect::new(0.0, 0.0, 800.0, 600.0);
        graphics::set_screen_coordinates(ctx, new_screen_view).unwrap();
        Ok(())
    }

        graphics::clear(ctx, Color::WHITE);
        graphics::present(ctx)
    }
}
```

**Hardware and Software:**
 - ggez version: `0.6.1`
 - OS: Windows 11

### LINKED ISSUE #1003: Adapt Canvas implementation to remember current Canvases
Recently, @liquidev asked on Discord whether Canvases could be used like this:
```rust
canvas.draw_stuff(|ctx| {
  // draw things to the canvas here
}) ;
```
They can't.
But what if we actually changed the API to allow this? Specifically, I'm thinking about changing/expanding the API to include `graphics::use_canvas`, similar to `graphics::use_shader`.

As a possible implementation I could imagine to wrap the canvas data in a new struct `CanvasHandle` (again, like we do with shaders), which is wrapped in a Rc. Inside the context we could store `Weak<CanvasHandle>`s for all the Canvases created. Every once in a while we could go through this list and remove all `Weak<CanvasHandle>`s that cannot be upgraded to `Rc`s anymore, implicating that the Canvas has been dropped. (We don't currently do anything like this with shaders, because shaders are usually long lived and few).

This would of course add some overhead, mainly due to the use of the `Rc`s and having to upgrade and maintain the `Weak`s.

The question is: What would be the use cases of such functionality? I dimly remember that it might have been useful to have something like this when implementing the resolve shader workaround for multisampled canvases, but I really don't have any actual use cases in mind right now, so I'd be interested in hearing some opinions on that.

### LINKED ISSUE #1007: z-value in DrawParam
Currently there is no way of defining z-order in ggez. The z-order is defined only through the order in which the draw calls are made.

This is not good.
First off there are many situations where it's just more comfortable to set the z-order explicitly than it is to dance the "which draw call next"-dance. Then there is `Spritebatch`. You can't set z-order in a spritebatch, except through clearing it and pushing all sprites into it again in the new desired order. This is possible, but ridiculous and unnecessary.

The change wouldn't need to be big. We could just introduce a z-value field to `DrawParam` and adapt the function that calculates the matrices to respect that value and we'd be done. Alternatively we could change `dest` from a 2D point to a 3D one, but that would require changing a lot of code, both in our examples and also for people updating their ggez.

I lean towards the first approach, as it's both simpler and possibly less confusing, as ggez IS a 2D engine and it therefore makes sense for positions to be specified in two dimensions, though I realize this point is very much debateable.

### LINKED ISSUE #1012: WGPU transition

aleokdev:
- Closes #1012
- Closes #962
- Closes #562
- Closes #1007
- Closes #989
- Closes #951
- Closes #1003
- Closes #751
- Closes #594
- Closes #441 
Possibly also https://github.com/ggez/ggez/issues/921, https://github.com/ggez/ggez/issues/560, https://github.com/ggez/ggez/issues/368.


aleokdev:
Merge done, now starting to remove unrelated changes.

jazzfool:
Also closes #562, #1007, #989, #951, #1003, #751, #594, #441.
Possibly also #921, #560, #368.

README.md
```diff
@@ -53,12 +53,12 @@ your own libraries atop ggez.
 
 ### Supported platforms
 
- * Fully supported: Windows, Linux
- * Not officially supported but might work anyway: Mac
+ * Fully supported: Windows, Linux, MacOS
```
nobbele, line 56: Why did you add MacOS here and Android, iOS, Web below?

jazzfool: wgpu supports Metal natively. I've also tested it on Windows and MacOS. In theory, wgpu + winit should also support Android, iOS and Web but I haven't tested that. AFAICT, the only major blocker here is that gilrs does not have support for Android.

nobbele: Weren't the mac issues winit related

jazzfool: What were the issues?

PSteinhaus: There were several. The one that comes to my mind is https://github.com/ggez/ggez/pull/949.
Then there were:
- https://github.com/ggez/ggez/issues/577
- https://github.com/ggez/ggez/issues/577
- https://github.com/ggez/ggez/issues/541
- and perhaps https://github.com/ggez/ggez/issues/900

but we can't say if these are still current.

aleokdev: All of these seem to be closed though.

PSteinhaus: Yes, but only because Mac support was officially dropped.

PSteinhaus: I too think that it could work. But neither of us owning a Mac hasn't changed and probabaly will not change in the forseeable future. Theoretically I could try to build a Hackintosh, but to be honest that's just not worth it to me right now.

src/graphics/canvas.rs
```diff
@@ -1,388 +1,526 @@
...
-        panic!("Can not be called with a non-matrix DrawParam");
-    };
-    let new_src = Rect {
-        x: param.src.x,
-        y: (1.0 - param.src.h) - param.src.y,
-        w: param.src.w,
-        h: param.src.h,
-    };
-    param.src(new_src)
+    }
```
PSteinhaus, line 333: I'm wondering whether adding a `DrawParam`-like struct (`TextDrawParam` or something) might be worth it for having similar benefits (having comfortable ergonomics where you can emid most things, as they have default values).
But there are only 4 parameters and 1 of them will pretty much always be set over the default (pos), so it would only shorten it by the other 3 parameters. Really not sure whether that's worth the weight of creating/maintaining `TextDrawParam` though.

PSteinhaus:
Why doesn't `InstanceArray` respect z-order? I realize it's easier this way and being drawn as an instanced draw it could only uphold inner z-order anyway, but it's a bit confusing to have `Canvas` use a queue respecting it and `InstanceArray` use one that doesn't. To me as a user it would make more sense if ordering was handled in a unified way over all of ggez.

src/graphics/canvas.rs
nobbele: `TextDrawParam` would be useful. Shouldn't it be possible to impl From<DrawParam> for TextDrawParam if that's added.

PSteinhaus: That would be possible, but From/Into doesn't work transitively, meaning that we wouldn't get access to all the nice `From` implementations for `DrawParam`. A user would have to explicitly create a `DrawParam`, just to `into` it to a `TextDrawParam`. So if we want the same level of comfort for `TextDrawParam` we'd still have to implement `From` nicely there as well.

jazzfool:
> How do you feel about making `InstanceArray` respect z-order?

Yes, this would make sense, but it would make the current API (`push`, `set`, etc) a lot less efficient. I've been thinking about a different solution though, where we run a compute shader over the instance data prior to drawing, to produce a list of sorted instance indices. We could also produce these sorted indices on the CPU following every push/set/etc but I expect running the compute shader every frame will still outperform the CPU sorting for very large `InstanceArray`s.

README.md
jazzfool: I have a Mac system and I can confirm #577 and #553 are no longer issues. I do notice relatively high CPU usage across all the examples (30%) but the framerate is capped to the refresh rate correctly. I did not notice any jitter, so overall #541 is probably okay too.

PSteinhaus:
Yes, the loss in efficiency would be clear downside of this. Still, maintaining inner z-order would make it worth it to me. The fact that `InstanceArray` sorts z-values internally could be laid out clearly, to slightly alleviate the loss by making users aware of it.
The compute shader idea sounds really interesting. It has the definite upside that it only needs to be performed once per draw and that it makes working with the `InstanceArray` on the CPU-side easier/more efficient (for example then we could really treat it as an array, allowing to set instances at specific indices).

examples/01_super_simple.rs
```diff
@@ -24,26 +35,23 @@ impl event::EventHandler<ggez::GameError> for MainState {
...
+
+        canvas.finish(&mut ctx.gfx)?;
 
-        graphics::present(ctx)?;
         Ok(())
     }
 }
 
 pub fn main() -> GameResult {
+    env_logger::init();
```
PSteinhaus, line 52: This is probably a leftover from debugging and should go, right?

src/graphics/canvas.rs
```diff
@@ -1,388 +1,526 @@
...
+
+                if draw.state.params != state.params {
+                    if let Some((bind_group, layout)) = &draw.state.params {
+                        canvas.set_shader_params(bind_group.clone(), layout.clone());
+                    }
+                }
+
+                if draw.state.text_shader != state.text_shader {
+                    canvas.set_text_shader(state.text_shader.clone());
+                }
```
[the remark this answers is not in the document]
jazzfool: Good catch. Though, this shouldn't affect the blend modes. I remember noticing a slight difference in the blend_modes example too. Likely the preset blend modes have some differences- I'll have to take a look soon.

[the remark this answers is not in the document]
PSteinhaus: You're right, the change didn't fix the example.

examples/blend_modes.rs
```diff
@@ -58,65 +60,73 @@ impl MainState {
...
+                graphics::DrawParam::new()
                     .dest(pos + Vec2::from(REL_POSITIONS[i]))
                     .color(TRI_COLORS[i]),
-            )?;
+            );
         }
 
         // draw text naming the blend mode
-        let text = graphics::Text::new((name, self.font, 20.0));
+        canvas.set_blend_mode(BlendMode::ALPHA);
```
PSteinhaus, line 73: This currently doesn't have any kind of effect. The text still gets drawn with whatever blend mode was set before.

src/graphics/internal_canvas.rs
```diff
@@ -0,0 +1,616 @@
...
+    }
+
+    fn flush_text(&mut self) {
+        if self.queuing_text {
+            self.queuing_text = false;
+            let mut premul = false;
+            if self.blend_mode == BlendMode::ALPHA {
+                premul = true;
+                self.set_blend_mode(BlendMode::PREMULTIPLIED);
+            }
```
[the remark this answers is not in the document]
jazzfool: Good suggestion. User can now opt-out with `canvas.set_premultiplied_text(false)`.

examples/blend_modes.rs
jazzfool: I'm not sure what you mean exactly, but the blend_modes example should now produce an identical result to the original renderer.

PSteinhaus: My mistake, [0bc5f18](https://github.com/ggez/ggez/pull/1020/commits/0bc5f18a7cf0081e708b5c49692a6e3db7315897) fixed it.

README.md
PSteinhaus: That's good to hear, thanks for checking! Sadly I think MacOS needs to keep a special not-officially-supported-but-should-work status still, as long as we have no Macs ourselves, since we just can't promise any kind of serious support while not owning one.

jazzfool:
For now, the Z sorting within `InstanceArray` is done on the CPU. GPU sorting may be worth investigating if performance isn't acceptable.

src/graphics/instance.rs
```diff
@@ -39,29 +53,98 @@ impl InstanceArray {
...
+        self.dirty.load(SeqCst)
+    }
+
+    /// Uploads the instance data to the GPU.
     ///
-    /// Prefer this over `push` where possible.
+    /// You do not usually need to call this yourself as it is done automatically during drawing.
+    pub fn flush(&mut self, gfx: &GraphicsContext) {
+        self.flush_wgpu(&gfx.wgpu)
+    }
```
PSteinhaus, line 129: When _would_ a user need to call this manually?

PSteinhaus:
> For now, the Z sorting within `InstanceArray` is done on the CPU. GPU sorting may be worth investigating if performance isn't acceptable.

The bunnymark example went down from up to 35k bunnies at stable 120 FPS to 11k on my machine. That's still acceptable to me. The value for old ggez was around 32k. For users who don't care about z-order we could perhaps also give them the option to opt-out of it for `InstanceArray`, if they really need that performance. An easy, though hacky, way would be to give it a flag, which toggles whether `layers` (in `flush_wgpu`) actually gets created and filled as a BTreeMap, or whether `indices` just ends up being the range from 0 to `instances.len`.
A better way might be to have two instance array shaders, the old and the new one, and then have two structs `InstanceArray` and `InstanceArrayUnordered` (or the like) using the new, or the old approach respectively. But I don't know whether that might be overkill. I feel like it could be justified though.

aleokdev:
Considering how most tilemaps don't care about order in tiles of the same layer, adding an unordered version (or adding a const parameter) would be really useful.

PSteinhaus:
I just noticed that we might need to take a bit more caution with MSAA sampling counts. On my machine only `NumSamples::One` and `NumSamples::Four` work, for example. Anything else gives me:
```
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a RenderPass
      note: encoder = `<CommandBuffer-(0, 1, Vulkan)>`
    In a pass parameter
      note: command buffer = `<CommandBuffer-(0, 1, Vulkan)>`
    attachment's sample count 16 is invalid        // 16 being an example here
```

jazzfool:
This is an issue with WGPU, which only currently supports 1 and 4 samples. Hopefully in the future this will change.

PSteinhaus:
Alright, then we should remove all the other `NumSamples` variants till then.

PSteinhaus:
The reworked #[1c7a264](https://github.com/ggez/ggez/pull/1020/commits/1c7a264a916bfac5ef50b178e8c58b51ec31369b) `InstanceArray` gives me 21k bunnies at 120 FPS, but it gives me that both when using the ordered version and the unordered one...
Why does the `InstanceArrayInner` have to be behind a `Arc<RwLock<>>`?

examples/graphics_settings.rs
```diff
@@ -203,9 +214,6 @@ struct Opt {
     /// what level of MSAA to try to use (1, 2, 4, 8)
```
PSteinhaus, line 214: Maybe we should change the comment here (and the one in line 241) to only mention levels 1 and 4 for the time being?
Or do we leave it be?

jazzfool:
Strange performance characteristics... disabling ordering should return it to the original performance.
As for the `Arc<RwLock<>>`, it's really ugly and I tried to avoid it for as long as possible, but the problem is that when the user clones the `InstanceArray` and it is later automatically flushed, the original copy of the `InstanceArray` is left unchanged. This becomes especially problematic as flushing can entail resizing, but the resized buffers don't propagate to the original `InstanceArray` and eventually the GPU runs out of memory.

jazzfool:
Removed the `Arc<RwLock<>>` by changing when the instances are flushed. Overall performance has improved as well (wrt to the unordered case).

PSteinhaus:
> As for the `Arc<RwLock<>>`, it's really ugly and I tried to avoid it for as long as possible, but the problem is that when the user clones the `InstanceArray` and it is later automatically flushed, the original copy of the `InstanceArray` is left unchanged. This becomes especially problematic as flushing can entail resizing, but the resized buffers don't propagate to the original `InstanceArray` and eventually the GPU runs out of memory.

Ah, I see. Didn't even notice that `InstanceArray` was cloneable. I really don't think it needs to be.

Now I get stable 120 FPS up to around 28k bunnies. At 40k I'm at 94 FPS.
If I change `DrawUniforms::from_param` to take the `param` by reference instead I get them up to around 30k bunnies and at 40k I'm at 101 FPS.
With [613444e](https://github.com/ggez/ggez/pull/1020/commits/613444e2c8d67baed4b8770fe39290fa7ac36660) I get stable 120 FPS up to 35-36k bunnies and at 40k I'm still around 118 FPS.

PSteinhaus:
Do we want to keep the old `use_shader` function that resets to the previous shader once the returned object goes out of scope?

PSteinhaus:
And in the logging example we might need to showcase how to show only those messages that come from selected libraries (as wgpu just spams the console with its logs else).

jazzfool:
> Do we want to keep the old `use_shader` function that resets to the previous shader once the returned object goes out of scope?

This feels quite inconsistent with the rest of the drawing API, but it might help make the API feel more familiar.

PSteinhaus:
> > Do we want to keep the old `use_shader` function that resets to the previous shader once the returned object goes out of scope?
> 
> This feels quite inconsistent with the rest of the drawing API, but it might help make the API feel more familiar.

That's exactly how I'm feeling. Maybe it's not so bad to lose it. In that case it might be desirable though to have something like `Canvas::get_shader` and `Canvas::get_text_shader`? Just because you might not know which shader is currently set when setting a new shader on the canvas, so you might need to get that shader to restore it after you're done drawing with your other shader?

PSteinhaus:
With [](https://github.com/ggez/ggez/pull/1020/commits/45d555847ad3431c61da463c547982d5a82ee2ae)[45d5558](https://github.com/ggez/ggez/pull/1020/commits/45d555847ad3431c61da463c547982d5a82ee2ae) I'm now at the following:

unordered:
- stable 120 FPS until around 24-25k bunnies, 87 FPS at 40k
- stable 120 FPS until around 25-26k bunnies, 91 FPS when not using `unzip`

ordered:
- stable 120 FPS until around 23-24k bunnies, 80 FPS at 40k
- stable 120 FPS until around 24k bunnies, 85 FPS at 40k when not using `unzip`

"not using `unzip`" meaning:
```rust
pub fn set(&mut self, instances: impl IntoIterator<Item = DrawParam>) {
    self.dirty = true;
    self.params = instances.into_iter().collect();
    self.uniforms = self.params
        .iter()
        .map(|x| {
            DrawUniforms::from_param(
                &x,
                [self.image.width() as f32, self.image.height() as f32].into(),
            )
            .as_std140()
        }).collect();
}
```

but changing it to the following:
```rust
pub fn set(&mut self, instances: impl IntoIterator<Item = DrawParam>) {
    self.dirty = true;
    self.params.clear();
    self.params.extend(instances);
    self.uniforms.clear();
    self.uniforms.extend(self.params
        .iter()
        .map(|x| {
            DrawUniforms::from_param(
                &x,
                [self.image.width() as f32, self.image.height() as f32].into(),
            )
            .as_std140()
        }));
}
```
gives me:

unordered:
- stable 120 FPS until around 40-44k bunnies, 90 FPS at 64k

ordered:
- stable 120 FPS until around 37k bunnies, around 119 FPS at 40k, 90 FPS at 57k

jazzfool:
Good work! The perf is a lot better now indeed

aleokdev:
Intervening to comment that I'd really appreciate if a prerelease version was released when this PR is merged. Something like `0.8.0-dev.0` would do. I'm currently working on a project and need to do some really ugly workarounds to get some things to work as expected, mostly because there is no subcontext split.

nobbele:
> Intervening to comment that I'd really appreciate if a prerelease version was released when this PR is merged. Something like `0.8.0-dev.0` would do. I'm currently working on a project and need to do some really ugly workarounds to get some things to work as expected, mostly because there is no subcontext split.

what do you mean? You could just not upgrade to 0.8 the second it gets released.

nobbele:
I do think a pre-release should be done though, to see if any issues come up with the new graphics backend.

aleokdev:
> what do you mean? Wouldn't it be the same as a full release

You kind of answered yourself already, but uploading a full release just after merging wouldn't be a wise decision IMO due to possible emerging issues that would need to break the interface to be solved.

aleokdev:
What's currently blocking the merge?

jazzfool:
Documentation could be more handholding, like the old docs.

PSteinhaus:
> Documentation could be more handholding, like the old docs.

That's right. Also, please use the opportunity to also document/comment the new internals as good as you can, as this will greatly help us and others maintain them.

> What's currently blocking the merge?

Apart from that there's the issue of text now being void of metrics for the user to draw from when positioning it. This would probably be tolerable for now if we at least had an offset parameter on `draw_text` with a relative interpretation ((0.5, 0.5) being centered), but without that text just isn't that useable yet.

jazzfool:
Text measurement already exists under `GraphicsContext::measure_text` (and `GraphicsContext::measure_bounded_text`).

jazzfool:
Caching text dimensions shouldn't be necessary as it's already quite cheap given that glyph_brush does glyph caching.

PSteinhaus:
> Text measurement already exists under `GraphicsContext::measure_text` (and `GraphicsContext::measure_bounded_text`).

I'm sorry, I totally missed those.

> edit: Do we need text dimensions for the offset? As I understand, for text (as stated in the docs), the offset param uses the absolute interpretation. This is probably also why the original ggez blend_modes example labelling appears almost identical to the wgpu branch - because the offset of (0.5, 0.5) was actually only moving by (0.5, 0.5). Text centering (both vertically and horizontally) is already available, along with text wrapping bounds.

Hah, you're right. It seems I forgot to adapt the offset behavior in that example. There was a time when (0.5, 0.5) actually _was_ a centering offset for a text (at least if that text was drawn directly, without queuing others together with it). That was ggez 0.6. But then we changed offset behavior again, making the offset on Spritebatch and therefore Text absolute again.
But [that's not the behavior I prefer](https://github.com/ggez/ggez/issues/562#issuecomment-963637292). I just though that changing it yet again, so close to a major change in graphics, that would probably turn it on its head anyway, wouldn't be that helpful, or just not worth it.

With the ability to center text vertically and horizontically the need for such a relative offset might have lessened. Yet it would still be useful, for example for cases like the `blend_modes` example, where we want to draw some text centered on a certain position. Sure, we could also use an appropriate `Rect` and center text horizontically and vertically to achieve the same effect in this case, but it's a tiny bit more complicated and it might not be enough, or simply not the desired behavior in other cases:
If you, for example, want to move the pivot point of a paragraph of text to its left-middle, to render a menu item, or something, and the scale of the text changes, then it would be annoying to calculate fitting `Rect`s to appropriate that change and keep the text centered on the same horizontal line. You might also want to rotate text around its center, or similar things.

All in all, I think relative offsets are just more fitting for `Text`, than absolute ones. But I don't insist on that opinion and am open to getting convinced of the opposite.

src/graphics/text.rs
```diff
@@ -1,623 +1,321 @@
...
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct TextParam {
+    /// Text layout boundaries. The top-left of this [Rect] determines
+    /// the positioning of the text.
+    ///
+    /// The size of the [Rect] only becomes relevant if drawing with [TextLayout::Wrap].
+    pub bounds: Rect,
+    /// Rotation of the text in radians.
+    pub rotation: f32,
+    /// How the text should be layed out.
```
PSteinhaus, line 113:
```suggestion
/// should be positioned and laid out.
#[derive(Debug, Clone, Copy, PartialEq)]
pub struct TextParam {
    /// Text layout boundaries. The top-left of this [Rect] determines
    /// the positioning of the text.
    ///
    /// The size of the [Rect] only becomes relevant if drawing with [TextLayout::Wrap].
    pub bounds: Rect,
    /// Rotation of the text in radians.
    pub rotation: f32,
    /// How the text should be laid out.
```

PSteinhaus:
For consistency we should probably also add the one remaining `From` implementation for `DrawParam` also to `TextParam`:
```rust
/// Create a `TextParam` from a location
impl<P> From<P> for TextParam
where
    P: Into<mint::Vector2<f32>>,
{
    fn from(location: P) -> Self {
        TextParam::new().dest(location)
    }
}
```
While looking at this I noticed that there's a small inconsistency between types here. `DrawParam` uses `mint::Point2<f32>` for `dest` (and `offset`), while `TextParam` uses `mint::Vector2<f32>` everywhere. I don't really care which of them we change, as long as the result is consistent. Since `DrawParam` and its API have been around for a while I guess changing `TextParam` to use `mint::Point2<f32>` too is possibly a bit better though.

nobbele:
`canvas.draw_text` shouldn't these be a generic `canvas.draw` taking something Drawable instead?

jazzfool:
> `canvas.draw_text` shouldn't these be a generic `canvas.draw` taking something Drawable instead?

The `Drawable` trait doesn't have any meaningful purpose and will only get in the way now that `TextDrawParam` has been split out from `DrawParam`

nobbele:
Drawable could get an associated type or generic type parameter?
Also `draw` should really be `draw_rect`or similar

PSteinhaus:
> > `canvas.draw_text` shouldn't these be a generic `canvas.draw` taking something Drawable instead?
> 
> The `Drawable` trait doesn't have any meaningful purpose and will only get in the way now that `TextDrawParam` has been split out from `DrawParam`

This is a bit of a sticky thing indeed. Text probably just has to be dealt with differently than the rest, so having `draw_text` makes sense in any case. For the other three types of drawables (`Mesh`, `Image` and `InstanceArray`) I think a `Drawable` trait might still make sense to simplify the API, allowing us to reduce the functions to `draw_text` and `draw`. This would make it necessary to include images in `Mesh`es though, or to just leave them out of their drawing function, which would mean that textured meshes could only ever be drawn through `InstanceArray`s (which doesn't make sense from a user stand point).

nobbele:
Looking at it, why is `TextParam` a separate thing even?
`layout` should be part of the `Text ` rather than `TextParam` and other than that, it just seems to be questionably missing `color` and `transform`

nobbele:
Having `Drawable` be a unified trait allows continued use of it in user-code.

jazzfool:
> Looking at it, why is `TextParam` a separate thing even? `layout` should be part of the `Text ` rather than `TextParam` and other than that, it just seems to be questionably missing `color` and `transform`

There seems to be some misunderstanding of what `Text` represents: It represents a fragment of text in a paragraph, so that a slice of `Text` (which is what `draw_text` takes) can make up a whole paragraph. Including `layout` in `Text` wouldn't make sense.

Reusing `DrawParam` for `Text` will only make things confusing; what are `color` and `transform` supposed to do when `Text` already contains those parameters? I'd also like to point out that the idea of splitting out into a separate `draw_text` is not so alien given ggez already has `draw_queued_text`.

Another problem that hasn't been mentioned is that `Mesh` and `Image` are now disjoint, meaning you can draw the same `Mesh` with a different `Image` without recreating the `Mesh`. As such, `Canvas::draw_mesh` now takes an **additional parameter** for an optional image. That throws a wrench into the method signatures of `Drawable`.

If you feel a unified drawing interface is truly necessary, then I propose having an `enum Draw` with each variant corresponding to each type of draw. I believe that, compared to a drawable trait, this improves user discoverability, ease of use (for example, if the user wants to store drawables, they no longer need to use `dyn`), and means that the drawing API isn't artificially constrained by trying to fit absolutely everything into a single interface.

PSteinhaus:
> There seems to be some misunderstanding of what `Text` represents: It represents a fragment of text in a paragraph, so that a slice of `Text` (which is what `draw_text` takes) can make up a whole paragraph. Including `layout` in `Text` wouldn't make sense.

So `Text` is now what `TextFragment` is currently in ggez. How about changing its name to `TextFragment` and adding `Text` similar to what currently is there in ggez? Then the layout would be part of `Text` and we might be able to come by without `draw_text`.

> Reusing `DrawParam` for `Text` will only make things confusing; what are `color` and `transform` supposed to do when `Text` already contains those parameters? I'd also like to point out that the idea of splitting out into a separate `draw_text` is not so alien given ggez already has `draw_queued_text`.

`color` could be an additional tint color, or it could just do nothing, like is currently the case in ggez ([and mentioned in the docs](https://docs.rs/ggez/latest/ggez/graphics/struct.TextFragment.html)). And `transform` is something that you already basically hand over with `TextParam`: A destination, a rotation, all that's missing is the offset (and the scale, but that can already be controlled via font size, so I'm not sure whether it's worth to have it).
The major difference is that the text shader currently desires a `bounds` rectangle to place the text into. But when `layout` would be part of `Text` (given `Text` would be `TextFragment`), then I see no reason why `bounds` couldn't be a sensible part of `Text` as well (Ok, well, apart from the reason that `bounds` could then no longer also encode the destination, without strangely overwriting the destination given through `DrawParam`).

> Another problem that hasn't been mentioned is that `Mesh` and `Image` are now disjoint, meaning you can draw the same `Mesh` with a different `Image` without recreating the `Mesh`. As such, `Canvas::draw_mesh` now takes an **additional parameter** for an optional image. That throws a wrench into the method signatures of `Drawable`.

That's what I was trying to get at with this:
> This would make it necessary to include images in `Mesh`es though, or to just leave them out of their drawing function, which would mean that textured meshes could only ever be drawn through `InstanceArray`s (which doesn't make sense from a user stand point, or does it?).

Losing that additional parameter would certainly be inelegant, but if one wanted to further reduce the API by force it would be possible.

> If you feel a unified drawing interface is truly necessary, then I propose having an `enum Draw` with each variant corresponding to each type of draw. I believe that, compared to a drawable trait, this improves user discoverability, ease of use (for example, if the user wants to store drawables, they no longer need to use `dyn`), and means that the drawing API isn't artificially constrained by trying to fit absolutely everything into a single interface.

But users could also easily create such an enum themselves, if they ever needed it, right? After all no one forces you to store an `Image` and an `InstanceArray` together as dynamic trait objects, if you need to do so for some reason.
Unifying them through a trait would let us uphold the simpler drawing API that we currently have. I see no reason not to do so when we can.
Cramming everything into one interface _does_ have its downsides. For `Text` I'm very unsure about whether I really want to. That probably comes down to whether we want to use the shader that you've written for text, or whether we just want to treat it like any other quad instead, which would allow us to easily unify, but may come at a cost, as drawing text often _is_ different, if one desires to keep it contained in some `bounds` rectangle.
Putting an image back into `Mesh`, like it is currently the case, would certainly be a step back for `Mesh`. Here I'm not sure either whether unification is important enough to legitimize that cost.

nobbele:
>There seems to be some misunderstanding of what Text represents: It represents a fragment of text in a paragraph, so that a slice of Text (which is what draw_text takes) can make up a whole paragraph. Including layout in Text wouldn't make sense.

Ah sorry. Having draw_text be separate is fine and like @PSteinhaus said, renaming and re-adding old `Text` that is `Drawable` would be nice. I can see the appeal but switching the meaning of `Text` will be confusing, especially when resources might use older versions.

>Another problem that hasn't been mentioned is that Mesh and Image are now disjoint, meaning you can draw the same Mesh with a different Image without recreating the Mesh. As such, Canvas::draw_mesh now takes an additional parameter for an optional image. That throws a wrench into the method signatures of Drawable.

I thought of something, isn't it possible to draw everything with an image?
- Mesh with an image is already done.
- InstanceArray with an image is already done but it would need to be moved out of InstanceArray like with Mesh but it looks that'd require a performance penalty though.
- Text with an image would just ignore it or, use the image for additional coloring of the text.

This seems pretty weird to manage though.

>This would make it necessary to include images in Meshes though, or to just leave them out of their drawing function, which would mean that textured meshes could only ever be drawn through InstanceArrays (which doesn't make sense from a user stand point, or does it?).

Mesh could be `Drawable` without a texture, and add a `draw_textured_mesh` method to draw with texture. That'd make the API look like

>If you feel a unified drawing interface is truly necessary, then I propose having an enum Draw with each variant corresponding to each type of draw. I believe that, compared to a drawable trait, this improves user discoverability, ease of use (for example, if the user wants to store drawables, they no longer need to use dyn), and means that the drawing API isn't artificially constrained by trying to fit absolutely everything into a single interface.

This is a good idea.

Also my old comment:
>draw should really be draw_rector similar

draw_image would be a better name.

---

To summarize, IMO the API should look like
```rs
// Generic drawing methods, using Drawable trait OR impl Into<DrawEnum>.
canvas.draw(text_object, draw_param);
canvas.draw(instance_array, draw_param);
canvas.draw(mesh, draw_param);
canvas.draw(image, draw_param);

// Specialized drawing methods.
canvas.draw_queued_text(text_fragments, text_param);
canvas.draw_textured_mesh(mesh, image, draw_param);
canvas.draw_instanced_mesh(mesh, instance_array, draw_param);
```

PSteinhaus:
> I thought of something, isn't it possible to draw everything with an image?
> 
>     * Mesh with an image is already done.
> 
>     * InstanceArray with an image is already done but it would need to be moved out of InstanceArray like with Mesh but it looks that'd require a performance penalty though.
> 
>     * Text with an image would just ignore it or, use the image for additional coloring of the text.
> 
> 
> This seems pretty weird to manage though.
>
> Another idea is to impl Drawable for tuples, (Mesh, Image) and (Mesh, InstanceArray)

I like the first idea more than the second, though I don't really think we should move the image out of `InstanceArray`. Implementing Drawable for tuples would give us a simple interface, but in that case I'd really prefer the `enum Draw` that @jazzfool suggested over it for the mentioned discoverability and flexibility.

> Mesh could be `Drawable` without a texture, and add a `draw_textured_mesh` method to draw with texture. That'd make the API look like

I like that idea. Then we could have `Mesh`, `Image` and `InstanceArray` as `Drawable` and would be left with the specialized functions `draw_textured_mesh`, `draw_instanced_mesh` and `draw_text`.

> > If you feel a unified drawing interface is truly necessary, then I propose having an enum Draw with each variant corresponding to each type of draw. I believe that, compared to a drawable trait, this improves user discoverability, ease of use (for example, if the user wants to store drawables, they no longer need to use dyn), and means that the drawing API isn't artificially constrained by trying to fit absolutely everything into a single interface.
> 
> This is a good idea.

Looking at the different parameters of the different drawing functions I guess this is the sensible way to go forward. Either this, or keeping all the seperate functions (which is a tad more elegant from an implementation standpoint, but I'm unsure which of both is better from a user standpoint; I assume the former, as this would allow people to just use `draw` almost like they're used to).

> 
> To summarize, IMO the API should look like
> 
> ```rust
> // Generic drawing methods, using Drawable trait.
> canvas.draw(text_object, draw_param);
> canvas.draw(instance_array, draw_param);
> canvas.draw(mesh, draw_param);
> canvas.draw(image, draw_param);
> 
> // Specialized drawing methods.
> canvas.draw_queued_text(text_fragments, text_param);
> canvas.draw_textured_mesh(mesh, image, draw_param);
> canvas.draw_instanced_mesh(mesh, instance_array, draw_param);
> ```
> 
> This would make the trait not try to unify every way of drawing but just the primitive/common ways to allow a simpler API. If specialized drawing methods are needed along with the generic ones, It'd be trivial to make a custom enum/trait.

I don't think that having both `draw(text_object, draw_param)` _and_ `draw_queued_text(text_fragments, text_param)` (btw.: why not just `draw_text`?) is a good idea. We should probably decide for one of them, shouldn't we? Having the abilty to treat text as a regular quad with `draw` (given that the implementation is written so that the standard quad shader is used) or as something to be contrained by a bounding rectangle with the other call is nice, but it could also be very confusing, as it leads to doubled structures (layout and possibly bounds being stored in either `Text`, or in `TextParam`, depending on what function you want to use) and perhaps functioniality that may be assumed to provide the same effect, while actually working slightly differently.

jazzfool:
Just a note about the text rendering; compared to current ggez, text is now automatically queued and flushed on the fly, so a `draw_queued_text` no longer really applies. Also, moving text rendering to use the quad shader would be quite a bit less efficient in terms of both memory and runtime performance as it would mean 4x as much vertex data is being uploaded for each glyph (right now 1 vertex = 1 glyph, with per-instance vertex stepping).

Regarding the `bounds` parameter of text, this could probably be changed to be a vector size rather than a rect, such that the `DrawParam` determines top-left of the text bounds. Other than that, I'm happy with this API as well (will be able to start working on this again in a week or so).

PSteinhaus:
> ... Also, moving text rendering to use the quad shader would be quite a bit less efficient in terms of both memory and runtime performance as it would mean 4x as much vertex data is being uploaded for each glyph (right now 1 vertex = 1 glyph, with per-instance vertex stepping).

I don't necessarily advocate for using the same shader, there may be a seperate text shader still. What I wonder about is whether it might be better to write this shader so that it treats the text as a simple quad, as the default shader does, rotating it and applying offsets without respect to some bounding rectangle.
But I think I'm still misreading your text shader. How exactly is the bounding rectangle used there? Aren't bounds something that `glyphbrush` handles for us?

> Regarding the `bounds` parameter of text, this could probably be changed to be a vector size rather than a rect, such that the `DrawParam` determines top-left of the text bounds. Other than that, I'm happy with this API as well (will be able to start working on this again in a week or so).

Right, that might be sensible.

jazzfool:
Right so there probably has been some misunderstanding with the text shader. How it works is that each glyph corresponds to a single vertex *in the vertex buffer*. However, each glyph is still rendered as a quad and the actual quad vertices are generated on the fly in the vertex shader. The `rect` stored in the vertex data describes the rectangle of that single glyph's bounds to generate the quad from; this is indeed just data fed in from `glyph_brush`. It doesn't relate to the text layout bounds or anything. Also not sure if this was clear before, but text is rendered glyph by glyph (quad by quad) rather than as one large quad.

The other transformations (scale and translation) have already been applied to the `rect` before being passed into the vertex buffer, but obviously since it's an AABB, rotation needs to be passed and applied separately. Moving text rendering to use `DrawParam` will actual involve scrapping all that and passing the whole mat4 since `Transform` allows for arbitrary transform matrices.

PSteinhaus:
Ah, thank you, now I see. That's more elegant than using the other shader indeed.
I'm sorry, I really misunderstood the way it worked and thought it would lead to inconsistency with the way other things are drawn, but it works just fine.

The only thing that's really missing now is the `offset` parameter, which is a bit more tricky, as it usually has to be applied _before_ scaling and translation. It's a bit more difficult to apply per-glyph. We'd have to get the total scaled dimensions from glyphbrush and then do some math to apply an appropriate shift to all glyph bounds that leads to the same effect as if the shader operated on just one big text quad per `Text` with this offset. I remember how nobbele and I tried to do something like this for `Canvas`es some time ago and it really wasn't either fun, nor did it work out properly in our case.

So scrapping everything in order to just use `DrawParam` might be sensible after all. Or we keep on using `TextParam`, but with an additional `offset` parameter (and a similarly redesigned process). Whatever turns out to make more sense.


One other thing that I just noted while fumbling a bit with the docs is that we currently don't have anything like `Canvas::screen_coordinates`. I realize that such a function can only really give you meaningful results as long as you don't use `Canvas::set_projection` instead of just `Canvas::set_screen_coordinates`, but with this given as an explicit warning in the docs I think keeping this functioniality might still be useful. What do you think?

PSteinhaus:
How about making `GraphicsContext::wgpu` public, instead of supplying `GraphicsContext::wgpu(&self)`?
Of course, we'd need to add a big fat **handle with care** warning to it, but it would give those who want to do more with wgpu than what we offer more power, wouldn't it?

PSteinhaus:
I just noticed that supplying a non-default source parameter to `draw_mesh` has weird consequences (using `Rect::new(0.0, 0.0, 0.5, 0.5)` scales the meshes down by half), while in 0.7 it leads to them being drawn normally with just the texture being cropped and stretched to fit the mesh (visually speaking). This effect is the same in this branch, but additionally there is this undesired scaling of the vertex positions.

jazzfool:
> How about making GraphicsContext::wgpu public

Not sure how this would give more power to users who want to use wgpu directly; having those handles as mutable doesn't provide anything extra. That would also mean the user could reassign those handles and basically break the entire graphics context. It's also a bit unnatural in terms of API ergonomics to have everything but one field behind accessors.

> there is this undesired scaling of the vertex positions

This is because the mesh is being scaled by the image size. This scaling also takes into account the image source rectangle, so a half src-rect will mean half image scaling (btw, you can disable this scaling with `DrawParam::image_scaling`). It seems appropriate to retain 0.7 behaviour though.

> we currently don't have anything like Canvas::screen_coordinates

Yeah I left this out deliberately as the screen coordinate-based functions seemed more like helpers. Can return this functionality if needed though, no problem.

jazzfool:
The suggested API changes have been implemented.

The only two quirks with this new drawing API is when you want to draw with a simple, untextured quad (i.e., no mesh or image handle). For now, I've added `impl Drawable for ()` to handle this case. The other quirk is that `InstanceArray` needs to be mutable when drawing (because it is flushed which is mutating). To solve that, `Drawable` methods are now based on `self` rather than `&self`, and instead `Drawable` is implemented for `&T` (+ `T` and `&mut T` for text in particular as I found that makes it more ergonomic as I went through the examples).

jazzfool:
I looked into the src rect scaling issue and I'm actually unsure how 0.7 handles this. Looking at the `animation` example, given each frame of the animation sheet is a src rect, then it must be taking that into account when scaling the quad. Using the exact same `DrawParam` on both the wgpu branch and 0.7 leads to the same results which can only be possible if 0.7 also uses src rect to scale.

src/graphics/instance.rs
```diff
@@ -222,3 +222,20 @@ impl InstanceArray {
...
+                mesh: canvas.default_resources().mesh.clone(),
+                instances: (&*self).into(),
+            },
+            param,
+        );
+    }
+
+    fn dimensions(self, _gfx: &mut GraphicsContext) -> Option<Rect> {
+        None
+    }
```
PSteinhaus, line 240: How about taking over [the current implementation](https://github.com/ggez/ggez/blob/ab8e9e81ed93dcbec6ba14e3b83d4716e6fdff90/src/graphics/spritebatch.rs#L219-L234)?

src/graphics/draw.rs
```diff
@@ -254,6 +254,18 @@ where
...
 
+/// All types that can be drawn onto a canvas implement the `Drawable` trait.
+pub trait Drawable {
+    /// Draws the drawable onto the canvas.
+    fn draw(self, canvas: &mut Canvas, param: DrawParam);
+
+    /// Returns a bounding box in the form of a `Rect`.
+    ///
+    /// It returns `Option` because some `Drawable`s may have no bounding box,
+    /// namely `InstanceArray` (as there is no true bounds for the instances given the instanced mesh can differ).
```
PSteinhaus, line 265: Ah, ok, now I see why it returns `None`.
Hmm... we could remove `dimensions` from this trait and then just give `InstanceArray` it's own `dimension` function that also takes an `Option<Mesh>`. Or maybe that's not worth it...

PSteinhaus:
> I looked into the src rect scaling issue and I'm actually unsure how 0.7 handles this. Looking at the `animation` example, given each frame of the animation sheet is a src rect, then it must be taking that into account when scaling the quad. Using the exact same `DrawParam` on both the wgpu branch and 0.7 leads to the same results which can only be possible if 0.7 also uses src rect to scale.

You're right. I looked at it and it seems that this is somewhat inconsistent in ggez 0.7:
- `Image` and `SpriteBatch` take src into account (the latter only on the instance-level)
- `Mesh` and `MeshBatch` don't
(- `Text` uses `SpriteBatch` internally and therefore ignores src too)

examples/03_drawing.rs
```diff
@@ -162,23 +162,27 @@ impl event::EventHandler<ggez::GameError> for MainState {
         // Draw a filled rectangle mesh.
         let rect = graphics::Rect::new(450.0, 450.0, 50.0, 50.0);
         canvas.draw(
-            None,
+            (), // <- equivalent to quad mesh
```
aleokdev, line 165: I'm not sure I like this syntax, to be honest. Couldn't we do something like `graphics::QuadMesh` (Unit type) instead?

PSteinhaus: It would clutter the namespace a little, but it would also be more explicit, yes.

I'd say adding `graphics::Quad` (without the "Mesh" to avoid confusion) as a unit type sounds like a good idea.

jazzfool:
Do we want to rename `InstanceArray` to `SpriteBatch`?

nobbele:
ImageBatch?

aleokdev:
> The suggested API changes have been implemented.
> 
> The only two quirks with this new drawing API is when you want to draw with a simple, untextured quad (i.e., no mesh or image handle). For now, I've added `impl Drawable for ()` to handle this case. The other quirk is that `InstanceArray` needs to be mutable when drawing (because it is flushed which is mutating). To solve that, `Drawable` methods are now based on `self` rather than `&self`, and instead `Drawable` is implemented for `&T` (+ `T` and `&mut T` for text in particular as I found that makes it more ergonomic as I went through the examples).

If `Drawable` takes in `T`, I believe it would make sense to remove `DrawParam`s directly, and instead draw an object that refers to the object to draw, as well as the parameters. This would allow having more versatility on what an object *actually needs* to be drawn (e.g. `z` param is not useful to InstanceArray).

E.g.:
```rs
let text = graphics::Text::new("Hello, world!")
            .set_font("LiberationMono")
            .set_scale(48.)
            .clone();

canvas.draw(text.prepare(graphics::DrawParam::new().dest(dest_point)));
```

The alternative would be exposing an associated type as @nobbele has already mentioned.

nobbele:
You don't think some interior mutability wrapper around (dirty, buffer, indices, capacity) would make sense? It would improve ergonomics and consistency.

jazzfool:
> interior mutability wrapper around (dirty, buffer, indices, capacity)

For the sake of making flushing immutable and restoring the old `Drawable` signatures, this sounds like a good idea.

jazzfool:
> remove DrawParams directly, and instead draw an object that refers to the object to draw

To be fair, there's really not any redundant parameters in `DrawParam` anymore. The only one I can think of is `src` which is a no-op for `InstanceArray` and `Text`. `z` is still relevant to `InstanceArray` (and conditionally within the instances themselves).

PSteinhaus:
> Do we want to rename `InstanceArray` to `SpriteBatch`?

I like `SpriteBatch` because it gives me the most immediate feeling that I know what this is about when I read it.
`ImageBatch` sounds a bit more clunky to me (but that's, of course, very subjective) and while it's more consistent in that we have `Image`s, but not `Sprite`s, it intuitively sounds to me as if multiple images are batched together, which isn't what the struct is about.

Now that I think of it `SpriteBatch` might not make that much sense either, as `InstanceArray` is now more powerful than `SpriteBatch` currently is, in that it can also take in an additional `Mesh` in, thereby getting rid of `MeshBatch` by functionally combining both into one thing. So to me it'd make sense to name it something else, to point out that it _is_ something else now.

`InstanceArray` is pretty abstract/technical, but for those looking specifically for instanced drawing, it might yet be helpful.
An alternative would be something like `Batch`, or `InstanceBatch`, to communicate the purpose in a slightly less technical way.

But all in all, I still think `InstanceArray` is fitting.

PSteinhaus:
> If `Drawable` takes in `T`, I believe it would make sense to remove `DrawParam`s directly, and instead draw an object that refers to the object to draw, as well as the parameters. This would allow having more versatility on what an object _actually needs_ to be drawn (e.g. `z` param is not useful to InstanceArray).

It _would_ give us more versatility on what is actually needed to draw something. You're right there. But, as @jazzfool said:

> To be fair, there's really not any redundant parameters in `DrawParam` anymore. The only one I can think of is `src` which is a no-op for `InstanceArray` and `Text`. `z` is still relevant to `InstanceArray` (and conditionally within the instances themselves).

And even that we could change (at least for `Text`) if we gave `Text` an option to an internal image, possibly giving it a custom texture.
But that very sentence exposes a bit of a weakpoint of the current ggez-design (that I now made you adapt even more), which is that more versatility might actually **be** useful for us when it comes to applying texture.

But my intuitive approach for that would be to not remove `DrawParam`, but instead add onto it with something like a `texture` field. This would mean that `InstanceArray` would have to be adapted though, to be able to handle multiple samplers per draw, and as far as I know the number of texture slots per draw call is limited, so we might also need to add in checks for that, but I digress.

For what we have (except the texturing) I think `DrawParam` makes sense. Replacing it with some kind of `Draw` object (that might either be an implementor of a corresponding trait, or an enum) that also includes the reference to the thing to be drawn would be an option as well, but one question that I have there is: How would you handle `InstanceArray` then? What would go in there? It would have to be a `DrawParam`, right?

But then again, we could keep `DrawParam` for that single purpose and switch everything else to use a combined `Draw` object (or the like). Hmm...
If one wanted to make texturing more flexible an alternative to allowing multiple images in one `InstanceArray` draw call would of course also be to just keep it impossible, add a `DrawParam::texture` field anyway and make it a no-op on `InstanceArray`. But that's really uglier than having a custom `InstancedDraw` object that just shows you what you can actually do...

[We already kind of have these](https://github.com/aleokdev/ggez/blob/53e358fd504dd65aa3a9ec5c58044f8190587e41/src/graphics/canvas.rs#L451-L464) anway, really. They're just not public.

The only thing that bogs me down, really, is that having `DrawParam` and nothing but `DrawParam` just feels _simpler_. Right now I'm not sure whether I want to blow up the API a little, to allow for more versatility, or just keep what we have, to conserve its simplicity.

PSteinhaus:
Ok, I found a bug. Not one you're really responsible for, but one that persists from ggez 0.7. It's #1008
I already looked into it a bit back then, but I couldn't find the cause yet.

nobbele (approved):
I haven't had time to check it *completely* but it looks good to me now.

PSteinhaus:
I agree.
Thank you @jazzfool and @aleokdev for all of your generous effort and help.
I'll maybe take a few more quick looks at this, sleep over it and then merge it tomorrow.

PSteinhaus:
Ok, now I'm really happy with all of this. If noone disagrees I will merge this tomorrow morning.

aleokdev:
Aw, the issues didn't auto-close :(

PSteinhaus:
> Aw, the linked issues didn't auto-close :(

That's not too bad ;)
It gives me an incentive to actually provide meaningful closing comments on some of them.

PR #1028, opened 2022-04-17
Provide access to keyboard scancodes

RossBrunton:
Not sure if this is something that ggez would like, but opening this MR because it's something I've needed for my project.

Basically, this exposes the winit scancodes to the EventHandler should the developer want access to them. They can use these codes to implement support for non US QWERTY keyboards.

Have added it as a seperate event hook so that existing code and guides can use the "simple" function in case they don't need scancodes.

I'm also not a keyboard expert, so there's probably much better ways to go about doing this!

===

Extend the EventHandler trait to include key up/down events that have a
scancode. This can be used by developers to have layout-independant key
bindings.

### DIFF
--- a/examples/04_snake.rs
+++ b/examples/04_snake.rs
@@ -19,7 +19,7 @@ use oorandom::Rand32;
 
 // Next we need to actually `use` the pieces of ggez that we are going
 // to need frequently.
-use ggez::event::{KeyCode, KeyMods};
+use ggez::event::{KeyCode, KeyMods, ScanCode};
 use ggez::{event, graphics, Context, GameResult};
 
 // We'll bring in some things from `std` to help us in the future.
@@ -433,13 +433,14 @@ impl event::EventHandler<ggez::GameError> for GameState {
     fn key_down_event(
         &mut self,
         _ctx: &mut Context,
-        keycode: KeyCode,
+        _scancode: ScanCode,
+        keycode: Option<KeyCode>,
         _keymod: KeyMods,
         _repeat: bool,
     ) -> GameResult {
         // Here we attempt to convert the Keycode into a Direction using the helper
         // we defined earlier.
-        if let Some(dir) = Direction::from_keycode(keycode) {
+        if let Some(dir) = keycode.and_then(Direction::from_keycode) {
             // If it succeeds, we check if a new direction has already been set
             // and make sure the new direction is different then `snake.dir`
             if self.snake.dir != self.snake.last_update_dir && dir.inverse() != self.snake.dir {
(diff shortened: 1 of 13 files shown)


PSteinhaus:
Thank you!
I think exposing the scancodes is a good idea. We've had problems with this before (people using different kinds of keyboards sometimes making the keyboard events almost useless), so this may help some people and it should also provide a cleaner way to do a lot of the input for everyone.

Personally, I think we should fold the functionality into the existing keyboard events. If someone is explicitly and only interested in the text then we already have `text_input_event` after all.
What I'm not yet sure about is perhaps also adding two new functions to the keyboard context for this (something like `scancode_pressed` and `scancode_repeated`), for consistency. Then, we could also change `pressed_keys` to return a `&HashSet<Scancode, Option<KeyCode>>`, or add something like `pressed_scancodes` instead. But maybe that's overkill.

RossBrunton:
I've changed the signature of key_down_event for this, is this what you had in mind?

PSteinhaus:
> I've changed the signature of key_down_event for this, is this what you had in mind?

Yes, that's what I was thinking of, thank you.

nobbele:
Sorry for not checking this earlier, I'll do it tomorrow.

nobbele:
Again sorry for delaying this, I've had to do school-related stuff.

I'm not so sure about stuffing scancode into the key_ callbacks, it makes an already long parameter list even longer

PSteinhaus:
Hmm, but do you agree that exposing scancodes is a good idea in general? Would you prefer an additional event, as in the original commit?

nobbele:
Yes, it's a good idea in general. The more power to the user, the better.

RossBrunton:
Perhaps the handler could take an object that has scancode, keycode, modifiers etc. as fields? That is, take the existing parameters, and bundle them in an object.

PSteinhaus:
> Perhaps the handler could take an object that has scancode, keycode, modifiers etc. as fields? That is, take the existing parameters, and bundle them in an object.

We could to that and I could be talked into it, but honestly I'm not sure whether it's really necessary... to me having 5 parameters in these two events isn't that horrible, really.

PSteinhaus:
@nobbele I tried bundling the parameters in this [PR](https://github.com/RossBrunton/ggez/pull/1). What do you think?

nobbele:
looks good

RossBrunton:
Hello, thanks for making those changes. Shall I merge them into my branch so they can be reviewed here?

PSteinhaus:
> Hello, thanks for making those changes. Shall I merge them into my branch so they can be reviewed here?

Yes please :)

nobbele (approved):

PR #1030, opened 2022-05-10
Improve error type

### LINKED ISSUE #1021: Improve Error type
Right now, (almost) all error variants of `GameError` have no actual information about the error that a program can parse except a string. End programs wanting to know the exact issue that happened (for example, to solve it) cannot do so.

The actual error string can be provided in the display implementation, as is common in most crates. This'll also reduce duplicate code in the crate.

Another issue is that error variant fields are very undescriptive, for example `ResourceNotFound(String, Vec<(std::path::PathBuf, GameError)>),`; I suggest we change the tuple variants for named member ones.

aleokdev:
Closes #1021.


src/graphics/mesh.rs
```diff
@@ -467,9 +471,7 @@ impl MeshBuilder {
         P: Into<mint::Point2<f32>> + Clone,
     {
         if points.len() < 3 {
-            return Err(GameError::LyonError(
-                "MeshBuilder::polygon() got a list of < 3 points".to_string(),
-            ));
+            panic!("MeshBuilder::polygon() got a list of < 3 points");
```
sokorototo, line 474: I don't think panicking here is good, panicking within a library robs a user of the ability to manage it.

src/error.rs
```diff
@@ -6,92 +6,117 @@ use std::sync::Arc;
 /// An enum containing all kinds of game framework errors.
 #[derive(Debug)]
 pub enum GameError {
```
sokorototo, line 8: Using [`thiserror`](https://crates.io/crates/thiserror) would really do some justice to this code.

sokorototo: Note I am not a maintainer, just thought I would drop my opinion

src/graphics/mesh.rs
nobbele: checking the input length manually is trivial though. I do agree it might be better to keep as Err

aleokdev: As nobbele said, this error is trivial and is well documented. I don't think an error is required here.

src/error.rs
aleokdev: I know it exists, but I'd rather not use it to improve compile times, as it only makes a difference on the internal cleanliness of the code and not outside.

PSteinhaus: I agree that `thiserror` isn't really necessary here, though it might make things a bit simpler to write.

src/graphics/mesh.rs
PSteinhaus: I too, would like to keep the error for the reason @sokorototo cited.

Imagine some kind of particle system, where custom vertices are generated randomly and there's a small bug in the code that sometimes creates vertices with less than 3 points.

In the panic-case, if this bug isn't found early enough, then the whole app might panic and crash in production.

In the error-case, it would only crash if one was too lazy to handle the result properly and just unwrapped it. If one wasn't and handled the result as one would expect, then he'd probably do something like throwing away the error, only logging it perhaps, since the app can probably go on even if some particles somewhere are missing.

The panic may be well documented and somewhat trivial, but to me this is one of these situations that we luckily have `Result`s for in Rust, to force our users to at least stop for a moment and reflect on what is supposed to happen when this fails.

src/error.rs
sokorototo: Shouldn't we first use the `--timings` flag to check what is causing slow build times? Because I don't think `thiserror` will cause a large deviation.

PSteinhaus: Honestly I just don't care about shedding that boilerplate code. It's not that much and it's perfectly readable. Sure using a macro like `thiserror` would make it look more tidy. But to me trading in compile time, even if just very very little of it, for that just isn't worth it, sorry.

PR #1032, opened 2022-05-15
Add scissor_rect to canvas

### LINKED ISSUE #603: [Question] How can I clip a Mesh or other screen area based on supplied bounds?
I want to create a vertical scrollable region inside of specified Rect and I would like the content that exceeds the bounds of the area to be masked. 

I found this issue #468, but it doesn't help since I am not working with a tile map or spritebatch. I have tried using the Canvas to draw things off screen, but it is slow and seems buggy. 

Some ideas of what could help me:

* If MeshBuilder or Mesh had a clipping function, 
  * I could use that around the entire scrollable viewport. 
  * Or I could use that to clip items at the edge of the viewport, so that they do not overflow outside the bounds.
* If there was something like SpriteBatch that works with random mesh types and text and use that to draw only the desired "src" Rect
* An example that shows how to use the gfx pipeline or whatever to do this.
* If Canvas supported smaller off-screen windows that were much more performant. 

If you have some ideas, let me know. Thanks!

jazzfool:
Closes #603

### DIFF
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -106,6 +106,8 @@ impl Canvas {
     ) -> Self {
         let defaults = DefaultResources::new(gfx);
 
+        let drawable_size = gfx.drawable_size();
+
         let state = DrawState {
             shader: defaults.shader.clone(),
             params: None,
@@ -115,9 +117,9 @@ impl Canvas {
             blend_mode: BlendMode::ALPHA,
             premul_text: true,
             projection: glam::Mat4::IDENTITY.into(),
+            scissor_rect: (0, 0, drawable_size.0 as _, drawable_size.1 as _),
         };
 
-        let drawable_size = gfx.drawable_size();
         let screen = Rect {
             x: 0.,
             y: 0.,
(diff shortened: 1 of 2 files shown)


src/graphics/canvas.rs
```diff
@@ -281,6 +283,26 @@ impl Canvas {
...
+
+    /// Returns the scissor rectangle as set by [`Canvas::set_scissor_rect`].
+    #[inline]
+    pub fn scissor_rect(&self) -> Rect {
+        Rect::new(
+            self.state.scissor_rect.0 as f32,
+            self.state.scissor_rect.1 as f32,
+            self.state.scissor_rect.2 as f32,
+            self.state.scissor_rect.3 as f32,
+        )
```
PSteinhaus, line 303: Why not simply copy it?

PSteinhaus:
Good catch with the `flush_text`.

gzhynko:
Looks great! Maybe also add a `reset_scissor_rect` (or something like `set_default_scissor_rect` to match other similar methods on `Canvas`)?

src/graphics/canvas.rs
jazzfool: scissor rect is stored as (u32, u32, u32, u32), not Rect - because scissor rects are in pixel coordinates

PSteinhaus: Nevermind then, thought I checked it, but I guess I was too quick again.

PR #1037, opened 2022-05-27
Improve sub-context passing ergonomics

nobbele:
this allows passing just `ctx` to functions that take sub-contexts.
You can still split borrow with a slightly uglier construction: `&(&ctx.fs, &ctx.gfx)`.
I think something like 99.9% of users won't ever have to split borrow so in my opinion, this is a good trade-off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,9 +154,9 @@ impl EventHandler for MyGame {
     }
 
     fn draw(&mut self, ctx: &mut Context) -> GameResult<()> {
-        let mut canvas = graphics::Canvas::from_frame(&ctx.gfx, Color::WHITE);
+        let mut canvas = graphics::Canvas::from_frame(ctx, Color::WHITE);
         // Draw code here...
-        canvas.finish(&mut ctx.gfx)
+        canvas.finish(ctx)
     }
 }
 ```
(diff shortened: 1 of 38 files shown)


PSteinhaus:
Interesting idea! It certainly improves ergonomics.
Two minor things:
1. Why did you call the context parameter in the `HasTwo` cases `extra`? Wouldn't something like `contexts` be more meaningful?
2. Do you think we should document how to split-borrow the context in some kind of example? I think you're right and most people won't ever use it, but I can imagine that there'll be enough people who _wonder_ why this is the way it is and how the intended use it. I think the best way would be to shortly summarize the idea behind it in the documentation of `Has` and `HasTwo`, including a short code example on how to operate when splitting the context.

It certainly is a bit ugly, but for a library of our size and ambition (i.e. making life as easy as possible) it makes sense to me.

nobbele:
> Why did you call the context parameter in the HasTwo cases extra? Wouldn't something like contexts be more meaningful?

fair, or `ctxs` or just `ctx`

> Do you think we should document how to split-borrow the context in some kind of example?

sure, I'll write some documentation tomorrow

PSteinhaus (changes requested):
Almost finished. I found some final places where the `Has`-pattern hadn't yet been adopted and also changed some tiny other things, see the [PR](https://github.com/nobbele/ggez/pull/1).

PSteinhaus (approved):

PR #1038, opened 2022-05-30
Add transparent option to the WindowMode settings.

Cosmic-Goat:
This very simple PR just adds a transparent option in WindowMode to allow for transparent windows.


nobbele:
Seems to work, thanks for you contribution!

PR #1041, opened 2022-06-08
fix context doesn't have a timer field

suryanshmak:
update to use the delta associated function of timer

### DIFF
--- a/docs/guides/HelloGgez.md
+++ b/docs/guides/HelloGgez.md
@@ -228,7 +228,7 @@ We'll use [`timer::delta`](https://docs.rs/ggez/0.7.0/ggez/timer/fn.delta.html) 
 
 ```rust,skt-update,no_run
 fn update(&mut self, ctx: &mut Context) -> GameResult {
-    self.dt = ctx.time.delta();
+    self.dt = timer::delta(ctx);
     Ok(())
 }
 ```


nobbele:
I'm not sure I get it. Context has a time field and you should use it over timer::delta

suryanshmak:
it doesn't unless you are talking about another Context type

nobbele:
I think you might have missed the 0.8.0-rc0 update (which master currently points to). See https://www.reddit.com/r/rust_gamedev/comments/v6a3kn/ggez_080rc0_is_out/

suryanshmak:
think i did but the version listed in the docs must be updated from "0.7" to "0.8.0-rc0". i'm just starting out with game dev in rust

nobbele:
0.8.0-rc0 isnt fully released yet so docs dont point to it. To be honest, we could've pushed devel to crates.io and kept master at 0.7 until 0.8 was properly released, I'll keep this in mind next time. Thank you for contributing and pointing this out!

suryanshmak:
also the [https://github.com/ggez/ggez/blob/master/docs/guides/GenerativeArt.md](canvas creation) guide needs to foreshadow that the `from_frame` only works with "0.8.0-rc0".

just read what you replied. sounds good to me!

PR #1049, opened 2022-06-21
added features for zip compression

### LINKED ISSUE #1039: Support for compression in Zip-based Filesystems
While testing out this library, I have discovered an issue with the `add_zipfile_bytes` function when used with a zip file created with the standard Unix zip utility. This issue was a panic in initializing the ContextBuilder, along the lines of:
```
thread 'main' panicked at 'Should never happen!: UnsupportedArchive("Compression method not supported")'
```
Upon some investigation with the `zipinfo -v` command, there are no issues with the file, and the compression method is `deflated`, a common compression algorithm that should be supported by any library dealing with zips. I then did some investigation into the zip crate that is used by ggez. Indeed, it does support the deflate compression method, provided that this feature is included. It is actually in the default features, but in the Cargo.toml file for ggez, we have the line
```toml
zip = { version = "0.5", default-features = false }
```
With this set of features, the zip crate will only support the compression method "stored"... which is no compression at all. This is contradictory to the expectations one might have when using any function dealing with zip files. Accordingly, I would like to request one of the following two changes be made:
1. Support standard compression algorithms by removing the `default-features = false` line from the Cargo.toml file
2. If it is not desirable to support compression, then make it abundantly clear in all documentation that involves interacting with zip files, that all zip files must be uncompressed.

PSteinhaus:
Closes #1039

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,15 +26,16 @@ path = "src/lib.rs"
 [features]
 default = ["c_dependencies", "audio", "gamepad"]
 bzip2 = ["zip/bzip2"]
+zstd = ["zip/zstd"]
 mp3 = ["rodio/mp3"]
 multithread-image-decoding = ["image/hdr", "image/jpeg_rayon"]
-c_dependencies = ["bzip2", "mp3"]
+c_dependencies = ["bzip2", "zstd", "mp3"]
 audio = ["rodio"]
 gamepad = ["gilrs"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


nobbele:
should we use a single "zip-compression" feature rather than two separate ones?

PSteinhaus:
Yeah, that's probably better.

PSteinhaus:
darn, forgot to change the commit message...

PR #1050, opened 2022-06-21
Adds deprecated functions, compatiblity changes and moves request_quit to be a method on context

nobbele:
While making this, I also realized we were exposing glam::Vec2 rather than mint::Point2 in some functions, removed the type alias to prevent such mistakes again.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -67,6 +67,16 @@ pub struct Context {
     pub quit_requested: bool,
 }
 
+impl Context {
+    /// Attempts to terminate the [`ggez::event::run()`](fn.run.html) loop by requesting a
+    /// [`quit_event`](EventHandler::quit_event) at the very start of the next frame. If this event
+    /// returns `Ok(false)`, then [`Context.continuing`](struct.Context.html#structfield.continuing)
+    /// is set to `false` and the loop breaks.
+    pub fn request_quit(&mut self) {
+        self.quit_requested = true;
+    }
+}
+
 // This is ugly and hacky but greatly improves ergonomics.
 
 /// Used to represent types that can provide a certain context type.
@@ -369,6 +379,15 @@ impl ContextBuilder {
 
         Context::from_conf(config, fs)
     }
+}
+
+/// Attempts to terminate the [`ggez::event::run()`](fn.run.html) loop by requesting a
+/// [`quit_event`](EventHandler::quit_event) at the very start of the next frame. If this event
+/// returns `Ok(false)`, then [`Context.continuing`](struct.Context.html#structfield.continuing)
+/// is set to `false` and the loop breaks.
+#[deprecated(since = "0.8.0", note = "Use `ctx.request_quit` instead.")]
+pub fn quit(ctx: &mut Context) {
+    ctx.request_quit();
 }
 
 #[cfg(test)]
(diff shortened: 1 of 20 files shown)


PSteinhaus:
Sorry, I started taking a look, but didn't really get to it yet.

src/graphics/draw.rs
```diff
@@ -161,18 +161,21 @@ impl DrawParam {
         self
     }
 
+    pub(crate) fn get_dest_mut(&mut self) -> &mut mint::Point2<f32> {
+        if let Transform::Values { dest, .. } = &mut self.transform {
+            dest
+        } else {
+            panic!("Cannot calculate destination value for a DrawParam matrix")
```
PSteinhaus, line 168: Do we perhaps want to change this to an Error?
It's not transparent that `dest` may panic on setting the destination.

src/context.rs
```diff
@@ -371,6 +381,15 @@ impl ContextBuilder {
...
 
+/// Attempts to terminate the [`ggez::event::run()`](fn.run.html) loop by requesting a
+/// [`quit_event`](EventHandler::quit_event) at the very start of the next frame. If this event
+/// returns `Ok(false)`, then [`Context.continuing`](struct.Context.html#structfield.continuing)
+/// is set to `false` and the loop breaks.
+#[deprecated(since = "0.8.0", note = "Use `ctx.request_quit` instead.")]
+pub fn quit(ctx: &mut Context) {
+    ctx.request_quit();
+}
+
```
PSteinhaus, line 392: Here we could even keep the old implementation of just setting `ctx.continuing` to `false`. This would lead to no `quit_event` being triggered, as it is in ggez 0.7 and in all versions prior.

On the other hand I like that this is no longer possible... but since this is a deprecated function it might make sense to keep the deprecated behavior still.

src/context.rs
```diff
@@ -67,6 +67,16 @@ pub struct Context {
     pub quit_requested: bool,
 }
 
+impl Context {
+    /// Attempts to terminate the [`ggez::event::run()`](fn.run.html) loop by requesting a
+    /// [`quit_event`](EventHandler::quit_event) at the very start of the next frame. If this event
+    /// returns `Ok(false)`, then [`Context.continuing`](struct.Context.html#structfield.continuing)
+    /// is set to `false` and the loop breaks.
```
PSteinhaus, line 74: Some of these links might have been broken by being moved. I'll try to update them with a quick patch this evening.

PSteinhaus (commented):
Looks pretty good. Thanks for doing this.

src/context.rs
nobbele: When was `quit_event` even called before

PSteinhaus: Only when winit requested a quit, when the user tried to close the window for example.

src/graphics/draw.rs
PSteinhaus: `draw_queued_text` is affected as well and already returns a `GameResult`, so no change there.
Having `dest` return one would be slightly inconvenient though (if we return one there we might as well return one on each `DrawParam` method as modifying a matrix in either of these ways doesn't make sense). So how about only actually returning the error for `draw_queued_text` and silently ignoring it in `dest`?

Issue #1053, opened 2022-06-26
Maybe move drawables constructor to GraphicsContext?


nobbele:
`Image::new(&(&ctx.gfx, &ctx.fs), filename", false)` -> `ctx.gfx.new_image(&ctx.fs, filename, false)`
`Canvas::from_frame(ctx, Color::WHITE)` -> `ctx.gfx.new_canvas_from_frame(Color::WHITE)`

If we did this, I don't think you'd ever need to do the ugly tuple borrow (gets rid of `HasTwo`). Also even more similar to how löve2d looks `love.graphics.newImage(filename, settings)`

PSteinhaus:
I think this is a bit too much.
Yes, it would get rid of `HasTwo` (if we also did something similar to the audio context I think), but as a user I'd be a bit annoyed to search for the constructor to a thing and find it somewhere outside. Discoverability is more important than ergonomics here, imo.

PR #1055, opened 2022-06-27
Removes HasTwo by making Filesystem (privately) clonable

nobbele:
Like said in #1053, this removes the ugly tuple borrow

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ glam = { version = "0.20", features = ["mint"] }
 mint = "0.5"
 gilrs = { version = "0.8", optional = true }
 approx = "0.5"
-bytemuck = "1.7"
+bytemuck = { version = "1.7", features = ["derive"] }
 pollster = "0.2"
 memoffset = "0.6"
 crevice = "0.8"
--- a/examples/cube.rs
+++ b/examples/cube.rs
@@ -18,7 +18,8 @@ type Isometry3 = Mat4;
 type Point3 = Vec3;
 type Vector3 = Vec3;
 
-#[allow(dead_code)]
+#[derive(Clone, Copy, bytemuck::Zeroable, bytemuck::Pod)]
+#[repr(C)]
 struct Vertex {
     pos: [f32; 4],
     tex_coord: [f32; 2],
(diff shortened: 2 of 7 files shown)


src/filesystem.rs
```diff
@@ -34,11 +34,12 @@
 
 use crate::{
     conf,
-    vfs::{self, VFS},
+    vfs::{self, OverlayFS, VFS},
     Context, GameError, GameResult,
 };
 use directories::ProjectDirs;
-use std::{env, io, io::SeekFrom, path};
+use parking_lot::Mutex;
```
PSteinhaus, line 41: why use `parking_lot::Mutex` instead of a standard mutex?

src/filesystem.rs
```diff
@@ -47,13 +48,29 @@ const CONFIG_NAME: &str = "/conf.toml";
...
+        Filesystem {
+            vfs: self.vfs.clone(),
+            resources_dir: self.resources_dir.clone(),
+            zip_dir: self.zip_dir.clone(),
+            user_config_dir: self.user_config_dir.clone(),
+            user_data_dir: self.user_data_dir.clone(),
+        }
+    }
+}
+
```
PSteinhaus, line 73: Why call this `PrivClone` instead of just implementing `Clone` for Filesystem?

PSteinhaus (commented):
It's a bit of a workaround, but I agree that it's the better workaround, compared to `HasTwo`.

src/filesystem.rs
nobbele: std mutex gained good performance on linux only recently but parking_lot still has better "low contention" performance, I don't expect mutexes to ever block (realized an issue here, mutexes won't work in wasm) so there will always be zero contention. And I always see parking_lot being recommended over std in general but maybe it doesn't make sense to bring in another dependency (and even if we did, it should probably be applied to all mutexes in that case)

nobbele: actually nvm parking_lot mutexes do work in wasm as long as they don't block, which they won't if there are no threads, which there aren't in wasm (says some github comment, didn't verify)

nobbele: I'll switch to std mutex for now though, we can decide on that later

PSteinhaus: > actually nvm parking_lot mutexes do work in wasm as long as they don't block, which they won't if there are no threads, which there aren't in wasm (says some github comment, didn't verify). Do std mutexes work there though?

Looking it up quickly it seems they do. I'd say we stick with `std` until we get a good reason to switch and then switch everywhere at once, yes.

src/filesystem.rs
PSteinhaus: Hah. That makes sense ^^
Hmm, maybe we should add some kind of reminder to the next 0.9 milestone for this? If we didn't change it by then then I'd just replace this with `Clone`.

For now we can keep it though. But please do add that comment, that's a good idea.

nobbele:
uh did I break it?

Issue #1052, opened 2022-06-26
Drawn image size is different between GGEZ 0.7.0 compared to 0.8.0-rc0 when using a src rectangle


RossBrunton:
**Describe the bug**
I have this image:
![Test](https://user-images.githubusercontent.com/662364/175829893-9bbb39de-1596-4f0c-bcda-a0fcb1294334.png)
Which consists of four tiles, and I want to only draw the top-left one. So I set the DrawParam's "src" rectangle to (0, 0, 0.5, 0.5), and then draw it to the screen.

On GGEZ 0.7.0, this gives this:
![0 7 0](https://user-images.githubusercontent.com/662364/175830001-3b6e883e-db94-4101-86a3-f2cea7c176bc.png)

While on GGEZ 0.8.0 gives this:
![0 8 0](https://user-images.githubusercontent.com/662364/175829958-5c4a0455-4c7d-4061-a5bf-585054696fd1.png)

(The background is a red 48x48 square).

It seems that when drawing the image, the "original" size is used now, rather than new sliced size.

**To Reproduce**
```rust
    fn draw(&mut self, ctx: &mut Context) -> GameResult {
        let mut canvas = graphics::Canvas::from_frame(
            ctx,
            graphics::CanvasLoadOp::Clear([0.1, 0.2, 0.3, 1.0].into()),
        );

        let blank = graphics::Image::from_solid(&mut ctx.gfx, 48, Color::RED);
        canvas.draw(&blank, graphics::DrawParam::new());

        let image1 = graphics::Image::from_path(ctx, "/Test.png", true)?;
        canvas.draw(&image1, graphics::DrawParam::new().src(Rect::new(0.0, 0.0, 0.5, 0.5)));

        canvas.finish(ctx)?;

        Ok(())
    }
```

This is the version in 0.8.0 draw function that created these.

**Expected behavior**
I expected the behaviour to be consistent with 0.7.0 and draw the pixels 1:1 because the scale is 1.0, although I don't know enough about graphics hardware to say if this is correct or not.

If this is the new intended behaviour, might be worth adding it to the patch notes as well as instructions to mitigate it. I know it can be worked around using:
```rust
param.src(rect).image_scale(false).scale([scale * w, scale * h])
```
But that doesn't seem intuitive or discoverable to me.

**Screenshots or pasted code**
```rust
use ggez::{
    event,
    graphics::{self, Color},
    Context, GameResult,
};
use glam::*;
use std::{env, path};
[...]

PR #1060, opened 2022-07-01
Fix WGPU bugs

### LINKED ISSUE #1051: Canvas::from_image not working properly for images smaller than window size (patch included)
**Description**
Using an intermediate canvas for drawing, created by `Canvas::from_image(..)`, will under certain circumstances cause an error such as the following:

```
Caused by:
   In a RenderPass
     note: encoder = `<CommandBuffer-(0, 1, Vulkan)>`
   In a set_scissor_rect command
   Invalid ScissorRect parameters
```

**To Reproduce**
This bug will only occur, if you create the Canvas from an image which has a size smaller than the current window.

Tested on ggez-0.8.0-rc0.

**Solution**
I googled around for the set_scissor_rect error, along with `wgpu` and found some hints that this might happen, if the `scissor_rect` is larger than the image size. Based on this hypothesis, I modified the ggez code. I found the corresponding part in 'ggez/src/graphics/canvas.rs' in the function `new` and changed it as follows:

```
- scissor_rect: (0, 0, drawable_size.0 as _, drawable_size.1 as _),
+ scissor_rect: (0, 0, target.width() as _, target.height() as _),
```

The existing assignment would always use the size of the drawable, which is basically the size of the window.

You also need to adjust the `screen` a few lines later on:

```
- w: drawable_size.0 as _,
- h: drawable_size.1 as _,
+ w: target.width() as _,
+ h: target.height() as _,
```

I am not sure if this is really fixed the cause or is just a fix by coincidence, but it sure worked for me.

Sorry for not sending in a proper pull request, but I am not so familiar with this and I thought it might be better to contribute this way than not at all.

### LINKED ISSUE #1054: InstanceArray appears to leak memory on Linux + Nvidia GPU
**Describe the bug**
The following:
```rust

        let mut canvas = graphics::Canvas::from_frame(
            ctx,
            graphics::CanvasLoadOp::Clear([0.1, 0.2, 0.3, 1.0].into()),
        );
        let mut batch = graphics::InstanceArray::new(ctx, None, 64, false);
        batch.push(graphics::DrawParam::new());
        canvas.draw(&batch, graphics::DrawParam::new());
        canvas.finish(ctx)?;

        Ok(())
    }
```

Appears to leak memory on my system, at a rate of about 1MiB per few seconds. The batch.push seems to be important, otherwise memory usage stays stable. Interestingly, more batch.pushes don't seem to cause memory to drain faster.

I tried running it through Valgrind and nothing stuck out to me. Though I don't know if that's because my Valgrind skills are lacking.

**To Reproduce**
* Run example code.
* Watch the "Memory" field in the system monitor.

**Expected behavior**
Memory usage to stay stable, as it does with this sample:
```rust

        let mut canvas = graphics::Canvas::from_frame(
            ctx,
            graphics::CanvasLoadOp::Clear([0.1, 0.2, 0.3, 1.0].into()),
        );
        canvas.finish(ctx)?;

        Ok(())
    }
```

```rust
use ggez::{
    event,
    graphics::{self, Color},
    Context, GameResult,
};

struct MainState {
}

impl event::EventHandler<ggez::GameError> for MainState {
        Ok(())
    }

        let mut canvas = graphics::Canvas::from_frame(
            ctx,
            graphics::CanvasLoadOp::Clear([0.1, 0.2, 0.3, 1.0].into()),
        );
        let mut batch = graphics::InstanceArray::new(ctx, None, 64, false);
        batch.push(graphics::DrawParam::new());
        canvas.draw(&batch, graphics::DrawParam::new());
        canvas.finish(ctx)?;

        Ok(())
    }
}

pub fn main() -> GameResult {

    let cb = ggez::ContextBuilder::new("drawing", "ggez");

    let (mut ctx, events_loop) = cb.build()?;

    let state = MainState {};
    event::run(ctx, events_loop, state)
}
```

Full repro example.

**Hardware and Software:**
 - ggez version: 0.8.0-rc0 and aa035d8e
 - OS: Linux Mint 20
 - Graphics card: Nvidia GTX 970
 - Graphics card drivers: Propriety Nvidia driver 470.129

### LINKED ISSUE #1058: InstanceArray doesn't draw all instances.
**Describe the bug**
When I pass in 4 DrawParam's, it draws the first 2. If I only pass the last 2, it only draws the 3rd one.

I assume the data isn't getting passed to WGPU properly.

**Hardware and Software:**
 - ggez version: [d6d41e4](https://github.com/ggez/ggez/commit/d6d41e4a2f308b995d3bdbc71b539e5ae1cbb7e7)

jazzfool:
Strictly internal changes - no user-facing behavior/API is altered
Closes #1054
Closes #1051
Closes #1058 (?)

### DIFF
--- a/src/graphics/internal_canvas.rs
+++ b/src/graphics/internal_canvas.rs
@@ -412,33 +412,12 @@ impl<'a> InternalCanvas<'a> {
             uniforms.as_std140().as_bytes(),
         );
 
-        let (bind_group, _) = BindGroupBuilder::new()
-            .buffer(
-                &instances.buffer,
-                0,
-                wgpu::ShaderStages::VERTEX,
-                wgpu::BufferBindingType::Storage { read_only: true },
-                false,
-                None,
-            )
-            .buffer(
-                &instances.indices,
-                0,
-                wgpu::ShaderStages::VERTEX,
-                wgpu::BufferBindingType::Storage { read_only: true },
-                false,
-                None,
-            )
-            .create(&self.wgpu.device, self.bind_group_cache);
-
-        let bind_group = self.arenas.bind_groups.alloc(bind_group);
-
         self.pass.set_bind_group(
             0,
             self.arenas.bind_groups.alloc(uniform_bind_group),
             &[uniform_alloc.offset as u32],
         );
-        self.pass.set_bind_group(2, bind_group, &[]);
+        self.pass.set_bind_group(2, &instances.bind_group, &[]);
 
         self.pass.set_vertex_buffer(0, mesh.verts.slice(..));
         self.pass
@@ -645,6 +624,7 @@ impl<'a> Drop for InternalCanvas<'a> {
 pub struct InstanceArrayView {
     pub buffer: ArcBuffer,
     pub indices: ArcBuffer,
+    pub bind_group: ArcBindGroup,
     pub image: Image,
     pub len: u32,
     pub ordered: bool,
@@ -655,6 +635,11 @@ impl InstanceArrayView {
         Ok(InstanceArrayView {
             buffer: ia.buffer.lock().map_err(|_| GameError::LockError)?.clone(),
             indices: ia.indices.lock().map_err(|_| GameError::LockError)?.clone(),
+            bind_group: ia
+                .bind_group
+                .lock()
+                .map_err(|_| GameError::LockError)?
+                .clone(),
             image: ia.image.clone(),
             len: ia.instances().len() as u32,
             ordered: ia.ordered,
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -127,14 +127,14 @@ impl Canvas {
             blend_mode: BlendMode::ALPHA,
             premul_text: true,
             projection: glam::Mat4::IDENTITY.into(),
-            scissor_rect: (0, 0, drawable_size.0 as _, drawable_size.1 as _),
+            scissor_rect: (0, 0, target.width(), target.height()),
         };
 
         let screen = Rect {
             x: 0.,
             y: 0.,
-            w: drawable_size.0 as _,
-            h: drawable_size.1 as _,
+            w: drawable_size.0,
+            h: drawable_size.1,
         };
 
         let mut this = Canvas {
@@ -419,7 +419,10 @@ impl Canvas {
         canvas.set_sampler(state.sampler);
         canvas.set_blend_mode(state.blend_mode);
         canvas.set_projection(state.projection);
-        canvas.set_scissor_rect(state.scissor_rect);
+
+        if state.scissor_rect.2 > 0 && state.scissor_rect.3 > 0 {
+            canvas.set_scissor_rect(state.scissor_rect);
+        }
 
         for draws in self.draws.values() {
             for draw in draws {
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -1,7 +1,7 @@
 use super::{
     draw::DrawUniforms,
     gpu::{
-        arc::{ArcBindGroup, ArcBuffer, ArcRenderPipeline, ArcShaderModule},
+        arc::{ArcBindGroup, ArcBindGroupLayout, ArcBuffer, ArcRenderPipeline, ArcShaderModule},
         bind_group::{BindGroupBuilder, BindGroupCache},
         growing::GrowingBufferArena,
         pipeline::PipelineCache,
@@ -18,7 +18,7 @@ use crate::{
     context::Has,
     error::GameResult,
     filesystem::Filesystem,
-    graphics::gpu::pipeline::RenderPipelineInfo,
+    graphics::gpu::{bind_group::BindGroupLayoutBuilder, pipeline::RenderPipelineInfo},
     GameError,
 };
 use ::image as imgcrate;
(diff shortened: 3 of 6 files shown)


PSteinhaus (commented):
Thank you for looking into this!

It's also nice that you refactored the surface configuration a bit. That's way better.

So, if I see correctly, bind groups are now owned by `InstanceArray`s instead of coming out of the arenas, fixing the memory leak.
This made me wonder about the other drawable structures, but I tested `Mesh`, `Image` and `Quad` and they're all fine, even when recreated in the draw loop continously (though recreating `Image` is of course enormously expensive and makes everything lag).

Is there any other place where such a problem could arise? I haven't looked into the arena usage much yet.

src/graphics/gpu/text.rs
```diff
@@ -43,6 +46,13 @@ impl TextRenderer {
         let cache_view =
             ArcTextureView::new(cache.create_view(&wgpu::TextureViewDescriptor::default()));
 
+        let cache_bind = BindGroupBuilder::new().image(&cache_view, wgpu::ShaderStages::FRAGMENT);
+        let cache_bind = ArcBindGroup::new(device.create_bind_group(&wgpu::BindGroupDescriptor {
+            label: None,
+            layout: &cache_bind_layout,
+            entries: cache_bind.entries(),
+        }));
+
```
PSteinhaus, line 55: Is this done so that `TextRenderer` ownes it's cache image bind group? Similar to `Image`?

PSteinhaus (approved):

src/graphics/gpu/text.rs
jazzfool: Yes, since `InternalCanvas` no longer relies on the bind group cache to bind images, all images need to have their own bind group stapled to them. This would've also been leaking anyway as the text renderer's image is recreated whenever the glyph cache needs to be resized.

jazzfool:
N.B.: src now scales draw as per #1052
I agree that this behavior is far more intuitive

PSteinhaus:
> N.B.: src now scales draw as per #1052 I agree that this behavior is far more intuitive

Thanks!
But a question: What do you think, should we make an exception for `Mesh` and [`draw_instanced_mesh`](https://docs.rs/ggez/0.8.0-rc0/ggez/graphics/struct.Canvas.html#method.draw_instanced_mesh), so that these are not automatically scaled when `src` is used?

I know that that would be an inconsistency but I'd argue that it feels more intuitive, as I probably want to draw my geometric shapes the way I specified, using `src` only to select the part o the texture I want to sample from.

jazzfool:
I think we should take it a step further and disable `image_scale` by default for mesh drawing. This would mean `image_scale` would need to become an enum, so that (e.g.) `ImageScale::Default` is `true` for non-mesh drawing and `false` for mesh drawing. This would fully return it to 0.7 functionality (e.g., ex 03_drawing: [0.7](https://github.com/ggez/ggez/blob/6d855cdb0f19df37911b183999cdac7a1015c81a/examples/03_drawing.rs#L161) vs [0.8-rc0](https://github.com/ggez/ggez/blob/684892e43f5b02cf5136bae77bcddc37b2b95a6a/examples/03_drawing.rs#L181)), and in that case (e.g.) `ImageScale::Yes` on mesh drawing would still be affected by the `src` parameter - but the difference is that you're explicitly overriding it in this case so the behaviour shouldn't be surprising.

PSteinhaus:
That sounds like the best solution to me. It gets us back to 0.7 behavior, is more transparent and doesn't need many changes.

An alternative would be to just lose `image_scale` completely, as it's functionally not really necessary but adds a little complexity ([which might always confuse people at first](https://github.com/ggez/ggez/issues/1062)) and then change scale based on image size directly in `Image::draw` and `InstanceArray::draw`, as we did in ggez 0.7 and before, but your proposed solution seems more transparent and therefore a bit more elegant perhaps.

src/graphics/internal_canvas.rs
```diff
@@ -398,7 +405,7 @@ impl<'a> InternalCanvas<'a> {
         let uniforms = InstanceUniforms {
             transform: (self.transform
                 * glam::Mat4::from(
-                    DrawUniforms::from_param(&param.src(Rect::one()), [1., 1.].into()).transform,
+                    DrawUniforms::from_param(&param.src(Rect::one()), None).transform,
```
PSteinhaus, line 408: Please add a comment like "image_scale is None here so that src isn't used for scaling the geometry" for later generations to better understand why there is a scale down there but not simply up here.

src/graphics/internal_canvas.rs
```diff
@@ -407,6 +414,15 @@ impl<'a> InternalCanvas<'a> {
...
             },
+            scale: if scale {
+                glam::Vec2::new(
+                    instances.image.width() as f32,
+                    instances.image.height() as f32,
+                )
+            } else {
+                glam::Vec2::ZERO
+            }
+            .into(),
```
PSteinhaus, line 425: And then please add a comment like "here we pass an additional scale parameter, which we need as we passed None as image_scale above"

PSteinhaus (commented):
Thanks again. I'd add my suggestions as "suggestions", but for some reason Github won't let me.

src/graphics/internal_canvas.rs
```diff
@@ -404,6 +415,17 @@ impl<'a> InternalCanvas<'a> {
                 z: param.color.b,
                 w: param.color.a,
             },
+            // this is the actual image scale that we apply in the vertex shader.
+            // we can't apply this when we first convert the instance array drawparams because we don't know the image size yet.
```
PSteinhaus, line 419: But... we do know `instances.image.width()` and `height()` in line 409 already, don't we?
I thought the reason you didn't use them above was that you didn't want to let the `src` field influence the scaling, but now I realize you already set that to `Rect::one`.

Btw.: is that `src(Rect::one)` necessary still? Now that the `image_scale` parameter is `None`?

jazzfool: We don't know the image size at the time draw params are inserted into the instance array, we only know it now, during the draw.

I suppose forcing src to one isn't necessary anymore, yes.

PSteinhaus: But can't we then just pass the image size as `image_scale` in line 409? (then we'd have to keep `Rect::one` I guess) instead of passing it to the shader this way?

I'm pretty sure you have a reason to do it this way, I just don't understand it yet.

jazzfool: We want to apply image scaling to each instance separately in its local space. If it's applied at world space then it will scale all the vertices of the instances which is undesired (i.e., it will scale the instance positions).

PSteinhaus: Aha! That makes sense.
Could you adapt the lower comment to say so and remove the `src(Rect::one)` above? I think then we should be done.

PR #1061, opened 2022-07-02
Remove/replace undesired panics

PSteinhaus:
We shouldn't panic when there's no unrecoverable error. Instead we should return the errors, so that the user can decide how to react to them.

In some cases I for now opted for simply printing a message instead. This is very primitive and perhaps not ideal. But I'm unsure as to whether we really want to return `GameResult`s on most of `DrawParam`s methods. What I do know is that I don't want these functions to panic.

Instead of printing we could also use `warn!`, but many users won't see these messages, as they probably won't use a logger to listen to them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,9 @@ The following list doesn't repeat the changes already mentioned above.
 * Improved `Text` performance through better glyph re-use
 * Changed the `Drawable` trait; this will downstream require changes in projects like `ggez-egui`
 * Version bumped `zip` to 0.6, `directories` to 4.0.1, `winit` to 0.26, image to `0.24` and `rodio` to 0.15
+* `DrawParam` methods such as `dest`, `scale`, etc. no longer panic when used on a `DrawParam` which holds the matrix
+  `Transform` variant; instead the matrix is lost and a new default `Transform` is created
+
 
 ## Deprecated
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -6,6 +6,9 @@ use std::sync::Arc;
 /// An enum containing all kinds of game framework errors.
 #[derive(Debug)]
 pub enum GameError {
+    /// An error when trying to get or set values from a `DrawParam` that are no longer available
+    /// because it has already been crunched down into a Matrix
+    DrawParamMatrixError,
     /// An error when intializing the graphics system.
     GraphicsInitializationError,
     /// An error in the filesystem layout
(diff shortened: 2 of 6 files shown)


nobbele:
this would just make the graphics display wrongly rather than panic, I don't see that as an improvement

PSteinhaus:
Ok, true. Then how about returning a `GameResult<Self>` instead of `Self`? It would hurt ergonomics, but only slightly, as an added `?` should be enough in most cases (`draw` returning `GameResult` by default and all).

I just don't want to panic here as I see no reason to necessarily crash the whole program just because some draw might not be performed correctly.

nobbele:
How about making matrix draw param a separate type, either by generic or a completely separate type so we dont have to implement the functions for matrix type draw params?

PSteinhaus:
That way we don't have to worry about what our `Transform` is all the time. Sounds good.

Then i guess a generic `DrawParam` would be the way to go? With the methods in question only implemented on `DrawParam<ParamValues>` and not on `DrawParam<ParamMatrix>`?

Creating two unique structs `DrawParamMatrix` and `DrawParamClassic` both implementing a new trait `DrawParam` sounds a bit ugly to me, as both names are just too long to be used so often.

PSteinhaus:
Ok, I just took a stab at implementing this and realized that we need to make the `transform` field dynamic for this approach to work, otherwise `InstanceArray` can't use one `Vec` to store all its `DrawParam`s... this makes it all a bit more ugly... I really don't want to make something used as massively as `DrawParam` dynamic if I can help it.

This problem would arise in both cases: When we use two independent structs or when we use generics.

nobbele:
I'm still not so sure about resetting but I don't mind it. Maybe keep the warnings though?

PSteinhaus:
Yeah, we should keep the warnings, I agree.
Question: Do we want to `println!`, `warn!`, `eprintln!` or do something else entirely there?

nobbele:
The thing with panic or error is whether failure is valid use or not. I would consider calling transform operations on a matrix drawparam to be a misuse of the type and therefore should panic. I looked at what other people think of whether to panic or not: "[Your library documents a precondition of a public API item that, when not met, causes a panic. Therefore, the user of your library has misused your library, and their code has a bug.](https://www.reddit.com/r/rust/comments/9x17hn/when_should_a_library_panic_vs_return_result/e9p5c9t/)" which seems to agree with what I said.

If it were to return an error: There's no reason for transform operations to be called from due to user interaction or anything so there's not really any use case in which you'd need to or be able to handle it anyway.

If it were to reset and warn: It would put the drawparam in an unexpected state and log a less noticeable message rather than being instantly informed about the misuse.

What I think we should do though is to thoroughly document that the methods can't be used on matrices.

PSteinhaus:
Hmm. I dislike panics, but I have to agree that they're probably the better choice here, compared to silently doing unexpected things. Meh.

Well then. I'll dump this PR and open a new one which simply informs about the panics instead.

PR #1071, opened 2022-07-26
reexported glam

### LINKED ISSUE #1067: Export glam
Fixes version mismatches
Resolves the issue of users copying code from examples and wondering why it doesn't compile

PSteinhaus:
what it says in the title
Closes #1067 

Still opened a PR for it, though it's only a single line, so we can directly decide over it here.

### DIFF
--- a/resources/dimmer.wgsl
+++ b/resources/dimmer.wgsl
@@ -11,10 +11,10 @@ struct Dim {
 [[group(1), binding(0)]]
 var t: texture_2d<f32>;
 
-[[group(1), binding(1)]]
+[[group(2), binding(0)]]
 var s: sampler;
 
-[[group(3), binding(0)]]
+[[group(4), binding(0)]]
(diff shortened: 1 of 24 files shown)


nobbele:
Should update examples to use this export

PSteinhaus:
Good point. Will do.

resources/dimmer.wgsl
```diff
@@ -11,10 +11,10 @@ struct Dim {
 [[group(1), binding(0)]]
 var t: texture_2d<f32>;
 
-[[group(1), binding(1)]]
+[[group(2), binding(0)]]
 var s: sampler;
 
-[[group(3), binding(0)]]
```
nobbele, line 17: Why did this change? Was it broken before

PSteinhaus:
It was. You can check the current example. It doesn’t run.

PR #1081, opened 2022-07-30
Android

nobbele:
Annoyingly it got based off of update-deps PR..


PSteinhaus:
`wgpu` 0.14, that we now use, depends on `winit` 0.27.1.
So it seems the opportunity for Android support has opened : )

PR #1090, opened 2022-08-31
shadow example fix and crevice re-export

PSteinhaus:
fixed shadows example;
re-exported crevice as a whole for the AsStd140 macro to work outside of ggez;

The thing is: Whatever I do, I still can't use the `AsStd140` macro outside of ggez (i.e. when creating a test project and depending on ggez, but not crevice directly). Whenever I try I only get this:

```
error[E0433]: failed to resolve: could not find `crevice` in the list of imported crates
  --> src\main.rs:13:10
   |
13 | #[derive(AsStd140)]
   |          ^^^^^^^^ could not find `crevice` in the list of imported crates
   |
   = note: this error originates in the derive macro `AsStd140` (in Nightly builds, run with -Z macro-backtrace for more info)
```
no matter if I `use ggez::graphics::crevice` or not

### DIFF
--- a/examples/shader.rs
+++ b/examples/shader.rs
@@ -2,7 +2,7 @@
 
 use ggez::glam::Vec2;
 use ggez::graphics::{self, Color, DrawMode};
-use ggez::{event, graphics::AsStd140};
+use ggez::{event, graphics::crevice::std140::AsStd140};
 use ggez::{Context, GameResult};
 use std::env;
 use std::path;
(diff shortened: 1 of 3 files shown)


nobbele:
Attempted the crevice re-export, couldn't find any way to do it either

PSteinhaus:
Then I'd say we simply don't re-export `crevice` for now and instead put comments into the example telling that you have to depend on the same version of it which `ggez` is using.

PR #1102, opened 2022-09-27
Fixed text example

PSteinhaus:
The new bounds no longer taking a desination point means that we have to adapt the text example a little.

My change fixes the example and is OK I think, but we could perhaps go another way and collect destinations together with texts while in `update`, so that we only need to draw what we have in `draw`, without that weird little if.

I also removed the reference to `queue` (and that you should use it for better performance, very happy that these times are gone), now that it's just no longer there :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,9 @@ The following list doesn't repeat the changes already mentioned above.
   done now), as it makes things like centering text on positions easier (see the blend modes example)
 * Also `Text` is now a first class citizen and can be drawn normally with `DrawParam`, implementing things like rotation
  that weren't possible in batched text rendering before
+* Changed how bounds on Text work as well as layouting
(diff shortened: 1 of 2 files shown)


nobbele:
Hm this is tricky

PSteinhaus:
What do you mean? `screen_coordinates` already has been re-added:
https://github.com/ggez/ggez/blob/a0069e83a3912a64400536bfabb19eeeb135b38d/src/graphics/canvas.rs#L294

nobbele:
not canvas, glyph_brush section

chamons (approved):

PSteinhaus:
> not canvas, glyph_brush section

Ah, ok sry. But I still don't get what you mean. Re-add what? And screen_coordinate used where, in which version?

nobbele:
You might be right yeah

nobbele:
changing h_align from End to Begin seems wrong but maybe not, I'll check tomorrow

PR #1103, opened 2022-09-27
Adds screen_position back to glyph_brush

nobbele:
This seems to work identical to before, TextAlign aligns to the bounds (or 0 if infinite bound, sounds a bit inconsistent but I can't think of a better value)

Should be equivalent to #1104 when alignment == anchor


chamons (approved):
I can validate that this works, since my PR duplicated this diff from #1104

PR #1109, opened 2022-10-16
fix InstanceArray::new_ordered typo

Calvin304:
i can provide an example program this fixes if necessary.


PSteinhaus:
Hah! Thank you!

That would have been an embarrasing bug to have in 0.8.0.


> i can provide an example program this fixes if necessary.

Thanks, but I think we can all immediately agree that this fix is a good idea :D

Issue #1118, opened 2022-10-21
Default shader params sampler


andrewbaxter:
could we pass the canvas's current sampler by default perhaps?

_Originally posted by @nobbele in https://github.com/ggez/ggez/pull/1115#discussion_r1001815442_

PSteinhaus:
I'd like to keep it explicit. Less opportunity for confusion and not too ugly.

andrewbaxter:
Ah okay. :+1:

andrewbaxter:
Feel free to close this and the PR.

nobbele:
I agree that it's not too bad as is

PR #1119, opened 2022-10-21
Default shader params sampler

andrewbaxter:
To be stacked on https://github.com/ggez/ggez/pull/1115 for #1118

I assume the intent is to simplify the interface with reasonable default behavior.  It didn't look easy to get the sampler from the canvas (would the user necessarily have a canvas during setup?) so I opted to make the default sampler into a new function and use that as the default.

Some other options were
* Take canvas as an argument to the builder
* ~If there are zero samplers, use the canvas's current sampler (but would prevent user from having zero samplers... would that ever be wanted.~  Edit: I realize we need to use the sampler in `build()` so I think this wouldn't work, or would be more or less the same as the above.

### DIFF
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -122,7 +122,7 @@ impl Canvas {
             params: None,
             text_shader: default_text_shader(),
             text_params: None,
-            sampler: Sampler::linear_clamp(),
+            sampler: default_sampler(),
             blend_mode: BlendMode::ALPHA,
             premul_text: true,
             projection: glam::Mat4::IDENTITY.into(),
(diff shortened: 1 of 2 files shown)


nobbele:
Thanks for trying this but we'll keep as is for now

src/graphics/canvas.rs
```diff
@@ -590,3 +590,8 @@ pub fn default_text_shader() -> Shader {
         vs_module: None,
     }
 }
+
+/// The default sampler
+pub fn default_sampler() -> Sampler {
+    Sampler::linear_clamp()
+}
```
PSteinhaus, line 597: That part we should definitely keep though.

PSteinhaus: Ok, solved in https://github.com/ggez/ggez/pull/1121 by implementing `Default` for `Sampler`.

PR #1133, opened 2022-11-04
fix buffer write queue order

PSteinhaus:
Wrote a function to replace `wgpu.queue.write_buffer` with in cases where writes should be placed in correct command encoder order using a `StagingBelt`.

If we can agree that such a function (the name can be changed of course) is better (in terms of easier to use) than exposing the `StagingBelt::write_buffer` API, then I'll go over all the usages of `wgpu.queue.write_buffer` and change them to this function where appropriate.

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -24,6 +24,7 @@ use crate::{
 use ::image as imgcrate;
 use crevice::std140::AsStd140;
 use glyph_brush::FontId;
+use std::num::NonZeroU64;
 use std::{collections::HashMap, path::Path, sync::Arc};
 use typed_arena::Arena as TypedArena;
 use winit::{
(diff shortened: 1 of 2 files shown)


src/graphics/context.rs
```diff
@@ -527,6 +528,37 @@ impl GraphicsContext {
         self.fcx.as_mut().map(|fcx| &mut fcx.cmd)
     }
 
+    /// Write to a wgpu using the [`wgpu::StagingBelt`] of the context, inserting your write into
+    /// the current [`wgpu::CommandEncoder`], making sure the order of operations inside a render pass
```
0----0, line 532: From my understanding, using `StagingBelt`s will not help with the order of operations *within a render pass*, because a `StagingBelt` cannot be used during a render pass--the `CommandEncoder` is borrowed by the `RenderPass`.  From the perspective of a `CommandEncoder`, I think a render pass might as well be considered "atomic" -- it's basically one command.

The `StagingBelt` may still be useful for other purposes, but I think this description is incorrect.

PSteinhaus: From my understanding of the [documentation of `write_buffer`](https://docs.rs/wgpu/latest/wgpu/util/struct.StagingBelt.html#method.write_buffer) this is how I interpret 

> The upload will be placed into the provided command encoder.

Why and how should it be placed inside the command encoder ignoring its queue order?

Ah, wait sorry, now I get what you mean.

The thing is:
We don’t actually need to use the staging belt `write_buffer` function during the lifetime of a `wgpu::RenderPass`, as these are only created on calling `Canvas::finish` inside the resulting `InternalCanvas` (maybe we should just rename everything from “canvas” to “render pass” now…).

And the buffers used there are coming from our `GrowingBufferArena`, as you pointed out, so we don’t need to worry about duplicate use during a pass there.

For the instance array and shader params problem this means we can use a staging belt as the command encoder isn’t borrowed yet.

The only two cases where we indeed couldn’t use it would be `InternalCanvas::set_projection` and `InternalCanvas::flush_text` (though I’m not sure whether the latter is actually broken right now).

PSteinhaus: Ah, ok, no. You’re right.
During the renderpass we can’t queue staging belt operations, so we can’t smuggle in a buffer write this way say, between two draws of an instance array, meh.

Then taking a look into whether this could be solved via a growing buffer arena is indeed a good idea.

PSteinhaus:
superseded by #1139

PR #1139, opened 2022-11-15
remove all aliased buffer overwrites

### LINKED ISSUE #1129: renderpasses can be sabotaged by events earlier in the wgpu queue
**Describe the bug**
Multiple draws & clear of a InstanceArray in a single frame draws only the last call

**To Reproduce**
I think `https://github.com/Bowarc/ggez0.8InstanceArrayBugSortOf` would explain wayy better that the following but.

let's call the first `ggez::graphics::InstanceArray` `IA1` and the second one `IA2`
if you use a `IA1` and `IA2` on a canvas, clearing the `IA1` and `IA2` and then using `IA1` on another canvas (In the same frame) it causes the first canvas to not show the draw of the `IA1` (**only** the `IA1`, `IA2` is not affected)

**Expected behavior**
I expected to see draws of `IA1` on both canvases

I made a quick usable code of ~150lines so you can test it: https://github.com/Bowarc/ggez0.8InstanceArrayBugSortOf

i have a quick video of my game prototype where i got this bug for the first time
https://cdn.discordapp.com/attachments/898362796658532373/1037228027253366814/2022-11-02_05-52-14.mp4

**Hardware and Software:**
 - ggez version: 0.8.1
 - OS: Windows10
 - Graphics card:  NVIDIA GeForce RTX 2070
 - Graphics card drivers: Default W10 Nvidia driver, { version: 526.47,  release date: 10/27/2022 }

### LINKED ISSUE #1137: Setting shader params twice does nothing
**Describe the bug**
If I draw two meshes with the same shader, but different parameters on each shader, the first set of parameters is used for both meshes.

**To Reproduce**
```rust
//! A very simple shader example.

use crevice::std140::AsStd140;
use ggez::event;
use ggez::glam::Vec2;
use ggez::graphics::{self, Color, DrawMode};
use ggez::{Context, GameResult};
use std::env;
use std::path;

#[derive(AsStd140)]
struct Dim {
    rate: f32,
}

struct MainState {
    dim: Dim,
    shader: graphics::Shader,
    params: graphics::ShaderParams<Dim>,
}

impl MainState {
    fn new(ctx: &mut Context) -> GameResult<MainState> {
        let dim = Dim { rate: 0.5 };
        let shader = graphics::ShaderBuilder::new_wgsl()
            .fragment_path("/dimmer.wgsl")
            .build(&ctx.gfx)?;
        let params = graphics::ShaderParamsBuilder::new(&dim).build(&mut ctx.gfx);
        Ok(MainState {
            dim,
            shader,
            params,
        })
    }
}

impl event::EventHandler<ggez::GameError> for MainState {
        self.dim.rate = 0.5 + (((ctx.time.ticks() as f32) / 100.0).cos() / 2.0);
        Ok(())
    }

        let mut canvas = graphics::Canvas::from_frame(ctx, Color::from([0.1, 0.2, 0.3, 1.0]));

        let circle = graphics::Mesh::new_circle(
            ctx,
            DrawMode::fill(),
            Vec2::new(100.0, 300.0),
            100.0,
            2.0,
            Color::WHITE,
        )?;
        canvas.draw(&circle, Vec2::new(0.0, 0.0));

        self.params.set_uniforms(ctx, &self.dim);
        canvas.set_shader(self.shader.clone());
        canvas.set_shader_params(self.params.clone());
        let circle = graphics::Mesh::new_circle(
            ctx,
            DrawMode::fill(),
            Vec2::new(400.0, 300.0),
            100.0,
            2.0,
            Color::WHITE,
        )?;
        canvas.draw(&circle, Vec2::new(0.0, 0.0));

        let uniforms = Dim {
            rate: self.dim.rate * 0.5,
        };
        let params = graphics::ShaderParamsBuilder::new(&uniforms).build(&mut ctx.gfx);
        canvas.set_shader_params(params);
        let circle = graphics::Mesh::new_circle(
            ctx,
            DrawMode::fill(),
            Vec2::new(700.0, 300.0),
            100.0,
            2.0,
            Color::WHITE,
        )?;
        canvas.draw(&circle, Vec2::new(0.0, 0.0));

        canvas.finish(ctx)
    }
}

pub fn main() -> GameResult {
    let resource_dir = if let Ok(manifest_dir) = env::var("CARGO_MANIFEST_DIR") {
        let mut path = path::PathBuf::from(manifest_dir);
        path.push("resources");
        path
    } else {
        path::PathBuf::from("./resources")
    };

    let cb = ggez::ContextBuilder::new("shader", "ggez").add_resource_path(resource_dir);
    let (mut ctx, event_loop) = cb.build()?;

    let state = MainState::new(&mut ctx)?;
    event::run(ctx, event_loop, state)
}
```

**Expected behavior**
The rightmost circle should have half the brightness of the middle circle.

image

**Hardware and Software:**
 - ggez version: 0.8.1
 - OS: Windows 11
 - Graphics card: Radeon 6900 XT

jazzfool:
closes #1129

All instances where wgpu.queue.write_buffer is used now only write to buffers that have been freshly allocated from an arena. Previously, in some instances (such as in ShaderParams), existing buffers regions were overwritten at the start of the queue, causing #1129.

Also fixes #1137 - This had nothing to do with the buffer overwriting issue, we were just missing a dirty_pipeline flag.

The only queued write that isn't reallocated is the queue.write_texture in TextRenderer, but this is fine as we operate under the condition that glyph_brush never tells us to write to a region of the glyph cache texture that's already been written to (which would in itself be problematic :P).

Regarding the original issue of #1129, that is InstanceArray, the fix was very simple; unconditionally executing the statement previously guarded by resizing the buffers when capacity changes. Using an arena here doesn't really make sense as between different mutations an IA can have wildly different buffer sizes and would be a pain to actually implement. I'm only hesitant of the performance effects -- Memory performance should be okay (given what the goal is -- allowing the user to change data between draws within a single pass -- buffers and data need to be duplicated, there's no getting around that, and the excess should be dropped by the end of the frame anyway), but the problem is that this could happen needlessly if the user isn't aware that IAs are effectively duplicated in memory if they're mutated between draws. Attention should be drawn in docs that IA mutations should be done in as large and early chunks as possible.

ShaderParams now have a local arena with a preset buffer size (UPDATES_PER_ARENA -- currently 16). The goal here is to minimize how many new bind groups need to be created by minimizing how many new buffers are encountered in combination with a specific ShaderParams instance - avoiding a sort of combinatorial explosion at the bind group cache. Ideally we would have ShaderParams be sent in two separate bind groups but as I understand this isn't possible due to the need to support lower-end hardware and also possibly for user experience when writing shaders.

### DIFF
--- a/src/graphics/shader.rs
+++ b/src/graphics/shader.rs
@@ -1,22 +1,19 @@
 use std::io::Read;
 use std::marker::PhantomData;
 
-use crate::{
-    context::{Has, HasMut},
-    GameError, GameResult,
-};
+use crate::{context::Has, Context, GameError, GameResult};
 
 use super::{
     context::GraphicsContext,
     gpu::{
-        arc::{ArcBindGroup, ArcBindGroupLayout, ArcBuffer, ArcShaderModule},
+        arc::{ArcBindGroup, ArcBindGroupLayout, ArcSampler, ArcShaderModule, ArcTextureView},
         bind_group::BindGroupBuilder,
+        growing::GrowingBufferArena,
     },
     image::Image,
     sampler::Sampler,
 };
 use crevice::std140::Std140;
-use wgpu::util::DeviceExt;
 
 #[derive(Debug, PartialEq, Eq)]
 enum ShaderSource<'a> {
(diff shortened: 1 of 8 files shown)


nobbele:
Didn't test it yet, but sounds great!
Could you keep the same public API? i.e keep shader params as move rather than reference

jazzfool:
Restored Clone on ShaderParams with Arc Mutex, but strongly recommend that this be changed back upon next breaking release as there's no reason for ShaderParams to be a cloneable shared resource (other than API stability)

nobbele:
yeah I agree with that, I'll be quite busy this weekend so I'm not sure I'll have to look at this yet

nobbele:
`ShaderParamsBuilder::build` seems to have changed signature for no reason? both that and `ShaderParamsInner::set_uniforms` can take a `Has<GraphicsContext>`

src/graphics/shader.rs
```diff
@@ -231,23 +231,89 @@ impl<'a, Uniforms: AsStd140> ShaderParamsBuilder<'a, Uniforms> {
...
-    pub fn build(self, gfx: &mut impl HasMut<GraphicsContext>) -> ShaderParams<Uniforms> {
-        let gfx = gfx.retrieve_mut();
-        let uniforms = ArcBuffer::new(gfx.wgpu.device.create_buffer_init(
-            &wgpu::util::BufferInitDescriptor {
-                label: None,
-                usage: wgpu::BufferUsages::UNIFORM | wgpu::BufferUsages::COPY_DST,
-                contents: self.uniforms.as_std140().as_bytes(),
-            },
-        ));
+    pub fn build(self, ctx: &mut Context) -> ShaderParams<Uniforms> {
```
nobbele, line 234: should've written the comment in the review

nobbele (changes requested):

nobbele:
can confirm it solves the issue

src/graphics/shader.rs
jazzfool: ctx.time is used in set_uniforms (which is called by build)

Issue #1142, opened 2022-11-25
game_id should set app-id on Wayland


snyball:
On Wayland (Linux/BSD) windows have an app-id attribute, which is separate from the title. winit supports setting this attribute using `.with_name` on `WindowBuilder`. ggez should support setting this attribute to the `game_id` given to `ContextBuilder::new`, which should then be (build-time conditionally based on the same predicate as in winit) forwarded to winits `.with_name` in `GraphicsContext::new_from_instance`. I think this is the least surprising behavior for `game_id` on Wayland.

My primary use case for having to set this attribute is to be able to keep `cargo watch -x 'run'` running in the background, and to have my window-manager automatically move the app window to a specific workspace on restart. An app-id attribute is necessary to do this, at least in sway wm.

I can create and submit a pull request for this, but am checking if maintainers agree first :)

PSteinhaus:
Thanks for the observation!
That's a good idea. I agree that setting it to `game_id` makes sense.

If you want to submit a PR for this I'll gladly take a look over it :)

PR #1143, opened 2022-11-26
Set Wayland app-id based on ggez game_id

snyball:
This implements the feature described in #1142, it has been tested and confirmed to set the app-id correctly in sway wm, and does not break builds for non-supported platforms by using the same `cfg` directive that winit uses.


PSteinhaus (approved):

PSteinhaus:
Looks good!
A quick `rustfmt` and we should be good to go :)

PR #1146, opened 2022-11-30
fix non-instanced draw calls - use as_std140 instead of bytemuck for uniforms

sphenlee:
Fixes the issues as discussed on Reddit and Discord:
https://www.reddit.com/r/rust_gamedev/comments/z5xhi9/ggez_081_only_text_is_rendering_help_needed_please/


nobbele:
uniforms should be Std140 too? I need read up on that again

qwitwa:
https://stackoverflow.com/questions/16270846/when-should-i-use-std140-in-opengl

PR #1166, opened 2023-01-31
Fix examples

a1phyr:
They should use the new API


PSteinhaus:
Good catch. Thanks for fixing all of these lints as well.

PR #1170, opened 2023-02-08
Revert "Update wgpu to 0.15"

nobbele:
Reverts ggez/ggez#1165

Just noticed it broke something severely


nobbele:
nvm github doesn't work like I want it to

PR #1175, opened 2023-02-17
fix text buffer overwrite

jazzfool:
sometimes the text buffer would not be recreated and thus overwritten
reverts to arena as before, but now falling back to a dedicated buffer if the text is too big


tversteeg:
Thanks for the PR!

PR #1186, opened 2023-04-16
doc: Added dependencies for fedora into build guide

arminveres:
Some fedora packages are called quite differently than in Debian based distributions. Therefore the corresponding package names were added.

**Please only open PRs that are based on and aimed towards the `devel` branch**

**For more information take a look at [the contribution guide](./CONTRIBUTING.md)**

### DIFF
--- a/docs/BuildingForEveryPlatform.md
+++ b/docs/BuildingForEveryPlatform.md
@@ -68,6 +68,12 @@ least you can install them with:
 yum install alsa-lib-devel
 ```
 
+### Fedora
+
+```sh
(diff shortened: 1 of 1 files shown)


nobbele:
is it actually different from RHEL/CentOS?

arminveres:
`alsa-lib-devel pkgconf-pkg-config` are shared, you are right, but [libudev](https://pkgs.org/download/rust-libudev-devel) is not. I have not found the packages for CentOS/RHEL. 
Though don't you think, it makes sense to fully list all packages for the distribution , even if it overlaps with some of the others? It makes it a tad easier to just copy over.

nobbele:
yeah if it's different then sure

Issue #1156, opened 2023-01-23
No way to create a blank image?


nobbele:
There's `from_solid` but that makes a square only?
`new_canvas_image` does the right thing though

PSteinhaus:
implemented in #1189

Is that what you had in mind?
I'd say we could also just rename this new `new_blank_canvas_image` function to `from_solid`, replacing the current `from_solid`.
I agree that having a few more simple controls/options on `from_solid` as a debug function makes sense, at least to allow it to create other shapes then squares.

Or would you prefer something more akin to a wrapper around `new_canvas_image`, just with `format` and `samples` hardcoded for easier use?

tversteeg:
I would prefer `from_color` over `from_solid` or maybe `new_solid_color` but the current one is also good.

nobbele:
I actually forgot exactly what I meant/wanted here

PSteinhaus:
> I would prefer `from_color` over `from_solid` or maybe `new_solid_color` but the current one is also good.

I think that `from_color` makes the most sense, too.
I made a little PR renaming it like this and removing `from_solid`, as it's the same function, just without the option to specify both dimensions: #1194

PR #1189, opened 2023-05-02
feat: create blank canvas

ZimboPro:
PR for #1156 

Not too sure if this is what was the intended solution based on the ticket

### DIFF
--- a/src/graphics/image.rs
+++ b/src/graphics/image.rs
@@ -49,6 +49,28 @@ impl Image {
         )
     }
 
+    /// A little helper function that creates a blank `Image` that is of the given width and height and optional color. The default color is `Color::WHITE`. Mainly useful for debugging.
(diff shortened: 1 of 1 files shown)


src/graphics/image.rs
```diff
@@ -49,6 +49,28 @@ impl Image {
         )
     }
 
+    /// A little helper function that creates a blank `Image` that is of the given width and height and optional color. The default color is `Color::WHITE`. Mainly useful for debugging.
```
tversteeg, line 52:
```suggestion
    /// A little helper function that creates a blank [`Image`] that is of the given width and height and optional color.
    /// 
    /// The default color is [`Color::WHITE`].
    /// Mainly useful for debugging.
```

tversteeg:
Thanks for the PR!

PR #1191, opened 2023-05-04
Use paths instead of strings where it makes sense

a1phyr:


src/filesystem.rs
```diff
@@ -127,8 +127,8 @@ impl Filesystem {
     pub fn new(
         id: &str,
         author: &str,
-        resources_dir_name: &str,
-        resources_zip_name: &str,
+        resources_dir_name: &path::Path,
```
tversteeg, line 130: Can you make these `impl Into<PathBuf>` too?

a1phyr: We don't need to allocate a `PathBuf` here, so I used `AsRef<Path>` instead

PR #1194, opened 2023-05-25
from_color

PSteinhaus:
renamed `new_blank_canvas_image` to `from_color` and removed `from_solid`, as `from_color` is just a more properly named replacement with the ability to specify width and height.

### DIFF
--- a/examples/colorspace.rs
+++ b/examples/colorspace.rs
@@ -99,7 +99,7 @@ impl MainState {
             graphics::Rect::new(0.0, 0.0, 400.0, 400.0),
             Color::WHITE,
         )?;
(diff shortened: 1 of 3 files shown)


tversteeg:
Is this OK as a breaking API change, or are there already other changes that also requires a new major version?

PSteinhaus:
Since we are still about to release 0.9.0 I'd say merging this breaking change into devel, so that it will be part of 0.9.0 is unproblematic.

a1phyr:
Should we really keep `Option<Color>` as a parameter or could we switch to `Color` ?

PSteinhaus:
I like that idea. It would make the examples more readable and is also more explicit.

PR #1203, opened 2023-06-27
Allow custom contexts!

Vixeliz:
This will allow custom context to be made for even more customization of how ggez works! Of course a default context will still be provided. Minimum changes to the user api due to this change has occured.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -410,8 +410,7 @@ impl ContextBuilder {
         ) -> GameResult<(C, winit::event_loop::EventLoop<()>)>,
     ) -> GameResult<(C, winit::event_loop::EventLoop<()>)>
     where
-        C: std::any::Any
-            + HasMut<ContextFields>
+        C: HasMut<ContextFields>
             + HasMut<timer::TimeContext>
             + HasMut<input::keyboard::KeyboardContext>
             + HasMut<input::mouse::MouseContext>
--- a/src/context.rs
+++ b/src/context.rs
@@ -85,7 +85,7 @@ pub struct ContextFields {
     /// The Conf object the Context was created with.
     /// It's here just so that we can see the original settings,
     /// updating it will have no effect.
-    pub(crate) conf: conf::Conf,
+    pub conf: conf::Conf,
     /// Controls whether or not the event loop should be running.
     /// This is internally controlled by the outcome of [`quit_event`](crate::event::EventHandler::quit_event),
     /// requested through [`event::request_quit()`](crate::Context::request_quit).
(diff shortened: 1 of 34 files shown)


Vixeliz:
I forgot to change all the examples over I will do that in a new commit here

src/event.rs
```diff
@@ -95,24 +101,25 @@ pub enum ErrorOrigin {
 ///
 /// For the error type simply choose the default [`GameError`](../error/enum.GameError.html),
 /// or something more generic, if your situation requires it.
-pub trait EventHandler<E = GameError>
+pub trait EventHandler<C, E = GameError>
 where
     E: std::fmt::Debug,
+    C: std::any::Any + HasMut<ContextFields> + HasMut<input::mouse::MouseContext>,
```
nobbele, line 107: why the Any bound?

src/event.rs
```diff
@@ -95,24 +101,25 @@ pub enum ErrorOrigin {
 ///
 /// For the error type simply choose the default [`GameError`](../error/enum.GameError.html),
 /// or something more generic, if your situation requires it.
-pub trait EventHandler<E = GameError>
+pub trait EventHandler<C, E = GameError>
```
nobbele, line 104: Default this to the standard context type

src/context.rs
```diff
@@ -344,17 +399,44 @@ impl ContextBuilder {
...
+        from_conf: F,
+    ) -> GameResult<(C, winit::event_loop::EventLoop<()>)>
+    where
+        F: Fn(String, conf::Conf, Filesystem) -> GameResult<(C, winit::event_loop::EventLoop<()>)>,
+        C: std::any::Any
+            + HasMut<ContextFields>
+            + HasMut<timer::TimeContext>
+            + HasMut<input::keyboard::KeyboardContext>
+            + HasMut<input::mouse::MouseContext>
+            + HasMut<GamepadContext>,
```
nobbele, line 418: Should probably make
```rust
macro_rules! trait_alias {
    ($name:ident = $base1:ident + $($base2:ident +)+) => {
        trait $name: $base1 $(+ $base2)+ { }
        impl<T: $base1 $(+ $base2)+> $name for T { }
    };
}

// The last + is actually needed
trait_alias!(HasMutCommonContexts, HasMut<ContextFields> + /* ... */ + HasMut<timer::TimeContext> +)
```

nobbele (changes requested):

src/event.rs
Vixeliz: Ah that was before I added the other bounds kind of slipped my mind. I'll fix that up

nobbele (approved):
Nice

PR #1208, opened 2023-06-28
Split up ggez into multiple crates

Vixeliz:
This would make adding more features easier while also not inflating one massive crate over time. It should also help with compile times.


Cargo.toml
```diff
@@ -19,57 +19,52 @@ categories = ["game-engines"]
...
-mint = "0.5.9"
+ggez_core = { path = "crates/ggez_core" }
+ggez_error = { path = "crates/ggez_error" }
+ggez_conf = { path = "crates/ggez_conf" }
+ggez_time = { path = "crates/ggez_time" }
+ggez_filesystem = { path = "crates/ggez_filesystem" }
+ggez_traits = { path = "crates/ggez_traits" }
+ggez_graphics = { path = "crates/ggez_graphics" }
+ggez_audio = { path = "crates/ggez_audio", optional = true }
+ggez_3d = { path = "crates/ggez_3d", optional = true }
```
a1phyr, line 64: You should probably put these in `[workspace.dependencies]` and specify them with the workspace syntax here, so you won't have to deal with paths in sub-crates

Vixeliz: Good point, i'll do this whenever I work on this pr next if I do. General reception seems so so to this. I think it would be nice but atm this pr kind of sucks

Vixeliz:
I'm gonna close this, this pr is very yuck. I'm still interested in this but not atm and not with this pr.

Issue #1211, opened 2023-06-29
Potential gpu memory leak


Bowarc:
**Describe the bug**
[The original message](https://discord.com/channels/1031224392174293002/1031224393126387764/1123947796102389860)
My gpu memory is getting filled fast when creating a new graphics::Image every ggez::event::EventHandler::Draw call

**To Reproduce**
I don't rly know why, i guess it's the fact that i'm creating a new `ggez::graphics::Image` every frame ?

**Expected behavior**
have my graphic memory not fill and my pc staying alive more than 10 secconds

**Screenshots or pasted code**
[This is the (ugly) code that lead to the problems ](https://github.com/Bowarc/ggez_potential_gpu_leak)

**Hardware and Software:**
 - ggez version: 0.9.0
 - OS: w 10.0.19045
 - Graphics card:  NVidia Geforce GTX 2070.
 - Graphics card drivers: up to date (driver version: 536.23)

Bowarc:
I don't know if that's relevant but the game im making (Vupa)
While standing still, render calls doesn't fluctuate, no new image (nor sound but i don't think that matters) are loaded,
This is the fame stat debug: (stays consistent across time in the test)
![image](https://github.com/ggez/ggez/assets/63136904/8dbe08e2-d62c-457e-9be9-3f249bafb569)

The render loop keeps loosing frames per sec.

Timings at the start:
Game render request fetching: 100-200µs,
Frame debug stat render request fetching: 118-150µs,
Gui render request fetching:  100-100ns,
Request rendering: 800-990+µs,

Timings at the end:
Game render request fetching: ~140µs,
Frame debug stat render request fetching: ~140µs
Gui render request fetching:  ~100ns,
Request rendering:  7ms,

The Request rendering only creates canvases(one for each game layer (map is drawn before characters etc..)), and render images, meshes, text, and egui windows on thoses canvases.
Then `.finish()` the canvases.

I gathered more infos with an external tool (process hacker).
Gpu stats across time
![image](https://github.com/ggez/ggez/assets/63136904/01f7b21a-116a-4a02-9ace-da53f8e95ab2)

fps infos ?
![image](https://github.com/ggez/ggez/assets/63136904/b2579ef3-2334-433d-9f7e-a78318055325)

Cpu stats across time
![image](https://github.com/ggez/ggez/assets/63136904/0edfc753-4129-43c3-8546-7705e23ee4ca)


Hope that helps :c

Bowarc:
[...]

Issue #1181, opened 2023-03-12
Memory leak when drawing an image


ToBinio:
**Describe the bug**
The system memory & gpu memory leaks when drawing an image.
I think this only happens when drawing the image for the first time.

this bug could be connect to #1128

or maby i am doing something wrong

**To Reproduce**
run this code best in release mode
```
use ggez::graphics::{DrawParam, Image, ImageFormat};
use ggez::{event, graphics, Context, GameResult};
use image::{DynamicImage, EncodableLayout};

const IMG_SIZE: u32 = 8000;

struct MainState {}

impl MainState {
    fn new(_ctx: &mut Context) -> GameResult<MainState> {
        Ok(MainState {})
    }
}

impl event::EventHandler<ggez::GameError> for MainState {
    fn update(&mut self, _ctx: &mut Context) -> GameResult {
        Ok(())
    }

    fn draw(&mut self, ctx: &mut Context) -> GameResult {
        let mut canvas =
            graphics::Canvas::from_frame(ctx, graphics::Color::from([0.1, 0.2, 0.3, 1.0]));

        let image = DynamicImage::new_rgb8(IMG_SIZE, IMG_SIZE);

        let image = Image::from_pixels(
            ctx,
            image.to_rgba8().as_bytes(),
            ImageFormat::Rgba8Unorm,
            IMG_SIZE,
            IMG_SIZE,
        );

        //comment out this line of code -> no memory leak
        // canvas.draw(&image, DrawParam::default());

        canvas.finish(ctx)?;

        Ok(())
    }
}

pub fn main() -> GameResult {
    let cb = ggez::ContextBuilder::new("super_simple", "ggez");
    let (mut ctx, event_loop) = cb.build()?;
    let state = MainState::new(&mut ctx)?;
    event::run(ctx, event_loop, state)
}

```
**Expected behavior**
No memory leak, image should be automatically dropped

**Hardware and Software:**
 - ggez version: 0.9.0-rc0 and current devel branch
 - OS: Windows 11
 - Graphics card: GTX 1660

Vixeliz:
Should be fixed with 0.9.2 so i'm gonna go ahead and close feel free to tell me to reopen if the problem still exist.

Issue #1159, opened 2023-01-23
Creating an Image/Canvas leaks tons of memory


nobbele:
like, crashes within a minute with a single image+canvas
TODO
caused by calling `new_canvas_image` every frame

nobbele:
Just creating one doesn't leak, only when you also use it

Vixeliz:
Should be fixed with 0.9.2 so i'm gonna go ahead and close feel free to tell me to reopen if the problem still exist.

Issue #1128, opened 2022-11-01
memory leak (probably command encoders not getting cleaned up)


Dumdidldum:
**Describe the bug**
Meshes are not cleaned up after they go out of scope

**To Reproduce**
Create a Mesh and drop it

**Expected behavior**
The mesh should be dropped on the GPU

**Screenshots or pasted code**
Adapted from the "super_simple" example
```
use ggez::{
    event,
    glam::*,
    graphics::{self, Color},
    Context, GameResult,
};
use log::LevelFilter;
use simplelog::{ColorChoice, TermLogger, TerminalMode};

struct MainState {}

impl event::EventHandler<ggez::GameError> for MainState {
    fn update(&mut self, _ctx: &mut Context) -> GameResult {
        Ok(())
    }

    fn draw(&mut self, ctx: &mut Context) -> GameResult {
        let mut canvas =
            graphics::Canvas::from_frame(ctx, graphics::Color::from([0.1, 0.2, 0.3, 1.0]));

        let circle = graphics::Mesh::new_circle(
            ctx,
            graphics::DrawMode::fill(),
            vec2(0., 0.),
            100.0,
            2.0,
            Color::WHITE,
        )?;

        canvas.draw(&circle, Vec2::new(0., 380.0));

        canvas.finish(ctx)?;

        Ok(())
    }
}

pub fn main() -> GameResult {
    TermLogger::init(
        LevelFilter::Info,
        simplelog::Config::default(),
        TerminalMode::Stdout,
        ColorChoice::Auto,
    )
    .unwrap();

    let cb = ggez::ContextBuilder::new("super_simple", "ggez");
    let (ctx, event_loop) = cb.build()?;
    let state = MainState {};
    event::run(ctx, event_loop, state)
}

```

**Hardware and Software:**
 - ggez version: 0.8.1
 - OS: Archlinux
 - Graphics card: Nvidia RTX 3070
 - Graphics card drivers: nvidia proprietary driver, version 520.56.06, from nvidia-all

jazzfool:
I can't seem to repro. note that the logger is spamming the terminal with info logs
Meshes themselves definitely do not leak (is your GPU memory going up or your system memory?)

Dumdidldum:
It seems like the system memory is going up (the games process that is)
The GPU does not seem to be affected actually, at least not in nvidia-smi

jazzfool:
Can you remove the logger and see if it still happens?

Dumdidldum:
[...]

PR #1214, opened 2023-07-07
Fix memory leaks

Vixeliz:
I think this is the cause for the following issues but need others to test for confirmation(Ik it at least fixes a memory leak in regards to drawing images just unsure if all of these issues fall under that same issue):

#1181
#1211
#1159
#1128


Vixeliz:
Closed in favor of a 0.9.1

PR #1219, opened 2023-07-09
WIP More proper memory leak fix

Vixeliz:
Not done yet, I think we should release 0.10 soonish so maybe include this then hence being pointed to devel.

One issue is buffer's are still building up but image's no longer are

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -708,8 +708,6 @@ impl GraphicsContext {
 
             self.staging_belt.recall();
 
-            self.bind_group_cache.clear();
-
             Ok(())
         } else {
             Err(GameError::RenderError(String::from(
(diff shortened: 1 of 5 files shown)


Vixeliz:
Oh also looking for input on the fetch_buffer function. I think the unwraps shouldn't fail but i'd like to know for sure so I can either comment them or replace them.

Vixeliz:
One other thing too set_text_image is a temporary workaround due to the fact that textrenderer at the moment stores a view and texture instead of a ggez::Image i'll most likely switch it to that. Speaking of the TextRenderer it seems the textrender may also be leaking memory. When rendering text buffers start piling on the gpu.

Vixeliz:
Closed in favor of a 0.9.2

PR #1220, opened 2023-07-09
Fix both an image memory leak and text memory leak. Prepare for 0.9.2

Vixeliz:
Hopefully fixes all the known memory leak's properly now. Testing is appreciated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.9.2
+
+## Fixed
+- Fixed memory leak caused by previous `Image` fix.
+- Fixed memory leak caused by `Text` rendering and creation
+
 # 0.9.1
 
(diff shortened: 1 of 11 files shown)


Vixeliz:
Fyi thanks to TrulyFalse(One of the main authors of the wgpu rewrite) on discord for help with both of these.

nobbele (approved):

nox:
Tbh this increased minimum Rust version, using `Option::is_ok_and`

Vixeliz:
@nox we target the latest stable release of rust. Sometimes it may compile on earlier versions but that is the target version.

nox:
The readme says

> ggez requires rustc >= 1.42 and is distributed on crates.io.

So you may want to remove that claim.

PR #1223, opened 2023-07-09
Basic implementation of async assset loading

nobbele:


sampullman:
All the commits here plus a few more are included in #1243, so it probably makes sense to close this one.

PR #1226, opened 2023-07-11
Remove Option from Drawable::dimensions

Vixeliz:


Vixeliz:
Meant to target at devel whoops

PR #1228, opened 2023-07-12
Optimizations for 3d

Vixeliz:
There are two part's of 3d taking significant frametime being aabb calculated(fixed by first commit) and cloning the mesh which I am less sure on how we want to handle without making user api annoying. Open to any ideas


Vixeliz:
I think this is ok to merge. If any problems arise I will fix them, but I personally can't see anything.

PR #1236, opened 2023-07-24
Added a tiny terminal game.

buraksenyurt:
Hello there,
I tried to write a very simple terminal game. The kind that will measure the attention of the player. Red balls are enemies, blues are allies, and greens are UFOs :) I hope I didn't code too badly.

### DIFF
--- a/examples/06_radar.rs
+++ b/examples/06_radar.rs
@@ -1,0 +1,247 @@
+//! Radar Game
+//!
+//! The aim of this game is to shoot vehicles that appear on the radar.
+//! However, it is necessary to pay attention to the distinction between allie and enemy.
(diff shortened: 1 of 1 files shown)


Vixeliz:
Sorry for the delayed response, I'll take a better look at this tomorrow and let you know if I spot anything, it would be cool to have more demo game examples though so I definitely want to take a look at this.

sampullman:
I also think it's good to include more demo games, but ideally they should cover something different than existing examples. This one is of similar complexity to the snake example, although it does have a good example of a mouse event in a game.

I think it could be alright to include if it were commented to the same extent as snake and astroblasto.

PR #1239, opened 2023-08-10
fix: use non-rust package in fedora dependency

arminveres:
Using rust-libudev-devel and rust-libudev-sys-devel installs cargo and rust through the package manager, which may not be preferable, if installed through rustup.


a1phyr (approved):
Thanks !

PR #1243, opened 2023-08-18
Web Support

Muvkina:
Ok here is one actually based on the branch here in the ggez repo lol.

I think we are getting *close* to this being ready to merge. 

**TODO**
- [x] All examples running on web: http://ggez.sampullman.com:4070
- [x] Document web setup/build
- [ ] Document minimal web example without Vite
- [ ] Decide whether to keep separate instanced drawing for webgl (alternatively only support WebGPU)
- [ ] Try to improve code for missing audio on init

Web (async) asset loading can be improved separately. The current solution of zip based solution is fine for the PoC.


metiulekm (commented):
I was looking at various Rust game engines and frameworks. I really liked ggez, but would like to target web, so I decided to test this PR. Here are some random comments :smiley:

Vixeliz:
Nice catch I'll fix both of these up soon

sampullman:
@Vixeliz I finally got around to testing and fixing some of the problems I had running this last year. Overall I think it's great work, and just needs a bit of cleanup and regression testing. I fixed the conflicts locally, as well as the comments from @metiulekm, and a few other minor build issues. Should I push directly to this branch, or is it easier if I merge a fork?

Here's a dump of some of my notes I haven't had a chance to address yet:
- Resource loading wasn't totally clear to me, it should be documented that they need to be available at `<web-server-root>/resources/`
- Audio throws console warnings and doesn't work for me, though it may be a problem on my end
- I understand why it makes sense for ggez to control the canvas creation, but I assume most users would want to handle styling/layout themselves. It can be worked around, but might be better to allow passing in a canvas element ID on initialization, if possible.
- Similar to `wgpu` examples, it would be nice to provide a complete example. Currently my procedure is:
```
cargo build --release --target wasm32-unknown-unknown
wasm-bindgen --out-dir ./web/wasm/ --target web ./target/wasm32-unknown-unknown/release/ggez-playground.wasm
```

In the `web` folder, I've installed Vite, and use it to run the following `index.html`:

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <link rel="icon" href="/favicon.png" />
    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
    <title>Hello ggez!</title>
    <style>
      html, body {
        height: 100%;
        padding: 0;
        margin: 0;
        box-sizing: border-box;
      }
      #ggez-wrap {
        display: flex;
        width: 100%;
        height: 100%;
        justify-content: center;
        align-items: center;
      }
    </style>
  </head>
  <body>
    <div id="ggez-wrap">
    </div>
    <script type="module">
      import game from "./wasm/ggez-playground";
      try {
      await game();
      } catch(_e) {
        // wgpu throws an error on init for control flow?
      }

      // Move canvas to our own element
      const ggezWrap = document.getElementById('ggez-wrap')
      const ggezBody = document.getElementById('ggez-body')
      ggezWrap.appendChild(ggezBody)
    </script>
  </body>
</html>
```

a1phyr:
I haven't got through all the Web-related in depth, but it looks fine.

However, this whole async story brings a lot of complexity, and it makes the API unwieldy. It works ok for 5 assets in the example, but it does not scale at all. I don't think that I fits very well with ggez' value of being easy to work with. 

Maybe we could suggest users to put their assets in a ZIP and start by downloading it somehow ? Something like this:

```rust
let assets = fetch_assets("/resources").await?;
let (context, events) = ContextBuilder::new("org", "game").add_zipfile_bytes().build()?;
```

sampullman:
@a1phyr I finally have some time to resume this, and I agree in general about the async complexity. I'm not sure adding a zip dependency the the web build is ideal either, but it might be simplest to do that first and explore other options later. If I can come up with something better I'll propose it here or post a demo on the discord.

Cargo.toml
```diff
@@ -17,59 +17,96 @@ exclude = ["docs", "resources", "!LiberationMono-Regular.ttf"]
...
+    "Node",
+    "Headers",
+    "Request",
+    "RequestInit",
+    "RequestMode",
+    "Response",
+]}
+instant = { version = "0.1", features = ["inaccurate", "wasm-bindgen"] }
+wasm-bindgen = "0.2"
+wasm-bindgen-futures = "0.4"
```
a1phyr: These should only be used with Wasm target

Cargo.toml
```diff
@@ -17,59 +17,96 @@ exclude = ["docs", "resources", "!LiberationMono-Regular.ttf"]
...
 bitflags = "2.1"
-zip = { version = "4", default-features = false, features = ["deflate-flate2"] }
+zip = { version = "8", default-features = false, features = ["deflate-flate2"] }
 directories = "6"
-wgpu = "26"
+wgpu = { version = "29", features = ["webgl"] }
```
a1phyr: I would prefer to have a feature (maybe enabled by default) that enables WebGL instead of always doing it.

Cargo.toml
```diff
@@ -17,59 +17,96 @@ exclude = ["docs", "resources", "!LiberationMono-Regular.ttf"]
 [lib]
 name = "ggez"
 path = "src/lib.rs"
+crate-type = ["cdylib", "rlib"]
```
a1phyr: Is it really necessary ? Isn't it only for the final executable?

src/audio.rs
```diff
@@ -24,29 +24,68 @@ use crate::filesystem::Filesystem;
 /// of your `Context` object.
 pub struct AudioContext {
     fs: Filesystem,
-    stream: rodio::OutputStream,
+    // `None` when no output device could be opened.
+    // [`Source`]/[`SpatialSource`] creation then fails gracefully instead of crashing
+    stream: Option<rodio::MixerDeviceSink>,
```
a1phyr, line 29: This is unfortunate :/

src/graphics/instance.rs
```diff
@@ -17,14 +19,60 @@ use std::{
...
 /// Traditionally known as a "batch".
 #[derive(Debug)]
 pub struct InstanceArray {
     pub(crate) buffer: Mutex<wgpu::Buffer>,
+    #[cfg(not(target_arch = "wasm32"))]
     pub(crate) indices: Mutex<wgpu::Buffer>,
+    #[cfg(not(target_arch = "wasm32"))]
     pub(crate) bind_group: Mutex<wgpu::BindGroup>,
+    #[cfg(not(target_arch = "wasm32"))]
     pub(crate) bind_layout: wgpu::BindGroupLayout,
```
a1phyr: Can you explain somewhere why is instanced drawing so different on web?

src/context.rs
```diff
@@ -376,15 +379,22 @@ impl ContextBuilder {
...
         )?;
 
-        for path in &self.paths {
-            fs.mount(path, true);
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            for path in &self.paths {
+                fs.mount(path, true);
+            }
         }
```
a1phyr: Nit: I think that you don't need the extra brackets.

src/context.rs
```diff
@@ -416,15 +426,22 @@ impl ContextBuilder {
...
```
a1phyr, line 532: Same here

a1phyr (commented):
I quickly went through it and still need to test, but here are a few comments in the meantime.

Also I couldn't set the target branch to `main`, do you think that it is possible?

Would it be possible to avoid using a JS framework for the examples ? I think there might be value in having something simple can could easily be reused, and I fear that it might get out-of-date. I am no expert on Web though, so it may not be pertinent.

Last question, do you mind if I release 0.10 now without this PR? The next version should come half June without new wgpu anyway.

Cargo.toml
sampullman: Yeah, I still need to clean up this and the rest of my Cargo.toml changes. I meant to document some more of my changes but it took me a lot longer than I thought to get everything working.

I think it's reasonable to disable the feature by default, at least until it's 100% solid.

src/audio.rs
sampullman: The problem I was having is that ggez initializes audio on start, and audio isn't available on web until the user interacts with the page.

I think it makes sense in general to not crash if audio isn't available, but maybe there's a cleaner way to handle it.

src/graphics/instance.rs
sampullman: Yeah, I need to write notes on a bunch of stuff. In short, WebGL doesn't support runtime sized storage buffers, and this is all a hacky workaround to switch `BufferUsages::UNIFORM` to `BufferUsages::VERTEX` on web.

Only supporting WebGPU would avoid this, but afaik it's not ubiquitous yet.

sampullman:
> Also I couldn't set the target branch to main, do you think that it is possible?

Whoops, I just did it.

> Would it be possible to avoid using a JS framework for the examples ? I think there might be value in having something simple can could easily be reused, and I fear that it might get out-of-date. I am no expert on Web though, so it may not be pertinent.

If you want an easy way to run web examples, something is needed beyond html+js to serve the wasm. There are workarounds like embedding as base64 but it's a hassle. Vite isn't being used as a framework here though, just a web server (and bundler for a "production" deploy on e.g. Github pages).

Everything works with html & plain JS, as long as you can serve the `.wasm` somehow. I could provide a bare minimum example in docs.

The JS glue is a bit heavy though, mainly due to my asset loading workaround.

> Last question, do you mind if I release 0.10 now without this PR? The next version should come half June without new wgpu anyway.

Yeah, no need to block on this.

Cargo.toml
sampullman: I needed "rlib" to get the examples running, wasm binaries consume ggez as rlib. I think you're right about cdylib though, I'm still not totally comfortable with rust/wasm build stuff.

Cargo.toml
sampullman: Added the feature, disabled by default for now.

Cargo.toml
sampullman: done

src/graphics/instance.rs
sampullman: Ok, after further thought I think WebGPU is the way to go. Platform support still isn't perfect (Firefox+Linux I think it the biggest missing thing?) but it's improving pretty quickly.

I don't think it's worth trying to maintain all these WebGL workarounds. I'm going to remove them and test a bit.

stefnotch: Yeah, going WebGPU-only is a reasonable choice. Linux support will come and the first version of ggez that runs on the web does not have to go out of its way to support every single use-case.

PR #1244, opened 2023-08-30
canvas/instancearray: custom sorting functions

jazzfool:
It was brought to my attention that GGEZ lacks any ability to support custom sorting, which would be helpful in, e.g., top-down games, where draws should be sorted by the Y coordinate. Currently the only way to sort is by using Z indices, which does not apply well to this scenario.

This PR changes two things:

1. The internal `BTreeMap` structures are replaced with a plain `Vec`.
2. The draw list (or indices) are stable sorted by a function which can be set by the user.

These changes apply to both `Canvas` and `InstanceArray`.

The default behavior remains identical to current, with draws ordered by Z and then by insertion (hence why a stable sort was chosen).

From my limited testing, average frame delta is essentially identical to the current `BTreeMap`-based sorting.

### DIFF
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -9,9 +9,9 @@ use super::{
     gpu::arc::{ArcBindGroup, ArcBindGroupLayout},
     internal_canvas::{screen_to_mat, InstanceArrayView, InternalCanvas},
     BlendMode, Color, DrawParam, Drawable, GraphicsContext, Image, InstanceArray, Mesh, Rect,
-    Sampler, ScreenImage, Shader, ShaderParams, Text, WgpuContext, ZIndex,
+    Sampler, ScreenImage, Shader, ShaderParams, Text, WgpuContext,
 };
-use std::{collections::BTreeMap, sync::Arc};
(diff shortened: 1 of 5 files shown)


Vixeliz:
Are you on the discord server for more rapid back and forth talking? (Then I'll update here with any relevant info)

jazzfool:
Have added support for 3D.

Ultimately for this change I don't believe making sort_by a dynamic function is necessary for two reasons:
1. Generally sorting behaviour should be pure; i.e., output should *only* rely on the arguments. This avoids any possible foot guns.
2. If the sorting behaviour changes on an external value, this can be moved *outside* the closure (i.e., at the call site of set_sort_by), instead of *inside* the closure.

Vixeliz:
This looks good imo, just making sure but u have tested just to make sure this works correctly right?

jazzfool:
Yep, just did one final pass to fix a bug with the 3D implementation. Confirmed that draw order is correct for 3D in RenderDoc. This should be ready for review/merge.

Vixeliz:
Kk i'll take another look at it but last time i looked, it looked all good

Vixeliz:
looks good to me ill merge it

PR #1246, opened 2023-08-31
fix(canvas): Reset shader params when setting default shader

jack-cooper:
Previous behaviour when using the `set_default_shader` function was to keep the old shader params under `self.state.params` around. This was conditionally giving me a panic with the following message:
```
The pipeline layout, associated with the current render pipeline, contains a bind group layout at index 3 which is incompatible with the bind group layout associated with the bind group at 3
```

Presumably the sensible fix is to simply also reset the shader parameters when we reset the shader. After discussion in Discord, I was told the same would apply to the text shader/params. I haven't used these at all, but the fix there looks trivially similar. If I'm at all wrong just let me know, thanks :smile:

### DIFF
--- a/src/graphics/canvas.rs
+++ b/src/graphics/canvas.rs
@@ -213,6 +213,9 @@ impl Canvas {
     }
 
     /// Resets the active mesh shader to the default.
+    ///
+    /// If any shader parameters were previously set with `set_shader_params`, these will also be
+    /// reset to the default of `None`.
     #[inline]
     pub fn set_default_shader(&mut self) {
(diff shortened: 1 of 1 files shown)


Vixeliz:
This looks good I just would add to the doc comments and mention parameters are also reset just to make it clear what is going on

jack-cooper:
Sounds very sensible to me, I've added a brief but hopefully clear explanation to each of them :+1:

Vixeliz:
Looks good thanks for the contribution

PR #1251, opened 2023-09-23
Fixed image.rs use of unstable library

voipllc:
fixed image.rs


a1phyr:
`Result::is_ok_and` is stable since rust 1.70. You should probably update your compiler.

Anyway, this code is inconsistent with poisoning handling and has unnecessary lookups and unwraps and should be refactored

PR #1252, opened 2023-09-25
Improve `Image::fetch_buffer`

a1phyr:


a1phyr:
There was a deadlock in the previous version

PR #1253, opened 2023-09-25
Update CI

a1phyr:
This also make `skeptic` not required at build for our users !


a1phyr:
https://github.com/ggez/ggez/blob/1a6c26d9f26b9877c26118720238bbeb80b65276/src/filesystem.rs#L59

Do you know why does this uses an `Arc` given that `OverlayFS` is `!Send + !Sync` ?

PR #1210, opened 2023-06-29
Fix Image::to_pixels()

Vixeliz:
Fixes: #1135

PR #1262, opened 2023-10-26
Moved fix of image.to_pixels and .encode to master

ryanpeach:
This is an actual bug fix from https://github.com/ggez/ggez/pull/1210. I hope it can go on main so that others don't encounter it.


Vixeliz:
I may talk with the other maintainers to do another minor release, otherwise in 0.10 it'll be included which now that im back working on ggez can hopefully be in the soon future

Issue #1084, opened 2022-08-12
Unclear how to use `InstanceArray` to draw `Mesh`es


snyball:
Used to do this with `MeshBatch` pre-`0.8-rc0`, requesting an example of how to use `InstanceArray` to do the same thing. Would be good to add it to the changelog for `0.8` as well I think.

PSteinhaus:
You’re right.
We should re-add the meshbatch-example (though under a slightly different name probably), to showcase how this works.

Until then:

You can use [`Canvas::draw_instanced_mesh`](https://docs.rs/ggez/0.8.0-rc0/ggez/graphics/struct.Canvas.html#method.draw_instanced_mesh) to draw the `InstanceArray` with an additional Mesh fed in, which will be used.

snyball:
@PSteinhaus Thanks, was able to port all my code to `0.8.0-rc0`!

The last hickup was that the `AsStd140` proc-macro expects `crevice` to be able globally, and because it's a re-export from  `Cargo.toml` it needs a specific version, so:

```rust
crevice = "0.8.0"
```

The version number is effectively a magic constant, so it should also be mentioned in the `changelog`. Alternatively re-export `crevice` as well, and update the shader example with a `use ggez::crevice` (that works right? idunno.)

PSteinhaus:
Good point. 
Re-exporting `crevice` seems right to me here. (Bonus points for reminding me that I need to update the examples afterwards, which I of course forgot last time I re-exported `glam` ;) )

PR #1274, opened 2024-03-03
Update docs with new crevice version

sampullman:
Related #1084

@nobbele I've been getting back into rust/ggez dev and looking through a few issues for low hanging fruit.

I could be mistaken, but don't think this is possible to solve without significant changes to `crevice`, since they hard code crate references in the proc macros.

This PR just updates the docs with the latest `crevice` version and adds a few notes in the examples. The issue can probably be closed.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -214,7 +214,7 @@ If you want to create your own structs to send as uniforms to your shaders, simi
 then those structs need to implement `AsStd140`, which can luckily be easily derived using crevice.
 
 So far, so good. But now I have to tell you that you also need to **depend on the version of crevice that we are using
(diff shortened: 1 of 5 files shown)


sampullman:
`skeptic` on the beta channel is failing to compile doc tests/examples. It seems it's an issue with detecting dependencies, but I'm not too familiar with `skeptic`.

Cargo.toml
```diff
@@ -71,7 +71,7 @@ keyframe = "1"
 keyframe_derive = "1"
 num-derive = "0.4"
 
-skeptic = "0.13"
+skeptic = { git = "https://github.com/Erk-/rust-skeptic.git", branch = "fix-rust-1-77" }
```
sampullman, line 74: @a1phyr This is a temp fix so we can keep doctests running in CI until I can find an alternative for `skeptic`.

I cherry-picked your other CI fixes here to test everything together.

PR #1275, opened 2024-03-11
Update `image` to 0.25

a1phyr:


sampullman:
Doc tests have been failing in CI, I believe since the latest Rust beta was released. It seems like something broke in `skeptic`, which is unfortunately unmaintained.

Otherwise this looks good!

a1phyr:
It works in stable but not in beta. After some research, it looks like `skeptic` fails to find dependencies (I opened budziq/rust-skeptic#141). Given that it will probably never be fixed, we should probably find a replacement or remove this check completely.

PR #1279, opened 2024-03-26
Fix CI

a1phyr:


sampullman:
The commit was included with #1274

PR #1281, opened 2024-04-21
Removed FAQ reference to no-longer-existing function

ketsuban:
The FAQ references `Context::print_resource_stats`, which no longer exists; it was superceded by `Filesystem::print_all` and `Filesystem::log_all` in commit 3e07581.

I am very out of practice with GitHub and accidentally only cloned the `master` branch, but it looks like the pull request still works when applied to the `devel` branch because I'm only changing one file which hasn't been changed between the two. I hope that's okay.


sampullman:
Thanks! As long as there's no conflicts, rebasing on `devel` is fine. It's better to squash the `typo` commit into the main one, but no worries.

PR #1283, opened 2024-05-07
Use `glam` feature of `crevice`

a1phyr:
This avoids many unnecessary conversions


sampullman:
Looks straightforward to me. Too bad we don't have benchmarks, though!

PR #1286, opened 2024-05-07
Fix borderless on Wayland

### LINKED ISSUE #1238: Fullscreen not working properly on Wayland
**Describe the bug**
The method `set_fullscreen(FullScreenType::Desktop)` on Wayland makes the window borderless but doesn't occupy the entire monitor screen.

**To Reproduce**
Running the following code should reproduce the issue:
```rust
use ggez::{*, Context, ContextBuilder, GameResult};
use ggez::graphics::Color;
use ggez::event::EventHandler;

fn main() {
    // Make a Context.
    let (mut ctx, event_loop) = ContextBuilder::new("my_game", "Cool Game Author")
        .build()

    // Create an instance of your event handler.
    // Usually, you should provide it with the Context object to
    // use when setting your game up.
    let my_game = MyGame::new(&mut ctx);

    // Run!
}

struct MyGame {
    // Your state here...
}

impl MyGame {
    pub fn new(ctx: &mut Context) -> MyGame {
        // Load/create resources such as images here.
        ctx.gfx.set_fullscreen(conf::FullscreenType::Desktop).unwrap();
        MyGame {
            // ...
        }
    }
}

impl EventHandler for MyGame {
   fn update(&mut self, _ctx: &mut Context) -> GameResult {
        // Update code here...
        Ok(())
    }

        let canvas = graphics::Canvas::from_frame(ctx, Color::WHITE);
        // Draw code here...
        canvas.finish(ctx)
    }
}
```

**Expected behavior**
The previous code should create a window in fullscreen.

**Hardware and Software:**
 - ggez version: 0.9.3
 - OS: Arch Linux
 - Graphics card: AMD Radeon RX 6800
 - Graphics card drivers: AMDGPU
 - Window manager: Kwin Wayland

a1phyr:
Closes #1238

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -881,12 +881,7 @@ impl GraphicsContext {
                 }
             }
             FullscreenType::Desktop => {
-                window.set_fullscreen(None);
-                window.set_decorations(false);
-                if let Some(monitor) = window.current_monitor() {
-                    window.set_inner_size(monitor.size());
-                    window.set_outer_position(monitor.position());
-                }
+                window.set_fullscreen(Some(winit::window::Fullscreen::Borderless(None)));
             }
         }
 


src/graphics/context.rs
```diff
@@ -881,12 +881,7 @@ impl GraphicsContext {
                 }
             }
             FullscreenType::Desktop => {
-                window.set_fullscreen(None);
-                window.set_decorations(false);
-                if let Some(monitor) = window.current_monitor() {
-                    window.set_inner_size(monitor.size());
-                    window.set_outer_position(monitor.position());
-                }
+                window.set_fullscreen(Some(winit::window::Fullscreen::Borderless(None)));
```
sampullman, line 884: I could be mis-remembering, but I thought the above code was a workaround for some other problem. Maybe fullscreen issues in Windows?

a1phyr: I don't think so, it has been like that since `ggez` use `winit`: https://github.com/ggez/ggez/pull/388/files#diff-35c3205180019dc523a7ec61de86d88ad0d0a9ab700191bb8e76590f3bf290f4R472-R487

sampullman: Fair enough! I may be remembering an old `winit` bug, unrelated to ggez.

svenstaro:
@a1phyr I would like for ggez to continue being maintained but neither I (nor apparently any of other maintainers) have time for it. Would you like to become maintainer?

sampullman:
Whoops, I thought a1phyr was already a maintainer.

It's hard to test this kind of change for regressions, but doesn't break anything for me on Mac/Ubuntu, at least.

svenstaro:
Oh indeed, they are! I didn't notice.

PR #1290, opened 2024-07-22
Add ability to set window icon with bytes

Vixeliz:
It's been a while since i've done any programming not sure if this is acceptable or not


Vixeliz:
I missed a couple things (updating the icon struct in WindowSetup) and also realized that conf of course needs to be able to be serialized and deserialized. Why is icon in conf anyways? I feel like that's something that usually wouldn't be? Would anyone be opposed of removing it from WindowSetup entirely? I'm not sure where would be best for it if not though.

PR #1293, opened 2024-08-29
Update wgpu to 22.1.0, winit to 0.29

tysen:
In 0.29, `winit` overhauled their keyboard API, upon which `ggez` relies. Rather than try to hide these changes from the `ggez` end user, this commit refactors `ggez::keyboard` to use the `winit` types.

`wgpu` 22.1.0 contains fewer changes relevant to the `ggez` end user, but the most important is that the `Window` in `GraphicsContext` is now wrapped in an `Arc` to take advantage of the new safe `create_surface`.

### DIFF
--- a/src/graphics/context.rs
+++ b/src/graphics/context.rs
@@ -49,7 +49,7 @@ pub(crate) struct FrameArenas {
 #[allow(missing_docs)]
 pub struct WgpuContext {
     pub instance: wgpu::Instance,
-    pub surface: wgpu::Surface,
+    pub surface: wgpu::Surface<'static>,
     pub device: wgpu::Device,
     pub queue: wgpu::Queue,
 }
@@ -59,7 +59,7 @@ pub struct WgpuContext {
 pub struct GraphicsContext {
     pub(crate) wgpu: Arc<WgpuContext>,
 
-    pub(crate) window: winit::window::Window,
+    pub(crate) window: Arc<winit::window::Window>,
     pub(crate) surface_config: wgpu::SurfaceConfiguration,
 
     pub(crate) bind_group_cache: BindGroupCache,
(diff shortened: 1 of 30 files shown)


src/input/keyboard.rs
```diff
@@ -107,128 +103,90 @@
...
+            && !self.previously_pressed_physical_keys.contains(physical_key)
     }
 
     /// Checks if a key with the corresponding scan code has been released this frame.
-    pub fn is_scancode_just_released(&self, code: ScanCode) -> bool {
-        !self.pressed_scancodes_set.contains(&code)
-            && self.previously_pressed_scancodes_set.contains(&code)
+    pub fn is_physical_key_just_released(&self, physical_key: &PhysicalKey) -> bool {
+        !self.pressed_physical_keys.contains(physical_key)
+            && self.previously_pressed_physical_keys.contains(physical_key)
```
a1phyr, line 189: `PhysicalKey` is `Copy`, it should maybe be passed by value. Is guess that you kept a reference for consistency with logical key functions ?

src/graphics/context.rs
```diff
@@ -853,7 +855,7 @@ impl GraphicsContext {
             FullscreenType::Windowed => {
                 window.set_fullscreen(None);
                 window.set_decorations(!mode.borderless);
-                window.set_inner_size(mode.actual_size()?);
+                let _ = window.request_inner_size(mode.actual_size()?);
```
a1phyr, line 858: Shouldn't this line be deleted ?

a1phyr (commented):
This is great ! Thanks a lot for this work !

src/graphics/context.rs
aleokdev: The `let _ =` is just there to ignore the result. I'm guessing `request_inner_size` has useful side effects.

PR #1294, opened 2024-09-06
Update winit to 0.30 (mvp)

woubuc:
This PR builds on top of the work done in #1293 to update winit to 0.29, and then adds the minimally necessary changes to update to winit 0.30 by moving the match arms in the `event::run` function into an internal `ApplicationHandler` implementation.

### Breaking change
As per the [changelog](https://github.com/rust-windowing/winit/releases/tag/v0.30.0), updating to winit to 0.30 constitutes a **cascading breaking change** due to a change in underlying system libraries for Wayland.

### MVP
I've only implemented the minimally needed changes to make ggez build with winit 0.30, as I needed winit 0.30 for my own use case. This code still uses the (now **deprecated** but not yet removed in 0.30) [`EventLoop::create_window`](https://docs.rs/winit/latest/x86_64-pc-windows-msvc/winit/event_loop/struct.EventLoop.html#method.create_window) function rather than creating the window inside the new application handler, as is the new winit approach going forward. Removing this deprecated usage will require more structural changes in the wider ggez codebase that I do not feel qualified enough to decide on or implement.

### Preparation
- `cargo check` only shows the deprecation warning as mentioned
- Tests seem to pass, save for things like fmt errors in files I did not touch so I haven't fixed those
- My own project builds and runs without any issue using this code using winit 0.30.

### Further upgrade work
As noted above, I don't intend to make the deeper changes needed to remove the deprecation warning at this point. I hope this PR can serve as a stepping stone for further work around this upgrade as a first hurdle has already been cleared (getting the project to build with winit 0.30). Feel free to close this if y'all feel that it doesn't contribute anything meaningful to the project in its current state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ libtest.rmeta
 Cargo.lock
 
 visual-tester/
+
+# Jetbrains IDE project directory
+.idea/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ zip = { version = "0.6", default-features = false, features = ["deflate"] }
 directories = "5.0"
 wgpu = "22.1.0"
 glyph_brush = "0.7"
-winit = { version = "0.29", features = ["serde"] }
+winit = { version = "0.30", features = ["serde"] }
 image = { version = "0.25", default-features = false, features = ["gif", "png", "pnm", "tga", "tiff", "webp", "bmp", "jpeg"] }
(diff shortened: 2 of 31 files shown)


sampullman:
Thanks! Once #1293 gets merged I'll take a closer look at this.

It would be nice to get it merged before the next release. There are already a ton of breaking changes, so it probably makes sense to tear off the bandaid and get packages up to date.

a1phyr:
Merged as #1298

PR #1298, opened 2024-11-25
Update winit to 0.30 (take 2)

tryone144:
I took the liberty to rebase #1294 onto `devel`. Considerations mentioned there still apply here.

- There should be no warnings anymore
- The custom event-loop example has been updated as well

---

> This code still uses the (now deprecated but not yet removed in 0.30) [EventLoop::create_window](https://docs.rs/winit/latest/x86_64-pc-windows-msvc/winit/event_loop/struct.EventLoop.html#method.create_window) function rather than creating the window inside the new application handler, as is the new winit approach going forward. Removing this deprecated usage will require more structural changes in the wider ggez codebase that I do not feel qualified enough to decide on or implement.

This is still the case. If I understand the design rationale behind winit correctly, creation of the window and initialization of the wgpu context should be done in the application handler (under control by the native event loop) after notified by the system (i.e. in the [`ApplicationHandler::resumed()`](https://docs.rs/winit/latest/winit/application/trait.ApplicationHandler.html#tymethod.resumed) callback).

In case of ggez, it appears to imply that the initialization of the `GraphicsContext` must happen after the event loop is started. This isn't a trivial change right now, as the `GraphicsContext` is tightly coupled to the general ggez `Context`. Possible solutions include:

1. Split creation of the `event_loop` and context initialization, i.e., create the event loop separately from calling `ContextBuilder::build()` and call `::build()` in winit's `resumed()` callback. This has the drawback that the final `ctx` cannot be passed to the user's state initialization (that implements `EventHandler`). This might be tackled by extending the `EventHandler` with a new `init()` function similar to [here](https://github.com/rust-windowing/softbuffer/blob/master/examples/utils/winit_app.rs).
2. Split initialization of `GraphicsContext` from `Context` and allow late initialization in the `resumed()` callback. This brings the ugliness of `Option<GraphicsContext>` and `.unwrap()`.

If there is a clear preference for any approach, I am open to create a PR for it (for further discussion).

---

Furthermore, winit suggests rendering in response to the [WindowEvent::RedrawRequested](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.RedrawRequested) event instead of in `about_to_wait()`:
> This is not an ideal event to drive application rendering from and instead applications should render in response to [WindowEvent::RedrawRequested](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.RedrawRequested) events.

This has not changed in this PR, but should be more straight forward than the initialization issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ zip = { version = "0.6", default-features = false, features = ["deflate"] }
 directories = "5.0"
 wgpu = "23"
 glyph_brush = "0.7"
-winit = { version = "0.29", features = ["serde"] }
+winit = { version = "0.30", features = ["serde"] }
 image = { version = "0.25", default-features = false, features = ["gif", "png", "pnm", "tga", "tiff", "webp", "bmp", "jpeg"] }
 rodio = { version = "0.20", optional = true, default-features = false, features = ["flac", "vorbis", "wav"] }
 serde = { version = "1.0", features = ["derive"] }
--- a/examples/eventloop.rs
+++ b/examples/eventloop.rs
@@ -10,95 +10,125 @@
 //! It is functionally identical to the `super_simple.rs` example apart from that.
 
 use ggez::graphics::{self, Color, DrawMode};
+use ggez::Context;
 use ggez::GameResult;
 use ggez::{event, GameError};
 
-use winit::event::{Event, WindowEvent};
-use winit::event_loop::ControlFlow;
+use winit::application::ApplicationHandler;
+use winit::event::{DeviceEvent, DeviceId, StartCause, WindowEvent};
+use winit::event_loop::{ActiveEventLoop, ControlFlow};
 use winit::keyboard::{Key, NamedKey};
+use winit::window::WindowId;
+
+struct CustomApplicationHandler {
+    ctx: Context,
+    position: f32,
+}
+
+impl ApplicationHandler<()> for CustomApplicationHandler {
+    fn new_events(&mut self, event_loop: &ActiveEventLoop, _: StartCause) {
+        if self.ctx.fields.quit_requested {
+            self.ctx.fields.continuing = false;
+        }
+        if !self.ctx.fields.continuing {
+            event_loop.exit();
+            return;
+        }
+
+        event_loop.set_control_flow(ControlFlow::Poll);
+    }
+
+    fn resumed(&mut self, _event_loop: &ActiveEventLoop) {}
+
+    fn window_event(
+        &mut self,
+        _event_loop: &ActiveEventLoop,
+        mut window_id: WindowId,
+        mut event: WindowEvent,
+    ) {
+        // This tells `ggez` to update it's internal states, should the event require that.
+        // These include cursor position, view updating on resize, etc.
+        event::process_window_event(&mut self.ctx, &mut window_id, &mut event);
+
+        match event {
+            WindowEvent::CloseRequested => {
+                self.ctx.request_quit();
+            }
+            WindowEvent::KeyboardInput { event, .. } => {
+                if Key::Named(NamedKey::Escape) == event.logical_key {
+                    self.ctx.request_quit();
+                }
+            }
+            // `CloseRequested` and `KeyboardInput` events won't appear here.
+            x => println!("Other window event fired: {x:?}"),
+        }
+    }
+
+    fn device_event(
+        &mut self,
+        _event_loop: &ActiveEventLoop,
+        mut device_id: DeviceId,
+        mut event: DeviceEvent,
+    ) {
+        // This tells `ggez` to update it's internal states, should the event require that.
+        // These include cursor position, view updating on resize, etc.
+        event::process_device_event(&mut self.ctx, &mut device_id, &mut event);
+
+        println!("Device event fired: {event:?}");
+    }
+
+    fn about_to_wait(&mut self, _event_loop: &ActiveEventLoop) {
+        // Tell the timer stuff a frame has happened.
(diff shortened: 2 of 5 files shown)


sampullman:
The changes so far look good! If I understand option 1 correctly, it's a pretty significant breaking change. It does seem like the better choice, though.

tryone144:
For the user-side this is a breaking change either way, because you will no longer be able to use the graphics context before the event loop has started (e.g. for uploading textures or meshes).

The main problem appears to be that we may only initialize the context after the event loop, but we want to use the context inside the event loop handler that has to be initialized before the event loop.

If I understand the winit event loop correctly, some platforms only allow use of window surfaces (i.e. create a window, get a wgpu surface) between calls to `resume()` and `suspend()` (or `exiting()`). Other platform require the drawing to the surface to happen in direct response to `WindowEvent::RedrawRequested`.

Without having an active window, the event loop should never receive `WindowEvent`s (or at least we can safely ignore them). So it might actually be possible to assume we have a `GraphicsContext` inside the `window_event()` handler, but not necessarily in `new_events()` or `about_to_wait()`.

```rust
struct App;

impl App {
  fn new() -> Result<Self, GameError> {
    Self {}
  }
}

impl EventHandler for App {
  fn update(&mut self, ctx: &mut Context) -> Result<(), GameError> {
    Ok(())
  }
  fn draw(&mut self, ctx: &mut Context) -> Result<(), GameError> {
    Ok(())
  }
  // …
}

struct GgezApplicationHandler {
  init: FnMut() -> Result<(), GameError>,
  ctx: Option<Context>,
  state: Option<App>,
}

impl ApplicationHandler for GgezApplicationHandler {
  fn resumed(&mut self, event_loop: &ActiveEventLoop) {
    // debug_assert xy.is_none()
    match self.init() {
      Ok((ctx, state)) => {
        self.ctx = Some(ctx);
        self.state = Some(state);
      },
      Err(err) => {
        error!("Error on init(): {e:?}");
        eprintln!("Error on init(): {e:?}");
        event_loop.exit();
      },
    }
  }

  fn suspended(&mut self, event_loop: &ActiveEventLoop) {
    // explicitly drop ctx and state
    let ctx = self.ctx.take();
    let state = self.state.take();
    // debug_assert xy.is_some()
    drop(ctx);
    drop(state);
  }

  fn window_event(&mut self, event_loop: &ActiveEventLoop, mut window_id: WindowId, mut event: WindowEvent) {
    // SAFETY: Only called between calls to `resumed()` and `supended()`
    let mut ctx = self.ctx.as_mut().unwrap();
    let mut state = self.state.as_mut()

    process_window_event(&mut ctx, &mut windowId, &mut event);

    // …
  }

  fn about_to_wait(&mut self, event_loop: &ActiveEventLoop) {
    if let Some(mut ctx) = self.ctx.as_mut() && let Some(mut state) = self.state.as_mut() {
      let time = HasMut::<TimeContext>::retrieve_mut(&mut ctx);
      time.tick();
      // …
      let res = state.update(&mut ctx);
      // …
    }
  }

  // …
}

pub fn event::run(init: FnMut() -> Result<(Context, EventHandler), GameError>) -> Result<(), GameError> {
  let event_loop = winit::event_loop::EventLoop<()>;
  let mut app = GgezApplicationHandler { init, ctx: OnceCell::new(), state: OnceCell::new()  };

  event_loop.run_app(&mut app).map_err(GameError::EventLoopError)
}

pub fn main() -> GameResult {
  event::run(|| 
    let cb = ContextBuilder::new();
    let ctx = cb.build()?;

    let app = App::new()?;

    (ctx, app)
  })
}
```

---

> Furthermore, winit suggests rendering in response to the [WindowEvent::RedrawRequested](https://docs.rs/winit/latest/winit/event/enum.WindowEvent.html#variant.RedrawRequested) event instead of in `about_to_wait()`

I have commits for the necessary changes as well. Keep in mind that this does split up game logic in `update()` and rendering in `draw()`. The former is called in the usual place in `about_to_wait()`, the later when needed in response to `WindowEvent::RedrawRequested`. To keep drawing, a call to `window.request_redraw()` is added after the `update()` logic, this will dispatch a `RedrawRequested` event for the next iteration.

Since `RedrawRequested` events can be issued from outside the event loop, we might render more often than updating. With the polling loop, this should seldom (if ever) be the case.
When running with vsync enabled, wgpu will block in `ctx.begin_frame()`, leading to no events being handled during the wait (16ms on 60Hz) and at least a frame delay between `update()` and `draw()`. Although this appears to be the case already with the current implementation.

I don't know if there is interest in it, but this latency could be reduced by pacing the generation of `RedrawRequested` events when vsync is enabled. Although I haven't found a stable cross-platform way of detecting when the next frame should be ready (either in winit nor wgpu). How complex should the "default" event loop implementation be?

a1phyr (commented):
Thanks for your work!

I think that we can keep using the deprecated API for now, and see later whether `winit` goes forward with removing it. 

About the `about_to_wait` stuff, let's keep it as is for now because this how it was before the PR, and do it as a followup, so we can go forward.

tryone144:
> I think that we can keep using the deprecated API for now, and see later whether `winit` goes forward with removing it.

Sure, should be safe for at least until winit v0.31. I have some notes for the required changes to make this work, but we can leave it be for now. As far as I can tell, only Android really benefits from the new API, everything else *should* work just fine.

> About the `about_to_wait` stuff, let's keep it as is for now because this how it was before the PR, and do it as a followup, so we can go forward.

I agree. Further research showed that this isn't a clear cut matter. As long as we always draw after a `RedrawRequested` event has been received (which we do) it should be fine. Wgpu allegedly takes care of acquiring the necessary locks.

PR #1300, opened 2024-12-28
Check if mesh is empty before drawing

snyball:
This fixes a crash when an empty mesh is drawn to a canvas. This can easily happen in e.g debug renderers based on `MeshBuilder`, when the builder is flushed without anything having been drawn to it.


a1phyr (approved):
Thanks!

PR #1302, opened 2025-01-15
Fix compiling when gamepad feature is disabled

orowith2os:
Previously, this would break compilation, because the trait bounds would never be met under this environment. This exposes the dummy object such that it's always met, and just a noop when running without gamepad support.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -15,20 +15,6 @@ use crate::graphics::GraphicsContext;
 use crate::input;
 use crate::timer;
 
-#[cfg(feature = "gamepad")]
-use crate::input::gamepad::GamepadContext;
-
-#[cfg(not(feature = "gamepad"))]
-/// Dummy gamepad context
-#[derive(Debug)]
-pub struct GamepadContext;
-
-#[cfg(not(feature = "gamepad"))]
-impl GamepadContext {
-    fn new() -> GameResult<Self> {
-        Ok(Self)
-    }
-}
 /// A `Context` is an object that holds on to global resources.
 /// It basically tracks hardware state such as the screen, audio
 /// system, timers, and so on.  Generally this type can **not**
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -1,4 +1,5 @@
 //! The `context` module contains functions and traits related to using the `Context` type.
+#![warn(missing_copy_implementations)]
```
Bowarc, line 2: This is useless, as it's already in lib.rs:188

src/context.rs
```diff
@@ -16,7 +17,7 @@ use crate::input;
 use crate::timer;
 
 #[cfg(feature = "gamepad")]
-use crate::input::gamepad::GamepadContext;
+pub use crate::input::gamepad::GamepadContext;
 
 #[cfg(not(feature = "gamepad"))]
 /// Dummy gamepad context
```
Bowarc, line 23: Consider adding Clone & Copy to the dummy context

orowith2os: Better that we not, just in case, to make sure nothing diverges too much between the two.

src/context.rs
orowith2os: I thought it didn't work here.... I'll check again before this is merged, just in case.

src/context.rs
Bowarc: Then, i think we should add something like
```rust
/// Dummy gamepad context
#[allow(missing_copy_implementations)]
#[derive(Debug)]
pub struct GamepadContext;
```

to get rid of the warning, maybe with a comment to explain

src/context.rs
orowith2os: Yeah, no, if I remove it, it'll trigger Copy warnings in users of ggez. I think we should keep it.

orowith2os:
Alright, fixed everything up nicely. Also properly got the warning_copy_implementations error to stop bothering us.

Bowarc:
Looks good to me

a1phyr (changes requested):

PR #1303, opened 2025-01-15
Update dependencies

orowith2os:
winit is untouched, see https://github.com/ggez/ggez/pull/1298 for updates on that

**Please only open PRs that are based on and aimed towards the `devel` branch**

**For more information take a look at [the contribution guide](./CONTRIBUTING.md)**


Bowarc:
Is there any reason to not remove the toml dependency ?

orowith2os:
Not sure, also not really my focus here. But I can take another look in the morning.

orowith2os:
`toml` removed, now it only relies on `toml_edit`, and not both.

a1phyr:
Could you use `basic_toml` instead ? I think that it is enough for our needs. Its API should look like `toml` 0.5. If it takes too much time, let it like that and I'll do it.

PR #1316, opened 2025-05-03
MeshBuilder: remove ambiguity in docs

meithecatte:
"Create a new mesh" is verbiage more appropriate for the functions in Mesh, which briefly create a MeshBuilder and then immediately finalize it. The documentation for the actual MeshBuilder functions should convey that one mesh will be built for the combination of all the builder functions that get called.


a1phyr (approved):
Thanks!

PR #1317, opened 2025-05-03
Mark the input::keyboard doctest as no_run

meithecatte:
Without this, `cargo test --doc` would open up a window...


a1phyr (approved):

a1phyr:
Thanks!

PR #1321, opened 2025-06-14
Fix raw_mouse_motion_event returning `dx` twice

tryone144:
Fix typo in call to `raw_mouse_motion_event`. This now properly reports the raw `y` delta as `dy`.


sampullman:
Good catch 👍

PR #1340, opened 2026-03-28
Various cleanups to prepare for version 0.10

a1phyr:
Close #1337


a1phyr:
Updated with latest `glam` and `crevice`

PR #1346, opened 2026-06-03
Remove deprecated authors field from guide

stefnotch:
The `authors` field in `Cargo.toml` is deprecated according to the documentation
https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field

So it can safely be removed from the ggez guide


a1phyr (approved):
Thanks!

PR #1349, opened 2026-07-02
Update `wgpu` and `glam`

a1phyr:


sampullman (approved):
Nice! I think the crevice update enables some cleanup, I was waiting on it for something, but can't remember exactly what.

a1phyr:
Seeing the changelog, I think that is was another crate.

PR #1350, opened 2026-07-07
fix: gracefully skip frame when surface is occluded or timed out

shiftz300:
## Problem

On macOS Metal, when the window is fully occluded (covered by another window), `wgpu::Surface::get_current_texture()` returns `CurrentSurfaceTexture::Occluded`. The existing code in `begin_frame()` uses:

```rust
Occluded | Timeout => {
    // TODO: Add proper way to skip frame
    continue;
}
```

This `loop { ... continue }` blocks the entire event loop indefinitely because `nextDrawable()` will not return until the surface becomes visible again. On macOS, the OS pauses the display link for occluded windows, so the surface stays occluded.

Additionally, `event.rs` called `event_loop.exit()` on any `begin_frame()` error, which is also undesirable for a transient condition like occlusion.

## Fix

Two changes:

1. **`src/graphics/context.rs`**: Replace the infinite `continue` loop with an error return.

2. **`src/event.rs`**: Restructure the `about_to_wait` handler to gracefully skip drawing when the surface is unavailable, while still running the necessary input state cleanup (mouse delta reset, keyboard/mouse state save).

## Testing

Tested on macOS 15 (Apple Silicon) with a ggez application. The application no longer freezes when the window is occluded.


a1phyr:
Thanks for the PR!

Could you change the `begin_frame` to return a `GameResult<bool>` instead? It would be better than having a special error string to match against.

src/graphics/context.rs
```diff
@@ -615,8 +617,7 @@ impl GraphicsContext {
                 }
                 wgpu::CurrentSurfaceTexture::Outdated => self.reconfigure_surface(),
                 wgpu::CurrentSurfaceTexture::Timeout | wgpu::CurrentSurfaceTexture::Occluded => {
```
a1phyr: Don't we want to retry on timeout instead of skipping the frame?

a1phyr:
You can rebase your MR on `master` to get the fixes for `clippy`.

